PR #3, opened 2020-10-26
Heroku Quality-of-Life Improvements

jacksonj04:
## Changes in this PR

- [x] Use Puma as the web server, instead of WEBrick.
- [x] Puma is configured with [Heroku's recommended options](https://devcenter.heroku.com/articles/deploying-rails-applications-with-the-puma-web-server)
- [x] Updates to Ruby 2.6.6
- [x] `app.json` has sensible defaults and requires appropriate environment configurations

## Future work

To complete this work, the application's deployment and development should be moved to use containers via Docker.


Procfile
```diff
@@ -0,0 +1 @@
+web: bundle exec puma -C config/puma.rb
```
tahb, line 1: What do you think about using the container deployment strategy instead? We have a [RFC that says we should run containers in live environments](https://github.com/dxw/tech-team-rfcs/blob/main/rfc-013-use-docker-to-deploy-and-run-applications-in-containers.md) so I think it might be best to start as we mean to go on - even if that means another card?

tahb: Here's an example from the last time I used Heroku with containers: https://github.com/LBHackney-IT/report-a-defect/blob/develop/heroku.yml

tahb (approved):

PR #4, opened 2020-10-26
[19] Users can see a start page for the planning journey

tahb:
## Changes in this PR

The service throws away the home page that comes from the Rails Template and replaces it with most (see commit) of the copy from the alpha. [The page on the prototype can be found here](https://dfe-hubs-prototype.herokuapp.com/sprint-one/support-start-page).

We create a new static page with HighVoltage rather than a controller since it seems a better fit for static content.

We have a card focused on addressing the styling so this work doesn't address styling considerations.

### Before
Screenshot 2020-10-26 at 14 16 23

### After

Screenshot 2020-10-26 at 12 24 10


app/views/pages/planning_start_page.html.erb
```diff
@@ -0,0 +1,15 @@
+<h1><%= I18n.t("planning.start_page.page_title") %></h1>
```
jacksonj04, line 1: I think this also needs including in the `<title>` for the page

jacksonj04 (changes requested):
One tweak to get the title in the `<title>` element, otherwise looks like a solid set of words.

app/views/pages/planning_start_page.html.erb
tahb: Ah yes, good spot 👍

tahb:
@jacksonj04 found another small thing on my way past https://github.com/DFE-Digital/buy-for-your-school/pull/4/commits/33bfa5b94abf7c65519186d528342f6a4296913d

jacksonj04 (approved):
LGTM 🚀

PR #10, opened 2020-10-27
Add step-by-step documentation for the release process

jacksonj04:

### DIFF
--- a/doc/release-process.md
+++ b/doc/release-process.md
@@ -1,0 +1,55 @@
+# Release process
+
+To create a new release and deploy it to production, follow this process:
+
+## 1. Create the release pull request
+
+1. Make a branch from `develop` called `release-xxx` (where `xxx` is the sequential release number)
+1. Move all features from the `Unreleased` section of [`CHANGELOG.md`](../CHANGELOG.md) to a new heading with the release number linked to a diff of the two latest versions, together with the date in the following format:
+
+  ```markdown
+  ## [Release XXX] - 2020-10-27
+
+  ...
+
+  [release xxx]:
+    https://github.com/DFE-Digital/buy-for-your-school/compare/previous-release...release-xxx
+  ```
+
+1. Create a commit for the release, including the changes for the release in the commit message
+1. Push the branch
+1. Open a pull request to merge the new branch to `develop` and get it reviewed
+
+## 2. Confirm the release and review the pull request
+
+The pull request should be reviewed to confirm that the changes currently are safe to ship and that [`CHANGELOG.md`](../CHANGELOG.md) accurately reflects the changes included in the release:
+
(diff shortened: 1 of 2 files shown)


doc/release-process.md
```diff
@@ -0,0 +1,51 @@
...
+1. Open a pull request to merge the new branch to `main` and get it reviewed
+
+## 2. Confirm the release and review the pull request
+
+The pull request should be reviewed to confirm that the changes currently are safe to ship and that [`CHANGELOG.md`](../CHANGELOG.md) accurately reflects the changes included in the release:
+
+1. Confirm the release with any relevant people (for example the product owner)
+1. Think about any dependencies that also need considering: dependent parts of the service that also need updating; environment variables that need changing/adding; third-party services that need to be set up/updated
+1. Merge the pull request with `main`
+1. Merge the `release-xxx` branch back into `develop`
```
tahb, line 30: Would we need a second pull request for this? I think we might end up with noisy merge commits going in both directions. To keep the history more succinct how about the release pull request goes to `develop` first and we're then able to manually 'promote' to main based off of a merge. We'd need to slacken the protection rules to allow it.

doc/release-process.md
```diff
@@ -0,0 +1,51 @@
...
+Once the pull request has been merged, create a tag against the merge commit in the format `release-xxx` (zero-padded again) and push it to GitHub:
+
+```sh
+git tag release-xxx merge-commit-for-release
+git push origin refs/tags/release-xxx
+```
+
+## 4. Validate the deployment
+
+Check in Heroku to make sure the deployment has gone as expected, and that no problems have arisen during build or release stages.
```
tahb, line 43: We could probably drop the specific reference "Heroku" from this and replace it with "live environment". I only mention it as it seems like a small bit of documentation that would easily missed after a infrastructure migration and could cause confusion.

doc/release-process.md
```diff
@@ -0,0 +1,51 @@
...
+
+## 5. Update Trello
+
+Features in the "Approved" column should be moved to "Done"
+
+## 6. Announce the release in Slack
+
+Let everybody know that a new release has been shipped.
```
tahb, line 51: May be worth including how we might want to do this to reduce decision fatigue. What do you think about deciding sending a message like RMI and RODA to a specific Slack channel?

Perhaps to keep our work in the open the DfE channel of #sct-buy-for-your-school might be a good place. A message such as:

> :ship: Release # for Buy for Your School is now going live [link to changelog] :ship:

tahb (approved):
LGTM. Shall we drop the branch protection on main?

jacksonj04:
> LGTM. Shall we drop the branch protection on main?

Yup!

PR #16, opened 2020-10-29
Replace contentful fixture with real API request

tahb:
## Changes in this PR

- replace the first question in the planning journey with a real Contentful API request
- the Contentful client is configurable through environment variables in keeping with a 12 factor app
- the first question we ask for is also configurable through an environment variable as opposed to hardcoding it this allows us a little flexibility
- handle the exceptional case where we ask Contentful for an Entry and it wasn't found, fail quickly and loudly to the user and report to the team via Rollbar. The rollbar includes 2 pieces of technical information to help us debug Contenful: `entry_id` and `space_id`

It's worth noting the relevant risk error we've identified to the ability to delete Contentful entries https://trello.com/c/cSuRk7Kj. Given they _can_ be deleted by users we should (and now do) handle that situation. 

## Screenshots of UI changes

Screenshot 2020-11-02 at 12 02 23
Screenshot 2020-11-02 at 12 09 21

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,3 +10,7 @@ ROLLBAR_ACCESS_TOKEN=ROLLBAR_ACCESS_TOKEN
 ROLLBAR_ENV=development
 
 DATABASE_URL=postgres://postgres@localhost:5432/buy-for-your-school-development
+
+# Contentful
+CONTENTFUL_SPACE=
+CONTENTFUL_ACCESS_TOKEN=
--- a/.env.example
+++ b/.env.example
@@ -13,5 +13,6 @@ DATABASE_URL=postgres://postgres@localhost:5432/buy-for-your-school-development
 
 # Contentful
 CONTENTFUL_SPACE=
+CONTENTFUL_ENVIRONMENT=master
 CONTENTFUL_ACCESS_TOKEN=
 CONTENTFUL_PLANNING_START_ENTRY_ID=
--- a/.env.example
+++ b/.env.example
@@ -14,3 +14,4 @@ DATABASE_URL=postgres://postgres@localhost:5432/buy-for-your-school-development
 # Contentful
 CONTENTFUL_SPACE=
 CONTENTFUL_ACCESS_TOKEN=
+CONTENTFUL_PLANNING_START_ENTRY_ID=
--- a/.env.example
+++ b/.env.example
@@ -12,6 +12,7 @@ ROLLBAR_ENV=development
 DATABASE_URL=postgres://postgres@localhost:5432/buy-for-your-school-development
 
 # Contentful
+CONTENTFUL_URL=cdn.contentful.com
 CONTENTFUL_SPACE=
 CONTENTFUL_ENVIRONMENT=master
 CONTENTFUL_ACCESS_TOKEN=
--- a/app/services/get_contentful_question.rb
+++ b/app/services/get_contentful_question.rb
@@ -1,8 +1,28 @@
 require "contentful"
 
 class GetContentfulQuestion
+  class EntryNotFound < StandardError
+    attr_accessor :message
+    def initialize(message)
+      @message = message
+    end
+  end
+
   def call(entry_id:)
-    contentful_client.entry(entry_id).raw
+    response = contentful_client.entry(entry_id)
+    if response.nil?
+      error_message = "The following Contentful entry identifier could not be found."
+      Rollbar.warning(
+        error_message,
+        contentful_url: ENV["CONTENTFUL_URL"],
+        contentful_space_id: ENV["CONTENTFUL_SPACE"],
+        contentful_environment: ENV["CONTENTFUL_ENVIRONMENT"],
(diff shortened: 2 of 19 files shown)


.env.example
```diff
@@ -10,3 +10,8 @@ ROLLBAR_ACCESS_TOKEN=ROLLBAR_ACCESS_TOKEN
 ROLLBAR_ENV=development
 
 DATABASE_URL=postgres://postgres@localhost:5432/buy-for-your-school-development
+
+# Contentful
```
jacksonj04, line 14: We should probably also specify the Contentful API endpoint here, so we can use the [preview API](https://www.contentful.com/developers/docs/references/content-preview-api/).

Pre-emptively, we might want the one for the [environment](https://www.contentful.com/developers/docs/concepts/multiple-environments/) as well.

app/services/get_contentful_question.rb
```diff
@@ -1,8 +1,16 @@
+require "contentful"
+
 class GetContentfulQuestion
```
jacksonj04, line 3: This might be better called `GetContentfulEntry` so the terminology lines up; also content types in Contentful include "question" and "decision", and this may well expand in future, but this method will retrieve them all.

app/services/get_contentful_question.rb
```diff
@@ -1,8 +1,22 @@
...
+    def initialize(message)
+      @message = message
+    end
+  end
+
   def call(entry_id:)
-    contentful_client.entry(entry_id).raw
+    response = contentful_client.entry(entry_id)
+    if response.nil?
+      error_message = "The following Contentful entry identifier could not be found: #{entry_id}"
```
jacksonj04, line 14: This should probably also include the environment, lest we try hunt down content in the wrong one. I don't think it needs the endpoint we're calling, as an entry of content will still exist in both live and preview.

jacksonj04 (commented):
A couple of tweaks around exactly how we interface with Contentful's mountain of different endpoints, environments and whatnot.

The process of retrieving content itself looks solid though!

.env.example
tahb: I agree that we will likely need those configurable too. For me, that point is probably not as part of this pull request as this story isn't driving that feature out. That said these do appear to be small changes so I've added 2 new commits along the lines of a preoptimsation. Happy to keep them in unless they become contentious in which case it may be easier to leave them until later to get this work through 👍

app/services/get_contentful_question.rb
tahb: Agree 🙌

tahb: Mm this service wasn't added in this pull request so I'm thinking it'll have to be a new commit.

app/services/get_contentful_question.rb
tahb: I agree with environment and think we should probably include the endpoint as well since it looks like it would change for the preview functionality https://www.contentful.com/developers/docs/references/content-preview-api/

spec/services/get_contentful_entry_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+        allow(contentful_client).to receive(:entry)
+          .with(anything)
+          .and_return(nil)
+
+        expect(Rollbar).to receive(:warning)
+          .with("The following Contentful entry identifier could not be found.",
+            contentful_url: ENV["CONTENTFUL_URL"],
+            contentful_space_id: ENV["CONTENTFUL_SPACE"],
+            contentful_environment: ENV["CONTENTFUL_ENVIRONMENT"],
+            contentful_entry_id: "123")
```
tahb, line 73: @jacksonj04 I made a further change in passing after your last look. I moved the entry ID out of the message and into the payload. This feels more natural and became obvious when I started adding more data in. 

I reckon if we put the entry_id in the message we'd end up with a unique rollbar event for each error too. My bad :)

tahb:
New commits: https://github.com/DFE-Digital/buy-for-your-school/pull/16/commits/bd44aa2ef8722878692853b864731f41261efeda, https://github.com/DFE-Digital/buy-for-your-school/pull/16/commits/9658568fbe5187617c549e3b0257ca27b209ff0b and https://github.com/DFE-Digital/buy-for-your-school/pull/16/commits/708630aa0f7875c6b7dca08c7832f85b95b17d9d. Back with you @jacksonj04

spec/services/get_contentful_entry_spec.rb
jacksonj04: Looks like a smart move to me!

jacksonj04 (approved):
Changes look good! ✅

jacksonj04 (changes requested):
Ugh, just spotted that the review app isn't happy...

jacksonj04 (changes requested):
The new Contentful API environment variables need adding to the `app.json` to allow review apps to run, but other than that this looks good!

jacksonj04 (approved):

PR #18, opened 2020-11-03
[75] Users can go to a second (or n) radio question if it exists

tahb:
## Changes in this PR

The service only renders radio questions. If the starting Contentful entry (which should be a radio at the moment) declares that the next question is also a radio then the service will route the user to a second question so that they can answer it.

Currently the Contentful Question model from Alpha that we have been using requires a `next` element/page. It was seemingly used to route the user to a confirmation page that we do not yet immitate (and may choose not to). The service will currently render the plan#show when there is no next question so there is a misalignment here, this scenario can never happen given the current question model. I think the service should - as a simplest first pass - keep looking for questions and if it can't find another assume the journey is complete. At the moment if an Entry is linked to that is not of type `question` then we fail loudly. Given the state of the Contentful set up this means the user journey can never succeed. I've created a new Contentful organisation (given we can't make a new space and don't want to break the alpha) to get the journey working for now. 

We still need to handle different types of question like text. We also should be smart about the copy of the form button when there is no next question. Feels like 2 new cards to me.

## Screenshots of UI changes

Screenshot 2020-11-04 at 12 32 25
Screenshot 2020-11-04 at 12 32 31
Screenshot 2020-11-04 at 12 34 14

If it goes bang 💥 the user will see this error page and we'll see this rollbar event (space ID omitted):

Screenshot 2020-11-04 at 12 38 04
Screenshot 2020-11-04 at 12 40 46

## Next steps

- update live environments with the new contentful API keys


app/services/create_planning_question.rb
```diff
@@ -5,14 +5,20 @@ def initialize(plan:)
   end
 
   def call
-    Question.create(
+    # QUESTION: Should we explode if we find an unexpected type?
```
jacksonj04: Initial thought is "yes", because if it's an unexpected type then it implies that the content designers had something in mind which the application cannot currently do.

But that's one for another PR, probably the multiple question types one.

app/services/create_planning_question.rb
```diff
@@ -23,9 +30,37 @@ def call
...
+    ALLOWED_QUESTION_TYPES.include?(content_type)
+  end
+
+  def unexpected_question_type?
+    !expected_question_type?
+  end
+
+  def send_rollbar_warning
+    Rollbar.warning(
+      "An unexpected Entry type was found instead of a question",
```
jacksonj04, line 57: "instead of a question" implies that question is the only valid type; this is fine for this PR, we just need to remember to tweak once we support multiple content types.

app/services/create_planning_question.rb
```diff
@@ -1,11 +1,18 @@
 class CreatePlanningQuestion
+  class UnexpectedContentType < StandardError; end
+
+  ALLOWED_QUESTION_TYPES = %w[question].freeze
```
jacksonj04, line 4: `ALLOWED_QUESTION_TYPES` should probably be `ALLOWED_CONTENT_TYPES`, or this feels a bit tautological

app/views/plans/show.html.erb
```diff
@@ -3,7 +3,9 @@
 <h1 class="govuk-heading-xl"><%= @plan.category.capitalize %></h1>
 <dl class="govuk-summary-list">
   <% @plan.questions.each do |question| %>
-    <dt class="govuk-summary-list__key"><%= question.title %></dt>
-    <dd class="govuk-summary-list__value"><%= question.answer.response.capitalize %></dd>
+    <div class="govuk-summary-list__row">
```
jacksonj04, line 6: Good catch!

jacksonj04 (approved):
Only one real change, but happy for this to go as is and we fix that (if you haven't already) in the PR for 87.

app/services/create_planning_question.rb
tahb: Good point. I could interpolate `ALLOWED_QUESTION_TYPES` here to keep it up to date.

app/services/create_planning_question.rb
tahb: Ah I did come back and answer my own question in a later commit https://github.com/DFE-Digital/buy-for-your-school/pull/18/commits/73b2df050fcc29b6f779f084778bb828f7e8d186 where I found I needed to deal with `unmanagedPage`. Then decided to add a new Contentful space which would have removed the need for error handling but it feels useful to keep in.

app/services/create_planning_question.rb
tahb: Indeed. I may even go one step further and include "Contentful" so we have a chance at understanding what content type it is associated with! So: `ALLOWED_CONTENTFUL_CONTENT_TYPES`.

PR #21, opened 2020-11-05
Refactor the getting of Contentful Entries to return an object instead

tahb:
This pull request has been extracted from the work being done to address https://trello.com/c/2r44iApA. In order to make this change I found some friction around reusing the `GetContentfulEntry` service. I think this refactoring offers improvement regardless of additional work.

Change the way we handle the Contentful API response so we pass down an object of the Contentful gem, rather than a blob of JSON (though we still do store the raw JSON).

### DIFF
--- a/app/services/create_planning_question.rb
+++ b/app/services/create_planning_question.rb
@@ -15,15 +15,15 @@ class CreatePlanningQuestion
     end
 
     question = Question.create(
-      title: contentful_response.dig("fields", "title"),
-      help_text: contentful_response.dig("fields", "helpText"),
-      contentful_type: contentful_response.dig("fields", "type"),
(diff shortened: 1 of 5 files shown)


app/services/create_planning_question.rb
```diff
@@ -41,7 +41,7 @@ def contentful_response
   end
 
   def content_type
-    contentful_response.dig("sys", "contentType", "sys", "id")
+    JSON(contentful_response.raw).dig("sys", "contentType", "sys", "id")
```
tahb, line 44: I can lose this dig too I've just realised.

jacksonj04 (approved):
This definitely looks loads easier to work with!

PR #27, opened 2020-11-09
Entry fixtures without next element are more realistic

tahb:
## Changes in this PR

Tiny change to correct 2 fixtures that attempt to replicate a response for entries with no 'next' element.

You can verify this by making a Contentful API request manually for an entry that has no next element, no `next` key is returned.


jacksonj04 (approved):
Good catch

PR #28, opened 2020-11-09
[88] Users can answer a short_text question

tahb:
## Changes in this PR

- we refactor the `CreatePlanningQuestion` service slightly so the `Entry` is injected and also an Answer is returned as part of the response. Injection helps keep the testing set up small and allows us more flexibility in approach. Whether we want to use different sub classes of `Answer` object or perhaps different types of creation service, for example: `CreateRadioQuestion => Question, Answer`.
- as well as radio questions, users can now answer a short text question
- we use the Question's Contentful type (for example, "radios", "short text") to dynamically decide which view to render to the user
- the `long_text` type is still to come
- the email address fixture was chosen but we haven't gone the extra step to validate that type of data yet
- capitalise answer responses on the way in rather than on the way out in rendering, it wouldn't make sense to capitalize an email address for example. A minor snag which we could easily remove altogether from this work.
- small fixes to all 3 factories so they work out of the box

On staging or locally you can use the starting entry ID of `hfjJgWRg4xiiiImwVRDtZ` to test this 🎉 

## Screenshots of UI changes

Screenshot 2020-11-09 at 16 01 16
Screenshot 2020-11-09 at 16 01 40
Screenshot 2020-11-09 at 17 24 37
Screenshot 2020-11-09 at 17 24 44


app/services/create_planning_question.rb
```diff
@@ -52,6 +47,33 @@ def unexpected_question_type?
...
+  def contentful_type
+    contentful_entry.type.tr(" ", "_")
+  end
+
+  def raw
+    contentful_entry.raw
+  end
+
+  def next_entry_id
+    return nil unless contentful_entry.respond_to?(:next)
```
tahb, line 73: Here's the nil guard I mentioned was on the way! It should allow us to solve this Rollbar event https://rollbar.com/dxw/dfe-buy-for-your-school/items/6/?utm_campaign=new_item_message&utm_medium=slack&utm_source=rollbar-notification

app/controllers/questions_controller.rb
```diff
@@ -14,8 +14,12 @@ def new
...
     redirect_to plan_path(@plan) unless @plan.next_entry_id.present?
 
-    @question = CreatePlanningQuestion.new(plan: @plan).call
-    @answer = Answer.new
+    contentful_entry = GetContentfulEntry.new(entry_id: @plan.next_entry_id).call
+    @question, @answer = CreatePlanningQuestion.new(
+      plan: @plan, contentful_entry: contentful_entry
+    ).call
+
+    render "new.#{@question.contentful_type}"
```
tahb, line 22: This choice implies a level of trust we have with Contentful types. What if a user deleted "short text" or renames it in such a way that it doesn't match directly to a view we have. This could actually happen.

Though `CreatePlanningQuestion` will raise an error if the Content model was unexpected, i.e it's not of type `question`. It does not check the type of a question itself, i.e "radios".

Should we be more cautious and continue trying to rescue the cases we do spot? If so, I'm happy to add this in a new small PR as this is already at 9 commits. Happy either way.

tahb: The fact I've written this is a probably a sign I think we should!

jacksonj04 (approved):

PR #31, opened 2020-11-11
[113] Enable Contentful entry preview

tahb:
## Changes in this PR

This pull request creates a new endpoint to enable us to configure urls within Contentful so that users can click "preview" and be redirected to the service. The new endpoint tries to redirect to the real question#new endpoint so that the preview functionality keeps high-ish parity.

Our previous routing (`/plan/:plan_id/question/new`) was incompatible as it had no space for `entry_id` to slot in. We need to either refactor it to fit an entry_id or create something as close as possible to enable previews. Given that we'd still need to create a plan record and that question/new feels like the right page to render question views (rather than `question/:entry_id/answer/new`) I've opted for a new endpoint for now.

[Contentful guidance for setting up an online environment](https://www.contentful.com/help/setup-content-preview/#preview-content-in-your-online-environment).

Add a beta banner so that we can add visual distinction between an environment only intended to be used for previews. This may be used on the staging _and_ live preview environments to help users distinguish. It adds a little protection should a real user find their way onto a preview app. 
 
## Screenshots of UI changes

Screenshot 2020-11-11 at 10 32 23
Screenshot 2020-11-11 at 10 32 31

## Next steps

- [ ] Add new Environment variable to all environments
- [ ] [Add new endpoints into the Contentful environment](https://app.contentful.com/spaces/rwl7tyzv9sys/settings/content_preview)


jacksonj04 (approved):
These all look good to me.

PR #32, opened 2020-11-11
[116] Users can be asked a long text question

tahb:
## Changes in this PR

As well as being asked a short text question the user can now be presented a long text question, which is essentially a larger input through a `textarea`. Tested with 1000 words. Tested with line breaks and formatting them back to the user.

You can test this with entry id `2jWIO1MrVIya9NZrFWT4e`.

Remove capitalisation of `answer.response` for now as it was doing weird things by downcasing all sentences after the first. I'd like to separately refactor the way we use a single generic Answer record down into an answer for each type to allow for this flexibility, i.e `RadioAnswer`,` LongTextAnswer` where the data we stored is of different types:

```
RadioAnswer.response => string (which we can capitalise on setting)
ShortTextAnswer.response => string 
LongTextAnswer.response => text
SelectAnswer.response => array
```

Worth a conversation first!

For now, the existing answer response column of type `string` works ok unless we expect more input than the tested 1000 words.

## Screenshots of UI changes

Screenshot 2020-11-11 at 12 27 10
Screenshot 2020-11-11 at 12 13 06


jacksonj04 (approved):
My only worries would be around the longer text, but planned refactoring of `Answer` into multiple types solves this so 🚢

PR #33, opened 2020-11-12
[116] Refactor answer creation to enable appropiate database types to be used

tahb:
## Changes in this PR

On the back of [finding some friction when adding LongText answer support](https://github.com/DFE-Digital/buy-for-your-school/pull/32) this change seeks to refactor the way we create and reference different types of answers.

- `RadioAnswer` is added and all existing answers (of that type) are migrated across
- `ShortTextAnswer` is added and all existing answers (of that type) are migrated across
- We introduce an `AnswerFactory` service that helps us create the right instance of an answer in a reusable way
- `Answer` is removed as both a table and a concept
- Eagerloading of all types of question answer has been whitelisted. This is probably the most controversial point of this proposition - though one we have discussed. The cost of eager loading at most 4 unique types of table for each question is thought to be acceptable in terms of performance, given the context of the app and that a single plan is only likely to have 1-30 questions. We also don't imagine a user will ever need to view more than 1 plan at a time, which might cause load issues. There may be something clever we can do with SQL to do selective includes by first interogating the questions. I fell into a rabbit hole. If there are any ideas for this I'd be happy to make this better and remove the whitelists.

Right now these new tables are identical however the approach should enable `LongTextAnswer` to start diverging the schema with a response type of `text`. These tables and associated models are free to also start diverging independently in order to collect different types of data, ie. hashes of multiple answers or documents.


config/environments/test.rb
```diff
@@ -59,5 +59,8 @@
     Bullet.enable = true
     Bullet.bullet_logger = true
     Bullet.raise = true # raise an error if n+1 query occurs
+    Bullet.add_whitelist type: :unused_eager_loading, class_name: "Question", association: :answer
```
jacksonj04, line 62: I think we can remove this line as `:answer` no longer exists

spec/models/question_spec.rb
```diff
@@ -21,4 +24,22 @@
 
     expect(question.raw).to eql("{\"foo\":{}}")
   end
+
+  describe "#answer" do
+    context "when a RadioAnwser is associated to the question" do
```
jacksonj04, line 29: Small typo in `RadioAnwser`

jacksonj04 (approved):
Approved, just a typo and an errant line I think we're safe to remove!

PR #34, opened 2020-11-12
release-001

tahb:
- address rails-template TODO
- any user can see a start page for the planning journey
- use GOV.UK Frontend framework
- users can see a basic form to start the planning journey sourced by a
Contentful fixture
- planning form is styled as a GOV.UK form
- validate that an answer is provided to a question
- the first planning question comes directly from Contentful
- handle the exceptional case when a Contentful entry is missing
- multiple radio questions can be answered in sequence
- users can be asked to answer a short text question
- Contentful can redirect users to preview endpoints

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog 1.0.0].
 
 ## [Unreleased]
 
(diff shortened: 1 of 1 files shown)


jacksonj04 (approved):
Exciting release time! 🚀

PR #39, opened 2020-11-16
Add foreign key contraints to the database

tahb:
## Changes in this PR

This proposal attempts to ensure we have better data integrity between our associations. 

I found when debugging locally and on staging that the database was starting to get into unexpected states. For example we could have an `answer` without a `question`, or the potential for a `question` to not have a `plan`. Adding cascade constraints will help us delete records in a reliable manner. Deleting a plan will delete all associated questions, and all questions will delete associated answers. The result of deleting a plan will now leave no orphan records that have no way of ever being referenced.

We may wish to use `:restrict` in future if we want to guard against accidental deletions.

[As pointed out in the Rails documentation we can do this in the model or at the database level for a better guarantee](https://guides.rubyonrails.org/active_record_migrations.html#active-record-and-referential-integrity).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog 1.0.0].
 
 ## [Unreleased]
 
+- add database foreign key constraints for better data integrity
+
 ## [release-002] - 2020-11-16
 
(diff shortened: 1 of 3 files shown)


jacksonj04 (approved):
Hooray for more data integrity!

PR #40, opened 2020-11-16
(53) Docker Compose used on CI and usable in dev

tahb:
## Changes in this PR

- Puma workers reduced from 2 to single process. If we want to scale the app with more resources we can go horizontally with more containers, rather than horizontally.
- Docker Compose can be used to run local dev servers using Docker, included `script/server`
- Docker Compose can be used to run local test servers using Docker, included in the readme. Both Docker and Non-Docker options are referenced in the Readme as there are trade-offs to using each
- Refactor the Dockerfile into build layers to help debug a problem with the assets being unavailable only on CI

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -5,6 +5,7 @@ services:
       context: .
       args:
         RAILS_ENV: "test"
+        BUNDLE_EXTRA_GEM_GROUPS: "test"
     command: bundle exec rake
     ports:
       - "3000:3000"
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -1,36 +1,39 @@
 version: "3.8"
 services:
-  web:
+  test:
     build:
       context: .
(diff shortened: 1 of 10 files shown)


docker-compose.test.yml
```diff
@@ -0,0 +1,40 @@
...
+    ports:
+      - "3000:3000"
+    depends_on:
+      - test-db
+    env_file:
+       - .env.test
+    environment:
+      DATABASE_URL: postgres://postgres:password@test-db:5432/buy-for-your-school-test
+      DATABASE_CLEANER_ALLOW_REMOTE_DATABASE_URL: "true"
+    volumes:
```
tahb, line 19: This line is the reason we need a new Docker Compose file for CI purposes. It doesn't need `tty` or `stdin_open` either as it will never be run interactively.

jacksonj04 (approved):
🚀

PR #46, opened 2020-11-25
(112) Contentful requests use a new Redis cache

tahb:
## Changes in this PR

- install redis
- use redis as a read cache when getting any Contentful entry
- set a time to live value of 48 hours to start with
- environment variable introduced so we can toggle the behaviour on and off (useful for the preview env too)


config/initializers/redis.rb
```diff
@@ -0,0 +1,14 @@
+module RedisCache
```
tahb, line 1: We're using a module here as opposed to a global variable which rightly violates a linting rule.

app/services/get_contentful_entry.rb
```diff
@@ -40,4 +45,39 @@ def send_rollbar_warning
...
+    redis_cache.get(cache_key)
+  end
+
+  def store_in_cache(raw_contentful_response:)
+    return unless ENV["CONTENTFUL_ENTRY_CACHING"] == "true"
+    return unless raw_contentful_response.present? &&
+      raw_contentful_response.respond_to?(:raw)
+
+    redis_cache.set(cache_key, raw_contentful_response.raw)
+    redis_cache.expire(cache_key, 172_800) # 48 hours
```
tahb, line 71: On further thought I'm wondering about 2 things:

1. 48 seems a long time for the content team to push through changes to users
2. Should this value also be configurable through env?

jacksonj04: Following discussion:

1. Yes, but planning to forcibly refresh cache at 3am brings this down to a maximum of 24 hours. If there is a need for a faster feedback loop we can look at webhooks.
1. Yes, in progress.

jacksonj04 (approved):

tahb:
@jacksonj04 I've added 2 new commits if you wanted another look otherwise I intend to merge when back from lunch :)

PR #48, opened 2020-11-26
Refactor Contentful interactions into dedicated interface class

jacksonj04:
## Changes in this PR

Contentful interactions are now handled by the `ContentfulConnector` class, to better separate concerns and reduce duplication when different kinds of Contentful API calls are made in future.

### DIFF
--- a/app/services/contentful_connector.rb
+++ b/app/services/contentful_connector.rb
@@ -1,0 +1,16 @@
+require "contentful"
+
+class ContentfulConnector
+  def initialize
+    @contentful_client ||= Contentful::Client.new(
+      api_url: ENV["CONTENTFUL_URL"],
+      space: ENV["CONTENTFUL_SPACE"],
+      environment: ENV["CONTENTFUL_ENVIRONMENT"],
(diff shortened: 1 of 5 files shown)


app/services/contentful_connector.rb
```diff
@@ -0,0 +1,16 @@
+require "contentful"
```
tahb, line 1: We can probably lose this from the `GetContentfulEntry` service now

spec/services/contentful_connector_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      CONTENTFUL_ACCESS_TOKEN: contentful_access_token
+    ) do
+      example.run
+    end
+  end
+
+  describe "#get_entry_by_id" do
+    it "returns a Contentful entry by making a call to Contentful" do
+      contentful_client = instance_double(Contentful::Client)
+      expect(Contentful::Client).to receive(:new)
```
tahb, line 23: It's nice that all this extra Client setup work is now homed and tested seperately 👍

tahb (approved):
LGTM. Will need another rebase on my changes to make TTL configurable since you last rebased this, I think.

PR #49, opened 2020-11-26
(112) Fix reading from the Redis cache

tahb:
## Changes in this PR

Addresses https://rollbar.com/dxw/dfe-buy-for-your-school/items/31/?utm_campaign=new_item_message&utm_medium=slack&utm_source=rollbar-notification

A string was being passed into the Redis cache that could not be parsed back out by `JSON.parse`. I should have run it locally one last time before creating the PR.

Instead of blindly storing the `entry.raw` JSON blob, we now serialise and deserialise it with `JSON.dump` and `JSON.restore`. 

Here's an illustration for how the data should change along the way before and after:

Before: 
```
Set -> 
  "{\"sys\"=>{\"space\"…
Get -> 
  JSON.parse('{"sys"=>{"space"…) -> 💥 JSON::ParseError
```

After: 
```
Set -> 
  "{"sys"=> {"space"=>".to_json -> "{\"sys\":{\"space\"
  JSON.dump("{\"sys\":{\"space\":) -> "\"{\\\"sys\\\":{\\\"space\\\": # This is what goes into Redis
Get -> 
  JSON.restore("\"{\\\"sys\\\":{\\\"space\\\":) -> "{\"sys\":{\"space\":
  JSON.parse("{\"sys\":{\"space\":) -> {"sys"=> {"space"=> # This is what's passed into Contentful::ResourceBuilder
```

There is a warning on the use of the `JSON.load` method [1] (aliased as `restore`) however since we are only using it to store trusted content from Contentful which only the team can edit, this is a "client under our control".

## Next steps

- [ ] Mark the rollbar event as resolved

### DIFF
--- a/app/services/get_contentful_entry.rb
+++ b/app/services/get_contentful_entry.rb
@@ -13,7 +13,7 @@ class GetContentfulEntry
       entry = find_and_build_from_cache
     else
       entry = @contentful_connector.get_entry_by_id(entry_id)
-      store_in_cache(raw_contentful_response: entry)
+      store_in_cache(entry: entry)
     end
 
     if entry.nil?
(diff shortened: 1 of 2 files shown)


tahb:
I have turned off caching on staging and flushed redis so that there should be no old and corrupt cache entries.

jacksonj04 (approved):

PR #55, opened 2020-11-30
(171) Automatically warm Contentful cache every night

tahb:
## Changes in this PR

- install Sidekiq to enable asynchronous tasks
- install Sidekiq Cron to enable schedule asynchronous tasks. When Sidekiq is initialised the contents of the schedule.yml is automatically seeded into Redis
- refactor the way we send cache key information to the Cache class to make it reusable for this work
- get all Contentful entries and update their cached values **every night at 3am**

Remember if you'd like to test this locally to turn caching on with the `CONTENTFUL_ENTRY_CACHING` environment variable.

## Screenshots
Screenshot 2020-12-03 at 15 33 31

## Next steps

- [x] set new environment variables on all Heroku environments for `REDIS_SIDEKIQ_URL`. Copy the same value from`REDIS_URL` and add a new database onto the end name eg. `/1`, so long as it's different from the `REDIS_CACHE_URL`.

### DIFF
--- a/app/jobs/warm_entry_cache_job.rb
+++ b/app/jobs/warm_entry_cache_job.rb
@@ -1,0 +1,22 @@
+class WarmEntryCacheJob < ApplicationJob
+  include CacheableEntry
+
+  queue_as :caching
+  sidekiq_options retry: 5
+
+  def perform
+    entries = GetAllContentfulEntries.new.call
+    entries.each do |entry|
+      store_in_cache(cache: cache, key: "contentful:entry:#{entry.id}", entry: entry)
+    end
+  end
+
+  private
+
+  def cache
+    @cache ||= Cache.new(
+      enabled: ENV.fetch("CONTENTFUL_ENTRY_CACHING"),
+      ttl: ENV.fetch("CONTENTFUL_ENTRY_CACHING_TTL", 60 * 60 * 48)
+    )
+  end
+end
--- a/app/jobs/warm_entry_cache_job.rb
+++ b/app/jobs/warm_entry_cache_job.rb
@@ -16,7 +16,7 @@ class WarmEntryCacheJob < ApplicationJob
   def cache
     @cache ||= Cache.new(
       enabled: ENV.fetch("CONTENTFUL_ENTRY_CACHING"),
-      ttl: ENV.fetch("CONTENTFUL_ENTRY_CACHING_TTL", 60 * 60 * 48)
+      ttl: ENV.fetch("CONTENTFUL_ENTRY_CACHING_TTL", 60 * 60 * 72)
     )
   end
 end
--- a/spec/jobs/warm_entry_cache_job_spec.rb
+++ b/spec/jobs/warm_entry_cache_job_spec.rb
@@ -1,0 +1,46 @@
+require "rails_helper"
+
+RSpec.describe WarmEntryCacheJob, type: :job do
+  include ActiveJob::TestHelper
+
+  before(:each) { ActiveJob::Base.queue_adapter = :test }
+
+  around do |example|
+    ClimateControl.modify(
+      CONTENTFUL_ENTRY_CACHING: "true"
+    ) do
+      example.run
+    end
+  end
+
+  describe ".perform_later" do
+    it "enqueues a job asynchronously on the caching queue" do
+      expect {
+        described_class.perform_later
+      }.to have_enqueued_job.on_queue("caching")
+    end
+
+    it "asks GetAllContentfulEntries for the Contentful entries" do
+      raw_response = File.read("#{Rails.root}/spec/fixtures/contentful/multiple-entries-example.json")
+      response_hash = JSON.parse(raw_response)
+      fake_contentful_entry_array = Contentful::ResourceBuilder.new(response_hash).run
+
+      get_all_contentful_entries_double = instance_double(GetAllContentfulEntries)
+      allow(GetAllContentfulEntries).to receive(:new).and_return(get_all_contentful_entries_double)
+      allow(get_all_contentful_entries_double).to receive(:call).and_return(fake_contentful_entry_array)
+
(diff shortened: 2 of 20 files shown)


app/jobs/warm_entry_cache_job.rb
```diff
@@ -0,0 +1,22 @@
...
+      store_in_cache(cache: cache, key: "contentful:entry:#{entry.id}", entry: entry)
+    end
+  end
+
+  private
+
+  def cache
+    @cache ||= Cache.new(
+      enabled: ENV.fetch("CONTENTFUL_ENTRY_CACHING"),
+      ttl: ENV.fetch("CONTENTFUL_ENTRY_CACHING_TTL", 172_800) # 48 hours
```
erbridge, line 19:
```suggestion
      ttl: ENV.fetch("CONTENTFUL_ENTRY_CACHING_TTL", 60 * 60 * 48)
```

erbridge, line 19: What happens if the connection to Contentful goes down for more than 48 hours (e.g. over the weekend)?

spec/jobs/warm_entry_cache_job_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  around do |example|
+    ClimateControl.modify(
+      CONTENTFUL_ENTRY_CACHING: "true"
+    ) do
+      example.run
+    end
+  end
+
+  describe "#perform_later" do
```
erbridge, line 14:
```suggestion
  describe ".perform_later" do
```

It's a class method isn't it?

spec/jobs/warm_entry_cache_job_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    it "asks GetAllContentfulEntries for the Contentful entries" do
+      raw_response = File.read("#{Rails.root}/spec/fixtures/contentful/multiple-entries-example.json")
+      response_hash = JSON.parse(raw_response)
+      fake_contentful_entry_array = Contentful::ResourceBuilder.new(response_hash).run
+
+      get_all_contentful_entries_double = instance_double(GetAllContentfulEntries)
+      allow(GetAllContentfulEntries).to receive(:new).and_return(get_all_contentful_entries_double)
+      allow(get_all_contentful_entries_double).to receive(:call).and_return(fake_contentful_entry_array)
+
+      described_class.perform_now
```
erbridge, line 30: This one tests `.perform_now` not `.perform_later` as indicated by the `describe`.

spec/services/cache_spec.rb
```diff
@@ -2,15 +2,14 @@
...
-    expect(result.key).to eql("redis-key")
+    result = described_class.new(enabled: "true", ttl: 123)
     expect(result.enabled).to eql("true")
     expect(result.ttl).to eql(123)
   end
 
   describe "#redis_cache" do
     it "returns an object that quacks like a redis instance (since MockRedis is used in Test)" do
-      result = described_class.new(key: anything, enabled: anything, ttl: anything)
+      result = described_class.new(enabled: anything, ttl: anything)
```
erbridge, line 12: Is `anything` a magic rspec thing?

erbridge (commented):
Looks like a sensible approach to me. A few comments.

app/jobs/warm_entry_cache_job.rb
tahb: If Contentful goes down for longer the cache will expire and the service will break. 

[48 hours was originally chosen](https://github.com/DFE-Digital/buy-for-your-school/blob/develop/app/services/get_contentful_entry.rb#L13) in an attempt to balance the likelihood of that sustained downtime occurring for longer than 48 hours against content within the service becoming overly stale. Content designers don't have a way to cache bust yet, the nightly task is the only mechanism for the moment.

We could set this cache to be a longer, say a week but it feels arbitrary? Perhaps 72 would cover a Friday night failure, through to Monday Morning to give Contentful time to fix it though you'd expect out of hours support for a critical failure. I'd feel more comfortable setting duration that was much longer when we've done the path analysis work to make sure the content is first valid.

48 hours is reused from the [cache on request work]().

erbridge: If everything's working fine, the cache would be updated every night, right? Does it make sense for the content cache to time out at all? If the connection breaks and it takes a long time to fix, do we want the cached content to ever go away? Food for thought. I suspect 72 hours (to allow for a failure spanning a weekend) is fine for now.

spec/jobs/warm_entry_cache_job_spec.rb
tahb: Quite right. I've wrapped this in a Active Job test helper method now 👍

spec/services/cache_spec.rb
tahb: It is indeed a magic thing. When your test isn't directly influenced by an argument but we have to honour the interface we can use `anything`. 

This removes the need for lots of noisy set up needed to get the tests to run, which can really get in the way when reading it when determining what's relevant and what's not.

app/jobs/warm_entry_cache_job.rb
tahb: Yes it would be updated every night. I agree there is value in always having a trustworthy set of data to fallback on in the cache, though treating Redis as persistent data store seems like a mismatch. 

Happy to change to 72 hours.

A future idea could be to add error handling to the job and _if_ it happened, to extend the expiry of entries in the cache and raise a very noisy error so we are alerted to the underlying issue.

tahb: I've changed to 72.

jacksonj04 (approved):
That test is a bit gnarly, but I can't think of a cleaner way to do it without just generating busywork.

It's a yes from me!

PR #61, opened 2020-12-01
Fix form flow failures

tahb:
## Changes in this PR

- [fix DoubleRenderError](https://rollbar.com/dxw/dfe-buy-for-your-school/items/34/)
- fix bug which allowed a user to easily refresh a question page to get to the next question. You can test this by going to https://dfe-buy-for-your-school-beta-s.herokuapp.com/plans/b313b903-319f-45fb-9bb5-f0cc1a2608ca/questions/new and refreshing your browser. After this change it is still technically possible for this to happen *if* a user manually edits/hacks the URL back to `question/new`. The chances of this are small, the impact is the plan#show blows up. We may wish to add additonal guard logic to question#new to protect against this if it becomes a problem. I'm concious not all steps will require answers so it might be better to wait.


jacksonj04 (approved):
This feels tighter to me, 👍

PR #63, opened 2020-12-02
(179) Show static content from Contentful as a journey step

tahb:
This work is based on top of the fixes here https://github.com/DFE-Digital/buy-for-your-school/pull/61 without them the test suite fails.

## Changes in this PR

- accept a new Contentful model of `StaticContent` that has the type `paragraphs`
- content steps don't require an answer, the journey show page doesn't add a line summary for steps that don't have answers
- we store the `ContentfulModel` on the step so we can start to ask questions of Steps to help with presentation. Instead of showing steps if they have or don't have associated answers, we can make the assumption that if the step is of type "staticContent" then they'll never have an answer, but questions always should!

## Screenshots of UI changes
Screenshot 2020-12-02 at 11 12 17

### DIFF
--- a/app/services/create_journey_step.rb
+++ b/app/services/create_journey_step.rb
@@ -3,8 +3,8 @@ class CreateJourneyStep
 
   class UnexpectedContentfulStepType < StandardError; end
 
-  ALLOWED_CONTENTFUL_MODELS = %w[question].freeze
-  ALLOWED_CONTENTFUL_QUESTION_TYPES = %w[radios short_text long_text].freeze
+  ALLOWED_CONTENTFUL_MODELS = %w[question staticContent].freeze
+  ALLOWED_CONTENTFUL_QUESTION_TYPES = %w[radios short_text long_text paragraphs].freeze
 
   attr_accessor :journey, :contentful_entry
   def initialize(journey:, contentful_entry:)
@@ -26,6 +26,7 @@ class CreateJourneyStep
     step = Step.create(
       title: title,
       help_text: help_text,
+      body: body,
       contentful_model: content_model,
       contentful_type: step_type,
       options: options,
@@ -71,6 +72,11 @@ class CreateJourneyStep
   def help_text
     return nil unless contentful_entry.respond_to?(:help_text)
     contentful_entry.help_text
+  end
+
+  def body
+    return nil unless contentful_entry.respond_to?(:body)
+    contentful_entry.body
   end
 
   def options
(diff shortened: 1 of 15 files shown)


db/migrate/20201202150359_add_contentful_model_to_step.rb
```diff
@@ -0,0 +1,5 @@
+class AddContentfulModelToStep < ActiveRecord::Migration[6.0]
+  def change
+    add_column :steps, :contentful_model, :string
```
tahb, line 3: Eee I need to figure out how best to migrate existing data so it isn't nil.

app/views/questions/new.paragraphs.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= content_for :title, @question.title %>
```
tahb, line 1: Well this needs removing as questions have been changed to steps 🤦

app/services/create_journey_step.rb
```diff
@@ -3,8 +3,8 @@ class UnexpectedContentfulModel < StandardError; end
 
   class UnexpectedContentfulStepType < StandardError; end
 
-  ALLOWED_CONTENTFUL_MODELS = %w[question].freeze
-  ALLOWED_CONTENTFUL_QUESTION_TYPES = %w[radios short_text long_text].freeze
+  ALLOWED_CONTENTFUL_MODELS = %w[question staticContent].freeze
+  ALLOWED_CONTENTFUL_QUESTION_TYPES = %w[radios short_text long_text paragraphs].freeze
```
tahb, line 7: Rename `ALLOWED_CONTENTFUL_QUESTION_TYPES` to `ALLOWED_CONTENTFUL_ENTRY_TYPES`

spec/models/step_spec.rb
```diff
@@ -42,4 +42,20 @@
...
+    context "when the contentful model is 'question'" do
+      it "returns true" do
+        step = build(:step, :radio, contentful_model: "question")
+        expect(step.question?).to eq(true)
+      end
+    end
+
+    context "when the contentful model is NOT 'question'" do
+      it "returns false" do
+        step = build(:step, :radio, contentful_model: "staticContent")
```
jacksonj04, line 56: Is the `:radio` needed here? It seems odd to be creating a radio step and then specifying a static content model.

jacksonj04 (approved):
One small query about a factory in a test, otherwise :shipit:

spec/models/step_spec.rb
tahb: Oh that's a copy pasta error 🙈

PR #64, opened 2020-12-02
Docker test instruction fixed

tahb:
## Changes in this PR

The docker-compose service in docker-compose.test.yml is called `test` rather than `web` so this command didn't work.


jacksonj04 (approved):
Good spot

PR #66, opened 2020-12-03
Add class to retrieve all entries from Contentful

jacksonj04:
## Changes in this PR

Adds a new `GetAllContentfulEntries` class which retrieves all entries from a Contentful instance.

## Next steps

This class will be used by subsequent work on mapping and cache-warming.


tahb (approved):
Looks good to me 🚀

PR #71, opened 2020-12-08
(221) Users can be asked a single date question

tahb:
## Changes in this PR

- add missing test coverage for LongTextAnswer that was noticed during this
- Single date questions can be asked

The language "Single" was picked because I expect we'll have to add support for another type of question where we need to ask for 2 dates to form a "Range".

## Screenshots of UI changes

Screenshot 2020-12-08 at 11 41 19
Screenshot 2020-12-08 at 11 41 15

### DIFF
--- a/app/views/steps/new.single_date.html.erb
+++ b/app/views/steps/new.single_date.html.erb
@@ -1,0 +1,9 @@
+<%= content_for :title, @step.title %>
+
+<%= form_for @answer, as: :answer, url: journey_step_answers_path(@journey, @step), method: "post" do |f| %>
+  <%= f.govuk_date_field :response,
+    legend: { text: @step.title, size: "xl" },
+    hint: { text: @step.help_text }
+  %>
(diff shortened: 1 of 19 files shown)


app/views/steps/new.single_date.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= content_for :title, @step.title %>
+
+<%= form_for @answer, as: :answer, url: journey_step_answers_path(@journey, @step), method: "post" do |f| %>
+  <%= f.govuk_date_field :response,
+    legend: { text: @step.title, size: "xl" },
+    hint: { text: @step.help_text }
+  %>
+  <%= f.submit I18n.t("generic.button.soft_finish"), class: "govuk-button" %>
```
jacksonj04, line 8: This will need tweaking to use the new Contentful-specified button text

jacksonj04 (approved):
One tweak to account for the change in how we set button content (once a merge and rebase has happened)

PR #72, opened 2020-12-08
Refactor step view logic into presenter

tahb:
Based on https://github.com/DFE-Digital/buy-for-your-school/pull/71

## Changes in this PR

I recently added a method into the Step model that I would have liked to add to a Presenter, this small change does that refactor so the next view logic is easier to add.


jacksonj04 (approved):

jacksonj04 (commented):
LGTM

PR #75, opened 2020-12-10
Step form refactorings for easier reuse

tahb:
## Changes in this PR

- refactor the step views so they only use a single definition of a create form
- use a layout variable to enable the views to be reused for both create and update journeys


LornaHarwood (approved):
This is clear and easy to follow.
I can see the benefit in this refactor and it all looks good to me

PR #76, opened 2020-12-10
(115) Journey map endpoint which shows the sequence of Contentful entries shown to a user

tahb:
## Changes in this PR

- a new hidden page at `/journey_map/new` for viewing the live map of all entries used in journey
- this map only supports a single linear journey, it starts from `ENV["CONTENTFUL_PLANNING_START_ENTRY_ID']` and continues until no `next` element can be found
- it doesn't read from the cache as we specified it should be live and reloadable
- it doesn't write back to the cache yet as that would introduce a cache busting mechanic, I think we should do the work to validate the paths look healthy before adding them to the cache

When designing the `EntryPathFinder` service and thinking about how this work could be extended to support multiple branches. We could replace the whole thing with a graphing tool or change this service so that it starts to build a proper tree data structure instead of a list.

## Screenshots of UI changes

Note the link in the bottom left.

Screenshot 2020-12-14 at 11 02 42

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog 1.0.0].
 - form button content is configurable through Contentful
 - users can be asked to provide a single date answer
 - add Webmock to prevent real http requests in the test suite
+- content users can see a journey map of all the Contentful steps
 
 ## [release-003] - 2020-12-07
 
--- a/app/controllers/journey_maps_controller.rb
+++ b/app/controllers/journey_maps_controller.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+class JourneyMapsController < ApplicationController
+  def new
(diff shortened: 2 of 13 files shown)


app/services/entry_path_finder.rb
```diff
@@ -0,0 +1,37 @@
+class EntryPathFinder
```
jacksonj04, line 1: This name to me suggest that "the job of this class is to find an entry path", not "the job of this class is to find the path from an entry". I can't think of a good alternative though.

jacksonj04 (approved):
One comment on gnarly class naming (but not blocking), other than that it's good to go!

app/services/entry_path_finder.rb
tahb: Have renamed to `BuildJourneyOrder` after our brief chat on Slack!

PR #79, opened 2020-12-14
Test suite detects external HTTP requests

tahb:
## Changes in this PR

Using Webmock will give the team more explicit feedback on any failures due to a real request being made.

Instead of seeing a real Contentful 404 we should see a Webmock error that explains that a real HTTP request was detected and that we should look to stub it.

### Before
Screenshot 2020-12-14 at 12 09 47

### After
Screenshot 2020-12-14 at 12 09 57


jacksonj04 (approved):
We should probably make a note to add this to the rails template as well.

tahb:
Indeed we should

PR #80, opened 2020-12-15
(121) Part 1 - Only generate a journey map if the journey path is valid

tahb:
Based on https://github.com/DFE-Digital/buy-for-your-school/pull/76

Do we need more rules? Can we think of any other reasons to invalidate a journey for now?

## Changes in this PR

My plan is that this is the first of a 2 parter. If this validation is OK then I'll look to apply it when warming the cache. If an error occurs we can extend the TTL of the existing entries to ensure a happy journey always persists.

- when the journey shows the user the same step twice we infer an infinite loop and show an error on `journey_map/new`
- when the journey steps a user see exceeds 50 we infer there's no end and show an error on `journey_map/new`
- generating a real fixture for 50+ entries slows down the test suite considerably (+2mins) so we opt for stubbing the limit to a lower value. I have a hunch it's to do with the internals of https://github.com/DFE-Digital/buy-for-your-school/blob/develop/spec/support/contentful_helpers.rb#L71 though I could see no value in continuing the investigation.

## Screenshots of UI changes

Screenshot 2020-12-15 at 15 00 51
Screenshot 2020-12-15 at 15 00 29

### DIFF
--- a/app/services/build_journey_order.rb
+++ b/app/services/build_journey_order.rb
@@ -1,5 +1,10 @@
 class BuildJourneyOrder
   class RepeatEntryDetected < StandardError; end
+
+  class TooManyChainedEntriesDetected < StandardError; end
+
+  ENTRY_JOURNEY_MAX_LENGTH = 50
+
   attr_accessor :entries, :starting_entry_id
 
   def initialize(entries:, starting_entry_id:)
@@ -27,6 +32,11 @@ class BuildJourneyOrder
     if entries.include?(entry)
       send_rollbar_error(message: "A repeated Contentful entry was found in the same journey", entry_id: entry.id)
       raise RepeatEntryDetected.new(entry.id)
+    end
+
+    if entries.count >= ENTRY_JOURNEY_MAX_LENGTH
+      send_rollbar_error(message: "More than #{ENTRY_JOURNEY_MAX_LENGTH} steps were found in a journey map", entry_id: entry.id)
+      raise TooManyChainedEntriesDetected.new(entry.id)
     end
 
     entries << entry if entry.present?
--- a/app/services/build_journey_order.rb
+++ b/app/services/build_journey_order.rb
@@ -1,4 +1,5 @@
 class BuildJourneyOrder
+  class RepeatEntryDetected < StandardError; end
   attr_accessor :entries, :starting_entry_id
 
   def initialize(entries:, starting_entry_id:)
@@ -22,6 +23,12 @@ class BuildJourneyOrder
 
   def recursive_path(entry_lookup:, next_entry_id:, entries:)
     entry = entry_lookup.fetch(next_entry_id, nil)
+
+    if entries.include?(entry)
+      send_rollbar_error(message: "A repeated Contentful entry was found in the same journey", entry_id: entry.id)
+      raise RepeatEntryDetected.new(entry.id)
+    end
+
     entries << entry if entry.present?
 
     if entry.respond_to?(:next)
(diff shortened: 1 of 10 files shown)


app/services/build_journey_order.rb
```diff
@@ -22,6 +28,17 @@ def entry_lookup_hash
...
   def recursive_path(entry_lookup:, next_entry_id:, entries:)
     entry = entry_lookup.fetch(next_entry_id, nil)
+
+    if entries.include?(entry)
+      send_rollbar_error(message: "A repeated Contentful entry was found in the same journey", entry_id: entry.id)
+      raise RepeatEntryDetected.new(entry.id)
+    end
+
+    if entries.count >= ENTRY_JOURNEY_MAX_LENGTH
+      send_rollbar_error(message: "More than 50 steps were found in a journey map", entry_id: entry.id)
```
jacksonj04, line 38: We could replace the `50` with the value of `ENTRY_JOURNEY_MAX_LENGTH` to have one fewer thing to adjust if needed.

```suggestion
      send_rollbar_error(message: "More than #{ENTRY_JOURNEY_MAX_LENGTH} steps were found in a journey map", entry_id: entry.id)
```

config/locales/en.yml
```diff
@@ -46,3 +46,9 @@ en:
     unexpected_contentful_step_type:
       page_title: "An unexpected error occurred"
       page_body: "The service has had a problem trying to retrieve the required step. The team have been notified of this problem and you should be able to retry shortly."
+    repeat_step_in_the_contentful_journey:
+      page_title: "An unexpected error occurred"
+      page_body: "One or more steps in the Contentful journey would leave the user in an infinite loop. This entry ID was presented more than once to the user: %{entry_id}"
+    too_many_steps_in_the_contentful_journey:
+      page_title: "An unexpected error occurred"
+      page_body: "More than 50 steps were found in the Contentful journey. Is the journey missing an end? The last Entry ID was: %{entry_id}"
```
jacksonj04, line 54: As with the Rollbar message, dropping the variable in here would be nice (although this ain't blocking by any means).

spec/services/build_journey_order_spec.rb
```diff
@@ -24,5 +24,71 @@
...
+        expect {
+          described_class.new(
+            entries: fake_entries,
+            starting_entry_id: "5kZ9hIFDvNCEhjWs72SFwj"
+          ).call
+        }.to raise_error(BuildJourneyOrder::RepeatEntryDetected)
+      end
+    end
+
+    context "when the journey visits more than 50 entries" do
```
jacksonj04, line 59:
```suggestion
    context "when the journey visits more than the maximum permitted number of entries" do
```

spec/services/build_journey_order_spec.rb
```diff
@@ -24,5 +24,71 @@
...
+        # Creating 50 linked entries in a fixture slows the test suite down
+        # We assume these 50 entries are chained together in a linear sequence.
+        fake_entries = fake_contentful_entry_array(
+          contentful_fixture_filename: "closed-path-with-multiple-example.json"
+        )
+        expect(BuildJourneyOrder::ENTRY_JOURNEY_MAX_LENGTH).to eq(50)
+        stub_const("BuildJourneyOrder::ENTRY_JOURNEY_MAX_LENGTH", 1)
+
+        expect(Rollbar).to receive(:error)
+          .with("More than 50 steps were found in a journey map",
```
jacksonj04, line 78:
```suggestion
          .with("More than #{ENTRY_JOURNEY_MAX_LENGTH} steps were found in a journey map",
```

spec/services/build_journey_order_spec.rb
```diff
@@ -24,5 +24,71 @@
...
+        end
+      end
+
+      it "raises a rollbar event with the last entry" do
+        # Creating 50 linked entries in a fixture slows the test suite down
+        # We assume these 50 entries are chained together in a linear sequence.
+        fake_entries = fake_contentful_entry_array(
+          contentful_fixture_filename: "closed-path-with-multiple-example.json"
+        )
+        expect(BuildJourneyOrder::ENTRY_JOURNEY_MAX_LENGTH).to eq(50)
```
jacksonj04, line 74: I'm happy leaving this as a fixed value (so there's a double-check that we're actually testing a given length of chain).

jacksonj04 (approved):
A few tweaks around reducing the number of places we declare the number 50 - I suspect the journey may end up in this ballpark! Other than that, this looks solid.

PR #81, opened 2020-12-15
(118) Users can edit their answers

LornaHarwood:
## Changes in this PR

-  The `Check Your Answers` page displays a `Change` link for each step and relating answer
- The `Edit` and `Update` actions are implemented to allow users to edit their answers once given

### `Check Your Answers` page with Change links
 
Screenshot 2020-12-15 at 15 59 41

### Navigates to the specific step to amend answer

Screenshot 2020-12-15 at 16 00 23

### Saves new content and updates answer on `Check Your Answers` page

Screenshot 2020-12-15 at 16 00 33

### DIFF
--- a/app/views/journeys/_long_text_answer.html.erb
+++ b/app/views/journeys/_long_text_answer.html.erb
@@ -1,4 +1,5 @@
 <div class="govuk-summary-list__row">
   <dt class="govuk-summary-list__key"><%= step.title %></dt>
   <dd class="govuk-summary-list__value"><%= simple_format(step.answer.response) %></dd>
+  <dd class="govuk-summary-list__actions"><%= link_to I18n.t("generic.button.change_answer"), edit_journey_step_path(@journey, step), class: "govuk-link" %></dd>
 </div>
--- a/app/views/steps/_edit_form_wrapper.html.erb
+++ b/app/views/steps/_edit_form_wrapper.html.erb
@@ -1,0 +1,5 @@
+<%= content_for :title, @step.title %>
+<%= form_for @answer, as: :answer, method: "put", url: journey_step_answer_path(@journey, @step, @answer) do |f| %>
+  <%= yield f %>
+  <%= f.submit I18n.t("generic.button.update"), class: "govuk-button" %>
+<% end %>
--- a/spec/features/visitors/anyone_can_edit_their_answers_spec.rb
+++ b/spec/features/visitors/anyone_can_edit_their_answers_spec.rb
@@ -1,0 +1,37 @@
+require "rails_helper"
+
+feature "Users can edit their answers" do
+  let(:answer) { create(:short_text_answer, response: "answer") }
+  scenario "The journey summary page displays a change link" do
+    visit journey_path(answer.step.journey)
+
+    expect(page).to have_content("answer")
+    expect(page).to have_content("Change")
+  end
+
+  scenario "The edited answer is saved" do
+    visit journey_path(answer.step.journey)
+
+    click_on(I18n.t("generic.button.change_answer"))
+
+    fill_in "answer[response]", with: "email@example.com"
+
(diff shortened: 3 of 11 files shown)


app/views/journeys/_long_text_answer.html.erb
```diff
@@ -1,4 +1,5 @@
 <div class="govuk-summary-list__row">
   <dt class="govuk-summary-list__key"><%= step.title %></dt>
   <dd class="govuk-summary-list__value"><%= simple_format(step.answer.response) %></dd>
+  <dd class="govuk-summary-list__actions"><%= link_to "Change", edit_journey_step_path(@journey, step), class: "govuk-link" %></dd>
```
tahb, line 4: We could use out internationalisation pattern to set the copy for the button here and in the other partials this is being added to. The advantage of this is that in future we can DRY it up and be able to change the copy in one place.

This feels like a generic button to me so we could add it within here https://github.com/DFE-Digital/buy-for-your-school/blob/develop/config/locales/en.yml#L7.

We can also use Internationlisation in our tests. Changing this hard coded language in this line: `click_on "Change"`

You can read more about how this is used here https://guides.rubyonrails.org/i18n.html#abstracting-localized-code and many other tests we already have can be used as a reference too.

app/views/steps/_edit_form_wrapper.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= content_for :title, @step.title %>
+<%= form_for @answer, as: :answer, method: "put", url: journey_step_answer_path(@journey, @step, @answer) do |f| %>
+  <%= yield f %>
+  <%= f.submit @step.primary_call_to_action_text, class: "govuk-button" %>
```
tahb, line 4: Given the edit journey doesn't "Continue" to subsequent steps as the create journey does, the use of `@step.primary_call_to_action_text` is arguably not needed here.

I think we could replace this with "Update".

If users don't like "Update" or another non-dynamic value we can then look at making that dynamic language too, though I doubt that'll be necessary. 

The use of Internationalisation from my other comment would then apply here too.

spec/features/visitors/anyone_can_edit_their_answers_spec.rb
```diff
@@ -0,0 +1,36 @@
+require "rails_helper"
+
+feature "Users can edit their answers" do
+  let(:answer) { create(:short_text_answer, response: "answer") }
+  scenario "The journey summary page displays a change link" do
+    visit journey_path(answer.step.journey)
+
+    expect(page).to have_content("answer")
+    expect(page).to have_content("Change")
+  end
```
tahb, line 10: Missing a new line here.

spec/features/visitors/anyone_can_edit_their_answers_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    click_on "Change"
+
+    fill_in "answer[response]", with: "email@example.com"
+
+    click_on(I18n.t("generic.button.next"))
+
+    expect(page).to have_content("email@example.com")
+  end
+
+  context "An error is thrown" do
```
tahb, line 23: Nice unhappy path coverage 👍

tahb (commented):
Looks good 👍 I've left 2 suggestions for additional improvements that I think we should look at.

spec/features/visitors/anyone_can_edit_their_answers_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    scenario "When an answer is invalid" do
+      visit journey_path(answer.step.journey)
+
+      click_on "Change"
```
jacksonj04, line 27: When these swap to use Internationalisation don't forget to update this too, or something will unexpectedly break in future!

tahb (approved):
LGTM 👍 🚀

PR #82, opened 2020-12-16
(121) Part 2 - Only warm the cache with entries that form valid journeys

tahb:
Based on https://github.com/DFE-Digital/buy-for-your-school/pull/80

## Changes in this PR

We use the validation added in https://github.com/DFE-Digital/buy-for-your-school/pull/80 in the nightly cache warming task.

If we find an error in the journey we no longer add it to the cache, instead we will extend all previous entries on the assumption that they were valid.

### DIFF
--- a/app/services/cache.rb
+++ b/app/services/cache.rb
@@ -26,4 +26,13 @@ class Cache
     redis_cache.set(key, value)
     redis_cache.expire(key, ttl)
   end
+
+  def extend_ttl_on_all_entries(extension: (60 * 60 * 24))
+    redis_cache.keys("contentful:entry:*").map do |key|
+      previous_ttl = redis_cache.ttl(key)
+      extended_ttl = extension + previous_ttl
+
+      redis_cache.expire(key, extended_ttl)
+    end
+  end
 end
--- a/spec/services/cache_spec.rb
+++ b/spec/services/cache_spec.rb
@@ -1,6 +1,10 @@
 require "rails_helper"
 
 RSpec.describe Cache do
+  after(:each) do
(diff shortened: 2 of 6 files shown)


app/services/cache.rb
```diff
@@ -26,4 +26,13 @@ def set(key:, value:)
     redis_cache.set(key, value)
     redis_cache.expire(key, ttl)
   end
+
+  def extend_ttl_on_all_entries(extension: (60 * 60 * 48))
```
jacksonj04, line 30: I'm not certain 48 hours is needed here; this task runs every 24 hours, and the default TTL on a cache is 72. I'd err towards 24 hours, but this is a weakly held weak opinion.

Either way, happy for this to just be a value we hard-code rather than using a constant etc. It's very unlikely it'll need much if any tuning.

spec/services/cache_spec.rb
```diff
@@ -80,4 +84,41 @@
...
+      cache = described_class.new(enabled: anything, ttl: anything)
+      redis = cache.redis_cache
+
+      redis.set("contentful:entry:123", "value")
+      expect(redis).to receive(:expire).with("contentful:entry:123", (60 * 60 * 48) + -1)
+
+      cache.extend_ttl_on_all_entries
+    end
+
+    context "when other redis keys exist" do
```
jacksonj04, line 99: Good test!

jacksonj04 (approved):

app/services/cache.rb
tahb: I was torn between 24 and 48 so happy to go with 24 👍

PR #84, opened 2020-12-16
(225) Refactor Steps so they now include an Contentful entry ID as a unique field

tahb:
This change is a prior refactor ahead of the linking of Journeys to the [creation of their Specification](https://trello.com/c/5plFnXEs). We know that each Step will need to be linked to another Contentful Entry as part of this work.

I imagine that for a Journey we would:

1. Get the Specification Entry
2. Loop through each of the linked Specification Block Entries
3. For each Block Entry, find an associated Step
4. Add it to a new document based on the answer

Step 3 is what this pull request seeks to make easier. Instead of having to go into the `raw` attribute of each step we can do something like: `journey.steps.find(contentful_id: block_entry_id)`

## Changes in this PR

- a new `contentful_id` column exists on Steps
- we were storing the Contentful ID in the raw attribute. We can use this to migrate all existing records so that they have this field
- change the column type from `binary` to `jsonb` so we can recieve a hash when we access `raw`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog 1.0.0].
 - users can be asked to provide multiple answers via a checkbox question
 - journey map shows an error to the content team if a duplicate entry is detected
 - journey map shows an error to the content team if the journey doesn't end within 50 steps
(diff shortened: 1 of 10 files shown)


jacksonj04 (approved):
Useful change!

PR #104, opened 2021-01-11
(225) Specification for a journey is shown as HTML

tahb:
## Changes in this PR

This is the first change to the codebase that starts to use the answers to the given questions to produce a specification document.

It is the first piece in what we imagine might be a series of changes. We would like the way specification documents are ordered, their show/hide logic and content to be controlled flexibly through Contentful. Right now, we are still working away on questions so the simplest first step was to prove we could do it when the Liquid was sourced from a local file. This reduces flexibility but enables us to start downloading documents to test with. It's imagined that we will eventually shed these files and replace them with calls to the Contentful API. 

Currently we only have support for text and radio answers so I've picked those from the "Menus and Ordering" section from the documentation the content team have been preparing: https://miro.com/app/board/o9J_lb34sdM=/?moveToWidget=3074457352505653106&cot=14 and https://docs.google.com/document/d/1LAxX2ysWk3cBn3LOjkdjeKhx4S28p4pOSlk0EG2FLkU/edit#

- we are starting with a basic HTML view of the spec, before moving onto other formats like .docx and .pdf
- we use Liquid as a templating engine based on it's simple (relative to ERB) syntax which we believe the content team could pick up. The Ruby gem is also maintainted by Shopify and appears a mature and stable choice
- the specification is rendered on the journey#show page for now as it was easy to implement given that's where the user ends their journey, it may have a different home in further iterations
- we hard code Contentful entry IDs into the view templates, this introduces a problem because each environment uses a different Contentful environment. Each Contentful environment has unique entry IDs for its' steps. We will have to come back and create a `catering.production.liquid` when the entry IDs are all in place
- to handle the case where specifications templates could be changed whilst a 'journey' was being progressed, we take a shallow copy of the template at the point the journey is started. This ensures the rendering of the spec should always be compatible with the questions presented to the user. It takes us toward our strategy of a forward facing world where users can build specifications but if they want the latest and greatest, they will get those changes in their next transaction. The dev team had a conversation about this before starting this work to agree that this was a good enough first pass
- apply basic styling to HTML elements that appear in Liquid templates

[This Liquid approach was first spiked in this branch](https://github.com/DFE-Digital/buy-for-your-school/tree/spike/create-specifications).

## Screenshots of UI changes

Screenshot_2021-01-11 Catering

## Next steps

- wipe the test journeys from the staging and prod database to enable the migration to work
- create a new `catering.production.liquid` file when the new entries are ready


spec/factories/journey.rb
```diff
@@ -2,6 +2,7 @@
   factory :journey do
     category { "catering" }
     next_entry_id { "47EI2X2T5EDTpJX9WjRR9p" }
+    liquid_template { "Your answer was { answer_47EI2X2T5EDTpJX9WjRR9p }" }
```
LornaHarwood, line 5: These internal curly braces should be double `{{ }}`

spec/features/visitors/anyone_can_see_a_planning_start_page_spec.rb
```diff
@@ -1,3 +1,5 @@
+require "rails_helper"
```
LornaHarwood, line 1: I'm not clear what this is for, could you explain please?

LornaHarwood (approved):
I have left a comment regarding a typo and I have also left a question comment too. Other than this it all looks good to me 👍

spec/features/visitors/anyone_can_see_a_planning_start_page_spec.rb
tahb: This test started failing after I made changes to the stylesheets. It was _only_ this feature test that was failing due to an newly added styling being missing, odd. I figured that there was something unique about this test (that I wasn't working on) that was different to the rest in such a way that the styles were not loaded correctly.

I spotted that we weren't included the rails_helper as all other tests do. Adding this back in should run the same Rails configuration required to load the assests and make them available correctly.

tahb:
@LornaHarwood @jacksonj04 [I added a new commit](https://github.com/DFE-Digital/buy-for-your-school/pull/104/commits/e475a566b66e18945d85848184a7f8233dcb272f) here after noticing I wasn't doing any error handling in terms of receiving invalid Liquid.

As explained in the commit: I figured it would be best to check the validity of the Liquid template when the user attempts to start the journey. Failing loudly and sending another error to Rollbar. This is a blunt tool but ensures that we never have real journeys that are stuck in progress with invalid liquid templates that can never render.

jacksonj04 (approved):
Taken a look at the new error handling, looks good to me!

tahb:
The next step of adding the production template is going in because the entries have been created and are to use, just in time  🙌

PR #105, opened 2021-01-11
(280) Load all questions at the beginning of a journey, and use task list pattern

LornaHarwood:
## Changes in this PR

- questions are now all loaded at the start of a journey, rather than step by step
- users start the journey at the task list instead of the first question
- check your answers pattern has been replaced by a task list

### Before
Screen Shot 2021-01-18 at 15 33 05
### After
Screen Shot 2021-01-18 at 15 29 59

Screen Shot 2021-01-18 at 15 30 42

## Next steps
- We want to only display the specification when all tasks have been completed
- Remove remaining used code
- Implement sections with the task list
- Implement grouping multiple questions into one task
- Disable the preview endpoint unless `ENV["PREVIEW_APP"]` is `true`
- Retrieve a complete list of questions from a `Category` entry, instead of chaining steps


app/controllers/journeys_controller.rb
```diff
@@ -11,15 +11,18 @@ def new
   end
 
   def show
+    Bullet.enable = false
```
erbridge, line 26: Why are we disabling bullet here?

app/assets/stylesheets/application.scss
```diff
@@ -3,6 +3,7 @@ $govuk-image-url-function: 'image-url';
 
 @import "base/frontend";
 @import "components/specification";
+@import "components/_task-list";
```
erbridge, line 6: Underscore should be unnecessary in the import. IF you're going to add it, I suggest being consistent with `_specification`, too.

app/assets/stylesheets/components/_task-list.scss
```diff
@@ -0,0 +1,69 @@
+// taken from https://github.com/alphagov/govuk-prototype-kit/blob/master/app/assets/sass/patterns/_task-list.scss
```
erbridge, line 1: Reference the exact commit you pulled it from, not `master`, so you can tell if it needs updating.

app/controllers/journeys_controller.rb
```diff
@@ -18,7 +30,7 @@ def show
     @steps = @journey.steps.map { |step| StepPresenter.new(step) }
 
     @answers = @journey.steps.that_are_questions.each_with_object({}) { |step, hash|
-      hash["answer_#{step.contentful_id}"] = step.answer.response.to_s
+      hash["answer_#{step.contentful_id}"] = step.answer&.response.to_s
```
tahb, line 33: Makes sense for now. I can see you've added a next step that says you'd like to only generate the spec when all questions have been answered 👍

spec/features/visitors/anyone_can_complete_a_journey_spec.rb
```diff
@@ -3,35 +3,36 @@
...
-    visit root_path
+    stub_get_contentful_entries(
+      entry_id: "5kZ9hIFDvNCEhjWs72SFwj",
+      fixture_filename: "closed-path-with-multiple-example.json"
+    )
+    journey = CreateJourney.new(category: "catering").call
+    step = journey.steps.find_by(contentful_id: "hfjJgWRg4xiiiImwVRDtZ")
 
-    click_on(I18n.t("generic.button.start"))
+    visit journey_step_path(journey, step)
```
tahb, line 35: I agree with cutting straight to the step path so long as we continue to have our first test here which exercises clicking the 'Start' button 👍

spec/features/visitors/anyone_can_complete_a_journey_spec.rb
```diff
@@ -40,59 +41,25 @@
...
       scenario "user can answer using free text" do
-        stub_get_contentful_entry(
-          entry_id: "hfjJgWRg4xiiiImwVRDtZ",
-          fixture_filename: "short-text-question-example.json"
+        stub_get_contentful_entries(
+          entry_id: "5kZ9hIFDvNCEhjWs72SFwj",
+          fixture_filename: "closed-path-with-multiple-example.json"
         )
+        journey = CreateJourney.new(category: "catering").call
+        step = journey.steps.find_by(contentful_id: "contentful-short-text-question")
```
tahb, line 60: I was expecting the `contentful_id` here to be `5kZ9hIFDvNCEhjWs72SFwj`. How does this work without it?

spec/features/visitors/anyone_can_complete_a_journey_spec.rb
```diff
@@ -247,40 +216,45 @@
...
         example.run
       end
     end
 
     scenario "returns an error message" do
-      stub_get_contentful_entry(
-        entry_id: "8as7df68uhasdnuasdf",
-        fixture_filename: "an-unexpected-question-type-example.json"
+      stub_get_contentful_entries(
+        entry_id: "contentful-unexpected-step-type",
```
tahb, line 223: Ah I see now we are faking the ID in the fixture so it has a more human readable but less realistic value? This seems inconsistently applied in this commit. If I were to add a new test I would be like to feel confident in which approach to take, human readable or realistic.

app/controllers/preview/entries_controller.rb
```diff
@@ -1,7 +1,12 @@
 class Preview::EntriesController < ApplicationController
   def show
-    @journey = Journey.create(category: "catering", next_entry_id: entry_id)
-    redirect_to new_journey_step_path(@journey)
+    @journey = Journey.create(category: "catering")
```
tahb, line 3: This change makes me think of another issue. The impact of creating a journey with _all_ steps for any request to preview. Specifically that it would pollute the database with a lot of information that is never used again. I know Contentful is configured to only use the dedicated Preview environment with separate database, but what do you think about making sure we prevent this feature working on the production environment, perhaps with ENV, in a separate piece of work?

app/views/steps/paragraphs.html.erb
```diff
@@ -4,4 +4,4 @@
 <div class="static-content">
   <%= simple_format(@step.body, wrapper_tag: "p", class: "govuk-body") %>
 </div>
-<%= link_to @step.primary_call_to_action_text, new_journey_step_path, class: "govuk-button" %>
+<%= link_to @step.primary_call_to_action_text, journey_path, class: "govuk-button" %>
```
tahb: A little surprised this works without having to specify which `@journey` we'd like to go and view. Is it being clever and taking `params[:id]`?

app/controllers/journeys_controller.rb
jacksonj04: I've added a next step to find a better way to tackle this, but the PR was getting huge and unwieldy and my brain couldn't find a neat solution.

When we query to get the `Journey` and all its `Step`s on line 28, we include all of the possible answer types which any `Step` might have. If we don't do this for an answer type which we use in the journey, then Bullet tells us to use eager loading when we hit up `.answer` on line 32.

The problem then becomes that when we visit the task list, unless it contains an example of every type of answer, Bullet will complain that we've been unnecessarily eager loading (eg in tests such as `anyone_can_edit_their_answers_spec.rb:25`).

The only way we can include only the relevant types of answer programmatically is by intentionally doubling up on the query, ie once to get a list of all the question types, and then only including the types of answer we know we might have.

The alternative would be to make sure that any time a factory spits out a `Journey` it includes one of each type of question.

app/views/steps/paragraphs.html.erb
```diff
@@ -4,4 +4,4 @@
 <div class="static-content">
   <%= simple_format(@step.body, wrapper_tag: "p", class: "govuk-body") %>
 </div>
-<%= link_to @step.primary_call_to_action_text, journey_path, class: "govuk-button" %>
+<%= link_to @step.primary_call_to_action_text, journey_path(@journey), class: "govuk-button" %>
```
tahb, line 7: Ah here it is. I think this belongs in the previous commit where it was removed.

spec/features/visitors/anyone_can_complete_a_journey_spec.rb
```diff
@@ -64,7 +64,7 @@
         fill_in "answer[response]", with: "email@example.com"
         click_on(I18n.t("generic.button.next"))
 
-        expect(page).to have_content("email@example")
+        expect(step.answer.response).to eql("email@example.com")
```
tahb, line 67: I would suggest that if we are making the user click into each question to check their answer, that we immitate that behaviour here but testing that it appears on the `page` after the additional click. Rather than only having the feature test check the database which I think is less valuable.

tahb: I think this applies to most of the tests in this commit.

tahb (commented):
Looks and works great :+1: 

I have some comments for minor things, a big one about testing the database rather than the page and couple of discussions for future work.

tahb:
Another observation. This actually replaces the need for journey_map/new doesn't it?

app/views/journeys/show.html.erb
```diff
@@ -4,7 +4,15 @@
 <dl class="govuk-summary-list">
   <% @steps.each do |step| %>
     <% if step.question? %>
-      <%= render "#{step.contentful_type}_answer", step: step, answer: step.answer %>
+      <% if step.answer %>
+        <%= render "#{step.contentful_type}_answer", step: step, answer: step.answer %>
+      <% else %>
+      <div class="govuk-summary-list__row">
```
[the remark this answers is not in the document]
erbridge: My past investigations into ERB linting have had no luck. I'd love to find one, though.

app/controllers/preview/entries_controller.rb
jacksonj04: Thought: our next move (or one reasonably soon) is moving away from chained steps to a list of steps. We could tweak the process of creating a journey to accept this list of steps in order, and then for preview we just create a journey and pass it a single step?

app/controllers/journeys_controller.rb
jacksonj04: Solved this using the whitelist pattern. TIL.

spec/fixtures/contentful/multiple-entries-example.json
```diff
@@ -92,14 +92,7 @@
                   {
                       "value": "Cleaning"
                   }
-                ],
```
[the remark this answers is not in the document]
jacksonj04: Removing the `next` here is intentional, as it refers to a non-existing step within this journey (and as such throws a `BuildJourneyOrder::MissingEntryDetected`)

app/views/journeys/show.html.erb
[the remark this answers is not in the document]
jacksonj04: This HTML is overwritten by our later styling fixes to match the task list pattern in the prototype kit.

app/controllers/preview/entries_controller.rb
tahb: Yes that would limit the 'pollution' in such a way that we were back where we were before this change, which would be good. Better and perhaps easier would be to only allow this endpoint to work if `ENV["PREVIEW_APP"]` is true.

Either way, a problem for another day I think.

tahb (approved):
Superb efforts 🙌  Let's :shipit:

PR #107, opened 2021-01-12
(259) Users can download their specification as a .docx document

tahb:
This work is based on https://github.com/DFE-Digital/buy-for-your-school/pull/104 which should be merged first.

## Changes in this PR

This work introduces a new download button at the end of the journey. The user should then receive the same specifcation contents that are already shown on the page in HTML, in the new document in .docx format. The user should be able to open this in Microsoft Word to make further changes.

- we use a html2word gem that sets the MIME type behind the scenes and allows us to write `format.docx` which otherwise appears rather magically

## Screenshots of UI changes

Screenshot_2021-01-12 Catering
Screenshot 2021-01-12 at 16 22 30
Screenshot 2021-01-12 at 16 34 18


app/controllers/journeys_controller.rb
```diff
@@ -23,6 +23,15 @@ def show
     @specification_template = Liquid::Template.parse(
       @journey.liquid_template, error_mode: :strict
     )
+
+    specficiation_html = @specification_template.render(@answers)
```
jacksonj04, line 27: This should be `specification_html`

app/controllers/journeys_controller.rb
```diff
@@ -23,6 +23,15 @@ def show
...
+
+    respond_to do |format|
+      format.html
+      format.docx do
+        render docx: "specification.docx", content: specficiation_html, layout: "specficiation"
```
jacksonj04, line 32: The other half of the typo.

jacksonj04 (approved):
Nice and clean, hooray for gems which already do the heavy lifting! Happy for this to go once  #104 is merged.

I've added [a card](https://trello.com/c/SvhVbvqq/287-add-html-validation-for-schema-templates) to cover the need for validating the HTML of the schema.

PR #110, opened 2021-01-14
(288) Radio question options can be provided by a new extended JSON format

tahb:
## Changes in this PR

This work starts the transitional work to migrate the existing options field that's required for radio and select questions to work. We are going from an array of strings to a hash of various options. For now we add a simple "value" to ensure this work remains a refactoring exercise.

- radio questions should behave exactly the same with existing entries that use the `Options` field 
- where `Extended Options` are present, they take precendent and any `Options` are ignored

## Next steps

The plan is to:

1. switch over all Contentful entries so that they have the option new format ✅ Dev ✅ Prod
2. deploy these changes
3. remove `Options` from the Content model called `questions`
4. rename `Extended Options` to `Options` in Contentful

Once that is done we have further work to do. We need to:

1. work out what to do with existing Steps that exist in databases with an Option column that has data in the old format (hopefully we can throw it all away)
2. set the new flavour of options on the `options` field when we create steps (rather than going via raw)
3. tidy up our transitional methods added here


LornaHarwood (approved):
This looks good to me.
I can understand from your clear description what this PR is trying to achieve.
From this, I can understand and follow what the code is doing and that it relates to the scope of what you have described within the PR.

tahb:
Given we have already discussed this approach I've gone through both Contentful environments and added the new `options` field and copied over the existing options. I have updated both Faye and Quay that they will see 2 fields for a time and what to do about them.

PR #113, opened 2021-01-15
Docker Compose server start auto purges any leftover pid file

tahb:
When we start the webserver it creates a pid file in tmp/.

When running the server locally with docker compose and you turn off your machine without gracefully terminating the server process the file remains.

When you start the server again you get an error asking you to remove the existing pid file.

This change automates this step so we don't have to do this manual step each time.

It doesn't affect production environments.

# Screenshots
## Before
Screenshot 2021-01-15 at 11 23 22


LornaHarwood (approved):
Looks good to me 👍

PR #114, opened 2021-01-18
(288) Only support Contentful 'ExtendedOptions' (dropping 'Options' support)

tahb:
## Changes in this PR

`Options` is no longer expected as a field that comes back from contentful.

In all cases for select and radio questions we now expect a field with the ID `ExtendedOptions`.

Within the codebase, above the handling of the Contentful response, we continue to use the existing `Options` language.

We recently merged in this change that introduced support for using both mechanisms at the same time https://github.com/DFE-Digital/buy-for-your-school/pulls. This change now removes the support for legacy option configuration and tidies up.


LornaHarwood (approved):
All looks good 👍

PR #115, opened 2021-01-18
(185) Users can be asked Radio questions that prompt for additional text

tahb:
## Changes in this PR

To review locally you can use Entry ID `QvmpHjUCDyx0EofNDn9hX` in your .env as the starting entry.

Use this [1] pattern from the DFE Form builder gem. We take the behaviour that is applied to the 'Other' option in the example and make it applicable to any radio option, or none.

This change also includes a refactoring of teh way we present answers to different answer types. Moving the logic from the view into tested presenters. These commits are on the end and can be easily removed if controversial in any way. The goal is to chip away at a secondary problem in passing.

- Contentful always supplies JSON for a list of options. It is now possible to provide more than a simple list of values in thei JSON blob. If the Content user provides keys `display_further_information` and `further_information_help_text` then the question will render an extra text input. This is per option and is not required.
- This new text input is then shown back as a composite string value to the user at the end of the journey.
- The field is a text area rather than a field as the content team wanted the user to drag the text area bigger if a longer answer was necessary.
- We think we may want to toggle validation through configuration in future. For now these text fields are optional and won't fire validation errors.

[1] https://govuk-form-builder.netlify.app/form-elements/radios/#radio-buttons-fieldset-with-conditional-content-and-a-divider

## Screenshots of UI changes

Screenshot 2021-01-18 at 14 18 24
Screenshot 2021-01-18 at 14 19 25
Screenshot 2021-01-18 at 14 20 18
Screenshot 2021-01-18 at 14 20 10

### DIFF
--- a/app/views/journeys/_radios_answer.html.erb
+++ b/app/views/journeys/_radios_answer.html.erb
@@ -1,5 +1,5 @@
 <div class="govuk-summary-list__row">
   <dt class="govuk-summary-list__key"><%= step.title %></dt>
-  <dd class="govuk-summary-list__value"><%= answer.response.capitalize %></dd>
+  <dd class="govuk-summary-list__value"><%= answer.response.capitalize %><%= " - #{answer.further_information}" if answer.further_information.present? %></dd>
   <dd class="govuk-summary-list__actions"><%= link_to I18n.t("generic.button.change_answer"), edit_journey_step_path(@journey, step), class: "govuk-link" %></dd>
 </div>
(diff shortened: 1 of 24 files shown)


app/views/journeys/_radios_answer.html.erb
```diff
@@ -1,5 +1,5 @@
 <div class="govuk-summary-list__row">
   <dt class="govuk-summary-list__key"><%= step.title %></dt>
-  <dd class="govuk-summary-list__value"><%= answer.response.capitalize %></dd>
+  <dd class="govuk-summary-list__value"><%= answer.response.capitalize %> <%= " - #{answer.further_information}" if answer.further_information.present? %></dd>
```
LornaHarwood: The space between the response and further information could be removed as it isn't needed.

LornaHarwood (approved):
All looks good to me :)

app/views/journeys/_radios_answer.html.erb
tahb: Thanks! This line is replaced in a later commit but a presenter but I've tidied up the commit history 👍

tahb:
I'm planning to hold this and let https://github.com/DFE-Digital/buy-for-your-school/pull/105 be merged first given th conflict potential.

PR #116, opened 2021-01-18
release-005

tahb:
## Changes in this PR

- users can see an initial slice of their specification as HTML at the end of their journey
- users can download their specification as a document in the .docx format
- checkbox answers are editable
- radio and select questions are created with the new `ExtendedOptions` field
- questions are now all loaded at the start of a journey, rather than step by step
- users start the journey at the task list instead of the first question
- check your answers pattern has been replaced by a task list
- radio questions can be configured to ask the user for additional text
- update the service name to the latest decision

## Next steps

Production Journeys all need to be destroyed for the migrations to succeed. Do so just before we deploy with `Journey.destroy_all`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog 1.0.0].
 
 ## [Unreleased]
 
+## [release-005] - 2021-1-19
+
 - users can see an initial slice of their specification as HTML at the end of their journey
 - users can download their specification as a document in the .docx format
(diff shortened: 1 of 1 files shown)


LornaHarwood (changes requested):
This needs the release link within the foot notes

tahb:
@LornaHarwood Good spot, thank you. I've added the bottom bit too 👍

jacksonj04 (approved):
📦 

Let's do this thing!

PR #117, opened 2021-01-19
Fix radio question label size inconsistency

tahb:
## Changes in this PR

- make radio questions "XL" again inline with all other question types
- reduce all question label size down one notch to "L" given the length of real questions we are starting to see

## Screenshots of UI changes

Radio questions had very form small labels/page title:

Screenshot 2021-01-19 at 11 06 37

Radio questions now have slightly bigger labels:
Screenshot 2021-01-19 at 11 05 54

Real questions looked _really_ big:
Screenshot 2021-01-19 at 10 56 45

We make them slightly smaller:
Screenshot 2021-01-19 at 11 07 57

### DIFF
--- a/app/views/steps/checkboxes.html.erb
+++ b/app/views/steps/checkboxes.html.erb
@@ -3,6 +3,6 @@
     checkbox_options(array_of_options: @step.options_list),
     :id,
     :name,
-    legend: { text: @step.title, size: "xl" }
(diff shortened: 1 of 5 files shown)


jacksonj04 (approved):
:shipit: 

Agree, the balance of size looks good. If we end up with lots of question types in future we might want to make these constants, but that feels like overkill for now.

PR #120, opened 2021-01-19
Responses used in Specifications are formatted by presenters

tahb:
## Changes in this PR
This fixes an immediate issue we can see where radio answers show "Yes" without the `futher_information`. It should be "Yes - Further information" and the logic already exists within presenters.

We need to get a fix for this quickly to unblock the production deploy. I've left a note to prompt us to refactor this controller code the next time we pass it.

### Before
Screenshot 2021-01-19 at 14 34 58

### After

The last answer is now a composite value.
Screenshot 2021-01-19 at 14 34 45

### DIFF
--- a/app/controllers/journeys_controller.rb
+++ b/app/controllers/journeys_controller.rb
@@ -28,8 +28,18 @@ class JourneysController < ApplicationController
     ).find(journey_id)
     @steps = @journey.steps.map { |step| StepPresenter.new(step) }
 
+    # TODO: Move this logic into a tested class along with a Presenter factory
     @answers = @journey.steps.that_are_questions.each_with_object({}) { |step, hash|
-      hash["answer_#{step.contentful_id}"] = step.answer&.response.to_s
+      answer = case step.answer.class.name
+      when "ShortTextAnswer" then ShortTextAnswerPresenter.new(step.answer)
+      when "LongTextAnswer" then LongTextAnswerPresenter.new(step.answer)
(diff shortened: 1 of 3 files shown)


app/controllers/journeys_controller.rb
```diff
@@ -28,8 +28,18 @@ def show
...
 
+    # TODO: Move this logic into a tested class along with a Presenter factory
     @answers = @journey.steps.that_are_questions.each_with_object({}) { |step, hash|
-      hash["answer_#{step.contentful_id}"] = step.answer&.response.to_s
+      answer = case step.answer.class.name
+      when "ShortTextAnswer" then ShortTextAnswerPresenter.new(step.answer)
+      when "LongTextAnswer" then LongTextAnswerPresenter.new(step.answer)
+      when "RadioAnswer" then RadioAnswerPresenter.new(step.answer)
+      when "SingleDateAnswer" then SingleDateAnswerPresenter.new(step.answer)
+      when "CheckboxeAnswers" then CheckboxAnswersPresenter.new(step.answer)
```
jacksonj04, line 38:
```suggestion
      when "CheckboxAnswers" then CheckboxAnswersPresenter.new(step.answer)
```

jacksonj04 (approved):
Looks good. In future we'll probably want to do work to break out the two answer components for the template, but that's a problem for down the road.

app/controllers/journeys_controller.rb
tahb: Done :+1:

PR #121, opened 2021-01-20
Previewing a question now specifies a liquid_template

jacksonj04:
`liquid_template` is a required part of a `Journey` which was not being provided whilst previewing a question, leading to the `Journey` not being persisted and a redirect to the step within the journey being impossible.

This commit adds a very short stub template to preview journeys.

It also adds a feature spec to ensure that a user attempting to preview a step does reach the intended view.

### DIFF
--- a/spec/features/visitors/anyone_can_preview_a_journey_step_spec.rb
+++ b/spec/features/visitors/anyone_can_preview_a_journey_step_spec.rb
@@ -1,0 +1,14 @@
+feature "Users can preview a journey step" do
+  scenario "the appropriate step is displayed" do
+    stub_get_contentful_entry(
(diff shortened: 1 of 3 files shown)


spec/features/visitors/anyone_can_preview_a_journey_step_spec.rb
```diff
@@ -0,0 +1,14 @@
+feature "Users can preview a journey step" do
+  scenario "user can answer using free text" do
```
tahb, line 2: This is a radio question rather than text but number of cares given is barely enough to point this out :)

tahb (approved):

PR #123, opened 2021-01-21
Pull a complete list of steps from a Category when a new Journey is created, instead of working sequentially through a list of steps

jacksonj04:
## Changes in this PR

- [x] Introduce a new GetEntriesInCategory service class, which can be passed a Category entry ID, and will return a list of Entries (in order) which make up the steps in that category's journey. This class should use GetContentfulEntry to take advantage of caching.
- [x] Refactor the CreateJourney service class (used by the JourneyController) to use this new class over BuildJourneyOrder.
- [x] Refactor the JourneyMapController to use this new service class over the existing BuildJourneyOrder.
- [x] Refactor cache warming to use this new service class over BuildJourneyOrder.
- [x] Remove BuildJourneyOrder and all references to it
- [x] Remove all references to the old CONTENTFUL_PLANNING_START_ENTRY ENV variable.
- [x] Remove new_journey_step_path (all steps are now created in CreateJourney), and refactor anything which relies on it.
- [x] Remove next_entry_id (we no longer care about what the next step in a journey is, only all the steps in order), and refactor anything which relies on it.
- [x] Remove old test fixtures which aren't in the new /steps and /categories folder structure.
- [x] Remove the /categories/radio.json fixture, in favour of /categories/radio-question.json.
- [x] Rename /steps/contentful-radio-question.json to /steps/radio-question.json for consistency.


tahb (approved):
We have all paired on this 🎉

PR #128, opened 2021-01-26
Step factories no longer create associated answers

jacksonj04:
The step factory was also creating associated answers every time a step was created. This does not mirror the behaviour of the system; answers are only created at the point a user provides one, not when a step is created.


tahb (approved):
I think this is right. 

Before hand it was faking the end state of the database which is indeed beyond it's responsibility as a factory.

PR #129, opened 2021-01-26
Hide specification until all tasks are complete

LornaHarwood:
## Changes in this PR

Specification now only shows when all steps are completed

### Before

localhost_3000_journeys_067ce62f-5326-48f5-9c01-32ad502a1d96 (1)

### After

localhost_3000_journeys_067ce62f-5326-48f5-9c01-32ad502a1d96

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog 1.0.0].
 
 ## [Unreleased]
 
+- specification is no longer displayed before all tasks are completed
+
 ## [release-005] - 2021-1-19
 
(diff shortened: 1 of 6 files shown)


tahb:
@jacksonj04 mentioned on Slack that this not be something we want to ship afterwards. I'll hold my review for now. Please poke me again if you'd like me to look at this!

tahb:
On the basis that we are (progressing work that does the opposite of this)[https://github.com/DFE-Digital/buy-for-your-school/pull/167] I think we should close for now. I've managed to take the `journey.all_steps_completed?` method and tests so this PR seems to have no more value as something open. We can always reopen it if we need it in future?

PR #132, opened 2021-01-27
Users can answer questions which expect a number

LornaHarwood:
## Changes in this PR

- Users can answer questions which expect a number
- An exception is rasied when AnswerFactory receives an unexpected question type

## Screenshots of UI changes
Screen Shot 2021-02-01 at 16 32 07

### DIFF
--- a/app/presenters/number_answer_presenter.rb
+++ b/app/presenters/number_answer_presenter.rb
@@ -1,0 +1,2 @@
+class NumberAnswerPresenter < SimpleDelegator
+end
--- a/spec/features/visitors/anyone_can_complete_a_journey_spec.rb
+++ b/spec/features/visitors/anyone_can_complete_a_journey_spec.rb
@@ -45,6 +45,37 @@ feature "Anyone can start a journey" do
       end
     end
 
+    context "when Contentful entry is of type numbers" do
+      scenario "user can answer using a number input" do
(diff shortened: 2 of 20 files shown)


jacksonj04 (changes requested):

spec/features/visitors/anyone_can_complete_a_journey_spec.rb
```diff
@@ -45,6 +45,19 @@
       end
     end
 
+    context "when Contentful entry is of type numbers" do
```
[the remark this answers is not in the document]
LornaHarwood: 👍 These have been added

app/presenters/number_answer_presenter.rb
```diff
@@ -0,0 +1,2 @@
+class NumberAnswerPresenter < SimpleDelegator
```
jacksonj04, line 1: For completeness we should add a test to confirm this behaves as intended - see `/spec/presenters` for some examples to base it on.

jacksonj04 (changes requested):
One very quick missing test, then I think this one's good to roll!

app/presenters/number_answer_presenter.rb
LornaHarwood: 💯 - I have added in a spec test for `NumberAnswerPresenter`

jacksonj04 (approved):
🚀

PR #136, opened 2021-01-28
(391) Specification template is now sourced from a Contentful Category

tahb:
## Changes in this PR

A few more changes than originally expected here. The most notable being a refactoring to how we get a Contentful Category. 

Since the Contentful Category is the new source of a specification_template we have to move our Liquid validations to a new home. I believe extracting the `GetCategory` process from `GetCategoryEntries` and switching to injecting a category (instead of `GetCategoryEntries` doing both parts) offers a natural place to move the liquid validation logic too, as well as simplify the test suite in places when we don't always care about the loading of the category.

- update all Contentful fixtures with the new required field called `specification_template`
- rename `GetContentfulEntry` into `GetEntry` to remove "Contentful", making our service names more consistent
- introduce `GetCategory` which uses `GetEntry` under the hood + liquid validations
- refactor `GetCategoryEntries` so all it does is load category steps, the `GetCategory` process happens before it and the `category` is injected down as a dependency

If you test this locally you may spot a bug where a `step.answer` that is an empty string is recognised by the Liquid `if {{ answer_x }}` as truthy. [I will fix this as a separate issue](https://github.com/DFE-Digital/buy-for-your-school/pull/137/commits/29f2cc026c75d120a12c4b88582edb770dab30c2) however the screenshots here have been produced with this fixed locally (so it looks right and tells the intended story).

## Screenshots of UI changes

Here is the the new field in Contentful:
Screenshot 2021-01-28 at 15 09 11

An empty spec with no answered questions (not loading the spec from the old hardcoded files):
Screenshot 2021-01-28 at 15 16 23

Answering a question continues to inject the answer into the page via the Liquid templating process:
Screenshot 2021-01-28 at 15 16 40


spec/support/contentful_helpers.rb
```diff
@@ -13,12 +13,12 @@ def stub_contentful_entry(
...
     fixture_filename:,
     stub_steps: true,
-    contentful_connector: instance_double(ContentfulConnector)
+    contentful_connector: instance_double(ContentfulConnector) # TODO: I suspect the double doesn't do anything and we need stub_contentful_connector
   )
     category = fake_contentful_category(contentful_fixture_filename: fixture_filename)
 
     expect(ContentfulConnector).to receive(:new)
-      .and_return(contentful_connector).at_least(:once) # This twice is a problem..
+      .and_return(contentful_connector).once
```
jacksonj04, line 21: Whoop!

spec/support/contentful_helpers.rb
```diff
@@ -29,9 +29,20 @@ def stub_contentful_category(
...
-  def stub_contentful_category_steps(category:, contentful_connector:)
+  def stub_contentful_category_steps(
+    category:,
+    contentful_connector: instance_double(ContentfulConnector)
+  )
+    return if category.steps.count.zero?
+
+    expect(ContentfulConnector).to receive(:new)
+      .and_return(contentful_connector)
+      .at_least(category.steps.count)
```
jacksonj04, line 44: Should this be `exactly`, given it makes one call per step and we know exactly how many steps there are?

app/services/create_journey.rb
```diff
@@ -1,17 +1,18 @@
 class CreateJourney
-  attr_accessor :category
+  attr_accessor :category_name
```
jacksonj04, line 2: This is much nicer, 👍🏼

spec/features/visitors/anyone_can_download_their_catering_specification_spec.rb
```diff
@@ -1,10 +1,14 @@
 feature "Users can see their catering specification" do
   scenario "HTML" do
-    journey = create(:journey, :catering, liquid_template: stub_liquid_template)
-    step = create(:step, :long_text, long_text_answer: nil, journey: journey, contentful_id: "NxJWpbiFeEAmvcw17EysX")
-    _answer = create(:long_text_answer, step: step, response: "Red tractor")
+    stub_contentful_category(fixture_filename: "category-with-liquid-template.json")
```
jacksonj04, line 3: This (down to `:7`) could be replaced with a `start_journey_from_category_and_go_to_question("category-with-liquid-template.json")`

spec/features/visitors/anyone_can_see_their_catering_specification_spec.rb
```diff
@@ -1,34 +1,39 @@
 feature "Users can see their catering specification" do
   scenario "HTML" do
-    liquid_template = stub_liquid_template(filename: "food_catering.liquid")
-    journey = create(:journey, :catering, liquid_template: liquid_template)
-    step = create(:step, :long_text, long_text_answer: nil, journey: journey, contentful_id: "NxJWpbiFeEAmvcw17EysX")
-    answer = create(:long_text_answer, step: step, response: "Red tractor")
+    stub_contentful_category(fixture_filename: "category-with-dynamic-liquid-template.json")
```
jacksonj04, line 3: Another candidate for `start_journey_from_category_and_go_to_question`.

jacksonj04 (approved):
A couple of little tweaks to reduce boilerplate, and one test I think we can be more explicit on, but this is all Good Stuff!

spec/support/contentful_helpers.rb
tahb: Yes it should be!

tahb:
All addressed. Thanks for the feedback 🙌

PR #137, opened 2021-01-28
Refactor answer finding into a tested service

tahb:
## Changes in this PR

- straight refactor to move some heavy controller code into a class with unit test coverage

### DIFF
--- a/app/services/find_all_journey_answers.rb
+++ b/app/services/find_all_journey_answers.rb
@@ -18,7 +18,8 @@ class FindAllJourneyAnswers
       when "CheckboxAnswers" then CheckboxesAnswerPresenter.new(step.answer)
       else raise UnexpectedAnswer.new("Trying to present an unknown type of answer: #{step.answer.class.name}")
       end
-      hash["answer_#{step.contentful_id}"] = answer&.response.to_s
+
+      hash["answer_#{step.contentful_id}"] = answer.response.to_s
     }
   end
 end
--- a/app/services/find_all_journey_answers.rb
+++ b/app/services/find_all_journey_answers.rb
@@ -1,0 +1,24 @@
+class FindAllJourneyAnswers
+  class UnexpectedAnswer < StandardError; end
+
+  attr_accessor :journey
+  def initialize(journey:)
+    self.journey = journey
+  end
+
+  def call
+    @answers = @journey.steps.that_are_questions.each_with_object({}) { |step, hash|
+      next unless step.answer
+
+      answer = case step.answer.class.name
+      when "ShortTextAnswer" then ShortTextAnswerPresenter.new(step.answer)
+      when "LongTextAnswer" then LongTextAnswerPresenter.new(step.answer)
+      when "RadioAnswer" then RadioAnswerPresenter.new(step.answer)
+      when "SingleDateAnswer" then SingleDateAnswerPresenter.new(step.answer)
+      when "CheckboxAnswers" then CheckboxesAnswerPresenter.new(step.answer)
+      else raise UnexpectedAnswer.new("Trying to present an unknown type of answer: #{step.answer.class.name}")
+      end
+      hash["answer_#{step.contentful_id}"] = answer&.response.to_s
+    }
+  end
+end
--- a/spec/services/find_all_journey_answers_spec.rb
+++ b/spec/services/find_all_journey_answers_spec.rb
@@ -1,0 +1,69 @@
+require "rails_helper"
+
+RSpec.shared_examples "returns the answer in a hash" do |factory_name, presenter, response|
+  it "returns the response for a #{factory_name} in the result" do
+    answer = create(factory_name, response: response)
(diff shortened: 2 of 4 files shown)


app/services/find_all_journey_answers.rb
```diff
@@ -6,14 +6,17 @@ def initialize(journey:)
...
     @answers = @journey.steps.that_are_questions.each_with_object({}) { |step, hash|
+      next if step.answer.nil?
+
       answer = case step.answer.class.name
       when "ShortTextAnswer" then ShortTextAnswerPresenter.new(step.answer)
       when "LongTextAnswer" then LongTextAnswerPresenter.new(step.answer)
       when "RadioAnswer" then RadioAnswerPresenter.new(step.answer)
       when "SingleDateAnswer" then SingleDateAnswerPresenter.new(step.answer)
       when "CheckboxAnswers" then CheckboxesAnswerPresenter.new(step.answer)
       end
```
jacksonj04, line 17: Could we slip an error case in here ([see similar](https://github.com/DFE-Digital/buy-for-your-school/pull/132/commits/04a440e43bac8f08a19da057a2518de30c387c90)) as we had to spend 10 minutes chasing this down whilst adding a number, and an explicit error would be handy

app/services/find_all_journey_answers.rb
```diff
@@ -6,14 +6,17 @@ def initialize(journey:)
 
   def call
     @answers = @journey.steps.that_are_questions.each_with_object({}) { |step, hash|
+      next if step.answer.nil?
```
jacksonj04, line 9: Could we do `next unless step.answer`? It seems more expressive, but it's a weakly held view.

jacksonj04 (commented):
Couple of lil' comments.

spec/services/find_all_journey_answers_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    context "when the answer is of type radio_answer" do
+      it_behaves_like "returns the answer in a hash", :radio_answer, RadioAnswerPresenter, "Catering"
+    end
+
+    context "when the answer is of type checkbox_answers" do
+      it_behaves_like "returns the answer in a hash", :checkbox_answers, CheckboxesAnswerPresenter, "Foo, Bar"
+    end
+
+    context "when a step does not have an answer" do
```
jacksonj04, line 50: There's an extra case we should test here for "when the step has an answer of an unknown type", and it should either return the bare `Answer` or (I think more robustly) raise an exception.

spec/services/find_all_journey_answers_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      expect(result).to include(
+        {"answer_#{relevant_answer.step.contentful_id}" => "Red"}
+      )
+      expect(result).not_to include(
+        {"answer_#{irrelevant_answer.step.contentful_id}" => "Blue"}
+      )
+    end
+
+    context "when the answer is of type short_text_answer" do
+      it_behaves_like "returns the answer in a hash", :short_text_answer, ShortTextAnswerPresenter, "Red"
```
jacksonj04, line 31: `"returns the answer in a hash"` looks like a copy/paste oversight!

app/services/find_all_journey_answers.rb
tahb: Good idea! 👍

jacksonj04 (changes requested):

spec/services/find_all_journey_answers_spec.rb
tahb: What do you think: "returns the answer against its step ID in a hash"?

jacksonj04: I'd be tempted to use something like "returns the answer as a string in a hash", "returns the answer as HTML paragraphs in a hash", "returns the answer as a formatted date in a hash" etc since this is the only place we also test that the formatting of the answer is as expected.

tahb: "returns the answer as a string in a hash" sounds good to me.
 
We do have unit tests for the presenters so here we could settle for checked the right presenter is used and trust them to do their job from there? Otherwise, we could repeat our tests if we feel like we need more confidence.

tahb: Oh do you mean different descriptions for each test!? Ah, so the reason I'm being abstract is so I can use the same shared example to reduce repitition. Happy to remove them ?

jacksonj04: Ohhhhh, it's not a description is it, it's the name of the pattern!

Ignore me, not enough coffee has been applied to brain.

jacksonj04 (approved):
Brain no do code. Ship it!

spec/services/find_all_journey_answers_spec.rb
tahb: Roger! 🙌

PR #144, opened 2021-02-02
Add rspec config to allow use of --only-failures command

LornaHarwood:
Added rspec config so that `--only-failures` and `--next-failure` commands can now be used


jacksonj04 (approved):
\o/

Looks good (and helpful) to me.

PR #145, opened 2021-02-02
Tell a user how they can recover their current journey

jacksonj04:
## Changes in this PR

- Users are now told how they can resume a journey (either by bookmarking a page to return to, or revisiting the same address) and an address is provided for this on the journey task list. This functionality has always been available, just never promoted.
- `robots.txt` now excludes all journey and step pages. Although these are (currently) accessible by anyone, they should not be easily discoverable.

### Before

localhost_3000_journeys_06021d31-a436-4e1e-86a2-04d162c76126 (1)

### After

localhost_3000_journeys_06021d31-a436-4e1e-86a2-04d162c76126

### DIFF
--- a/app/views/journeys/show.html.erb
+++ b/app/views/journeys/show.html.erb
@@ -25,6 +25,8 @@
   </li>
 </ol>
 
+<%= render "resume_journey_notice" %>
+
 <h1 class="govuk-heading-l"><%= I18n.t("journey.specification.header") %></h1>
 <%= link_to "Download (.docx)", journey_path(@journey, format: :docx), class: "govuk-button" %>
 <%= @specification_template.render(@answers).html_safe %>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -45,6 +45,10 @@ en:
       completed: Completed
   journey_map:
     page_title: "Contentful entry map"
+  resume:
+    notification:
+      title: "Returning to this specification"
+      body: "You can return to this page to make changes or view your specification at any time, either by bookmarking this page in your browser or by making a note of the following address:"
   errors:
     contentful_entry_not_found:
       page_title: "An unexpected error occurred"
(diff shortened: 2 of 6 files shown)


app/views/journeys/show.html.erb
```diff
@@ -25,6 +25,22 @@
   </li>
 </ol>
 
+<div class="govuk-notification-banner" role="region" aria-labelledby="govuk-notification-banner-title" data-module="govuk-notification-banner">
+  <div class="govuk-notification-banner__header">
+    <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">
```
erbridge, line 30: Maybe give it a unique ID in case we add more banners to the page. Something like `recovery-advice-banner`?

app/views/journeys/show.html.erb
```diff
@@ -25,6 +25,22 @@
   </li>
 </ol>
 
+<div class="govuk-notification-banner" role="region" id="govuk-recovery-advice-banner" aria-labelledby="govuk-recovery-advice-banner-title" data-module="govuk-notification-banner">
```
tahb, line 28: If you wanted to this seems like a great candidate for moving to a view partial. Happy to do it later as I think the view remains just about manageable.

config/locales/en.yml
```diff
@@ -45,6 +45,10 @@ en:
       completed: Completed
   journey_map:
     page_title: "Contentful entry map"
+  resume:
+    notification:
+      title: "Returning to this specification"
+      body: "You can return to this specification at any time, either by bookmarking this page in your browser or by making a note of the following address:"
```
tahb, line 51: We could also include a prompt for users to use this mechanism as a way to share between colleagues. As in "this is the intended way for now, don't look for another share link/button". Not a blocker.

public/robots.txt
```diff
@@ -1 +1,3 @@
 # See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
+
+Disallow: /journeys/
```
tahb, line 3: Well remembered!

tahb (approved):
Some suggestions. Nothing blocking 🦾

PR #146, opened 2021-02-02
(305) Task list items are grouped by new sections

tahb:
## Changes in this PR

A lot of changes here, I think I have broken it down so it's less scary that it might first appear!

- In the first 5 commits we do some tidying up refactors to make the new changes easier. We also update the Contentful fixtures so they include sections *as well as* the old steps so they are cross compatible (We remove these later)
- We then add new services whose job it will be to handle the extra requests for data from Sections. We don't plug these in yet to keep the change small. We add them alongside the existing code and include their unit tests
- Then the big one. We switch the responsibility of how step entries are found from the old services to the new ones. Creating a journey, warming the cache and the journey map all change at once (I couldn't split this down more without introducing chaos in the test suite)
- The final tidy up is to remove and the old way steps were found, including the fixtures (we are all switched over now)
- The task list at this point doesn't look any different! We've only been changing how we get steps. We end by changing the Journey#show page so that it divides the tasks by section

### Before
Screenshot 2021-02-02 at 18 22 11

### After
Screenshot 2021-02-02 at 18 20 20
Screenshot 2021-02-02 at 17 59 10

## Next steps
[ ]  - remove steps from the develop environment in Contentful
[ ]  - add sections to the master environment in Contentful (add everything to a single section to start?)
[ ] - remove steps from the master environment in Contentful


app/controllers/journeys_controller.rb
```diff
@@ -27,6 +27,9 @@ def show
       steps: [:radio_answer, :short_text_answer, :long_text_answer, :single_date_answer, :checkbox_answers]
     ).find(journey_id)
     @steps = @journey.steps.map { |step| StepPresenter.new(step) }
+    @section_groups = @journey.section_groups.each_with_object({}) { |group, result|
```
tahb, line 30: Not happy with this but short on time. Any ideas welcome!

spec/features/visitors/anyone_can_edit_their_answers_spec.rb
```diff
@@ -1,6 +1,10 @@
 require "rails_helper"
 
 feature "Users can edit their answers" do
+  before do
```
tahb, line 4: Not happy with this but short on time. Context in the commit. Any ideas welcome!

spec/features/visitors/anyone_can_view_a_list_of_tasks_spec.rb
```diff
@@ -2,10 +2,33 @@
 
 feature "Users can view the task list" do
   it "tasks are grouped by their section" do
+    stub_contentful_category(fixture_filename: "multiple-sections.json")
```
jacksonj04, line 5: I've [made a note](https://trello.com/c/iyXMeryJ/366-extract-startjourneyfromcategory-into-reusable-method-in-journeyhelpers) for extracting this stub a category, visit the root, click start boilerplate. Noting here so someone else remembers!

spec/features/visitors/anyone_can_edit_their_answers_spec.rb
jacksonj04: Also can't think of a quick/elegant fix. Suggest making a card to tidy it up later.

app/controllers/journeys_controller.rb
jacksonj04: My only thought would be if we moved initialising the `StepPresenter` into the `Step` itself, make a new `JourneySections` model which was actually stored in the database and related to `Step`, and then we could do a big query with eager loading, and pass the 'iterate over steps in a section' down to the view.

But that then has knock-on implications for the cache warming, and extracting answers for the spec, and checking for completion, plus we're back to needing to store order explicitly.

jacksonj04 (approved):
Shared some thoughts on the gnarly bits but can't offer any blinding insights into better ways. One small error message needs updating to reflect reality.

PR #147, opened 2021-02-03
Update start page content

jacksonj04:
## Changes in this PR

Update the start page content with revised copy.

### Before

dfe-buy-for-your-school-beta-s herokuapp com_

### After

localhost_3000_ (1)


jacksonj04:
I'm in two minds about keeping this as a bunch of strings in I18N vs just moving it into HTML. Thoughts?

tahb:
I'd be happy to keep this as i18n though there isn't a compelling case. Given you've already done the hard work to internationalise and our test suite doesn't have to duplicate all the copy. I have high confidence the page contains what it needs to and it shouldn't change an awful lot.

tahb (approved):
This looks good to me. The only concern I have is how far down the page (below the fold) the start button is, that said I think this is good enough for an early iteration 👍

PR #148, opened 2021-02-04
Improvements to journey map to assist content team

jacksonj04:
A collection of improvements to make the journey map easier to understand, and more useful for the content team.

## Changes in this PR

- The journey map is now clearer about what happens when you click a link
- The journey map now includes links to preview a step
- The journey map now displays markup to include a step's answer in the specification template

### Before

localhost_3000_journey_map_new (1)

### After

localhost_3000_journey_map_new


tahb (approved):
LGTM. 

A tiny suggestion would be to squash those the changelog entries into each commit that introdcued the change so it's self contained.

PR #149, opened 2021-02-04
Users can answer questions which expect a currency 

LornaHarwood:
## Changes in this PR

- Users can answer questions which expect a currency
- Currency is saved as a decimal with precision 11 to ensure that all currency amounts required for this service are valid

## Screenshots of UI changes
Currency step view has a prefix text of a `£` value to indicate that it requires a currency input

Screen Shot 2021-02-08 at 10 33 07

An error is thrown when non numerical values are entered

Screen Shot 2021-02-10 at 10 51 17

Screen Shot 2021-02-10 at 10 48 28

Currency values are saved as decimals in the database

Screen Shot 2021-02-08 at 10 38 16

The specification displays the currency with the full currency formatting to include the `£` and all appropriate commas

Screen Shot 2021-02-08 at 10 48 27

### DIFF
--- a/db/migrate/20210202120716_create_currency_answer.rb
+++ b/db/migrate/20210202120716_create_currency_answer.rb
@@ -1,0 +1,13 @@
+class CreateCurrencyAnswer < ActiveRecord::Migration[6.1]
+  def up
+    create_table :currency_answers, id: :uuid do |t|
+      t.references :step, type: :uuid
+      t.decimal :response, null: false, precision: 11, scale: 2
+      t.timestamps
+    end
+  end
+
+  def down
+    drop_table :currency_answers
+  end
+end
--- a/spec/factories/answer.rb
+++ b/spec/factories/answer.rb
@@ -35,4 +35,10 @@ FactoryBot.define do
 
     response { 150 }
   end
+
+  factory :currency_answer do
+    association :step, factory: :step, contentful_type: "currency", contentful_model: "question"
+
+    response { 1000.01 }
+  end
 end
--- a/spec/features/visitors/anyone_can_complete_a_journey_spec.rb
+++ b/spec/features/visitors/anyone_can_complete_a_journey_spec.rb
@@ -76,6 +76,28 @@ feature "Anyone can start a journey" do
       end
     end
 
+    context "when Contentful entry is of type currency" do
+      scenario "user can answer using a currency number input" do
+        start_journey_from_category_and_go_to_question(category: "currency-question.json")
+
+        fill_in "answer[response]", with: "1,000.01"
+        click_on(I18n.t("generic.button.next"))
+
+        click_first_link_in_task_list
+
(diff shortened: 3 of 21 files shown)


db/migrate/20210202120716_create_currency_answer.rb
```diff
@@ -0,0 +1,13 @@
+class CreateCurrencyAnswer < ActiveRecord::Migration[6.1]
+  def up
+    create_table :currency_answers, id: :uuid do |t|
+      t.references :step, type: :uuid
+      t.decimal :response, null: false, precision: 11, scale: 2
```
erbridge, line 5: Could you add an explanation about why you chose `precision: 11` here to the commit message, maybe?

spec/factories/answer.rb
```diff
@@ -35,4 +35,10 @@
 
     response { 150 }
   end
+
+  factory :currency_answer do
+    association :step, factory: :step, contentful_type: "currency", contentful_model: "question"
+
+    response { Decimal("1000.01") }
```
erbridge, line 42: Why `Decimal` rather than `1000.01` directly?

spec/features/visitors/anyone_can_complete_a_journey_spec.rb
```diff
@@ -76,6 +76,28 @@
...
+        expect(find_field("answer-response-field").value).to eql("1000.01")
+      end
+
+      scenario "throws error when non numerical values are entered" do
+        start_journey_from_category_and_go_to_question(category: "currency-question.json")
+
+        fill_in "answer[response]", with: "£100.00"
+        click_on(I18n.t("generic.button.next"))
+
+        expect(page).to have_content("is not a number")
```
erbridge, line 97: It would be better to show an error explaining for them not to include the £ sign, rather than "not a number". An improvement, not necessarily a blocker.

jacksonj04, line 97: It might be worth, given that people will likely copy and paste into this field, actually accepting values beginning with a "£" and just stripping it out as part of the save method (same as with commas). You can then have a test which makes sure "£100.00" is accepted as intended, plus another test making sure things like "fourteen shillings" is rejected as not a number.

db/migrate/20210202120716_create_currency_answer.rb
erbridge: Thanks.

spec/features/visitors/anyone_can_complete_a_journey_spec.rb
LornaHarwood: I agree - I have added in an error message to reflect this

spec/factories/answer.rb
LornaHarwood: I have amended this to be consistant with the other response inputs, by using 1000.01 directly 👍

jacksonj04 (approved):
💸

PR #155, opened 2021-02-08
Options are no longer capitalised in checkbox questions

jacksonj04:
## Changes in this PR

The desired behaviour is that fields as displayed to the user in checkbox questions are provided exactly as in Contentful. This commit changes the behavior of `checkbox_options` in `StepHelper` to no longer capitalise options.

## Next steps

NB: This does _not_ modify how checkbox options are rendered in specification output, as further work is planned to provide a wider range of options in output (such as an actual array of checked options) which will supersede this.

### DIFF
--- a/app/helpers/step_helper.rb
+++ b/app/helpers/step_helper.rb
@@ -3,7 +3,7 @@
 module StepHelper
   def checkbox_options(array_of_options:)
     array_of_options.map { |option|
-      OpenStruct.new(id: option.downcase, name: option.titleize)
+      OpenStruct.new(id: option.downcase, name: option)
     }
   end
 end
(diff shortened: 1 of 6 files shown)


app/helpers/step_helper.rb
```diff
@@ -3,7 +3,7 @@
 module StepHelper
   def checkbox_options(array_of_options:)
     array_of_options.map { |option|
-      OpenStruct.new(id: option.downcase, name: option.titleize)
+      OpenStruct.new(id: option.downcase, name: option)
```
erbridge, line 6: Should this be `.capitilize` instead?

jacksonj04: I don't think so, since we can't guarantee what the capitalisation of a string should be. I'm erring towards just leaving it as provided by Contentful.

erbridge: 👍

erbridge (approved):

PR #156, opened 2021-02-08
(372) Fix ordering of sections and steps

tahb:
## Changes in this PR

Sections and steps are now ordered correctly, fixing a bug that appeared when a specific combination of named sections were presented incorrectly.

We change the type of data structure we store for the `journey.section_group` and are explicit about the `order` we expect each section and each step (starting from 0).

- minor improvement to the Step factory to make the `contentful_id` dynamic with each new fake step
 
## Screenshots of UI changes

This is what Contentful specifies the order to be:

Screenshot 2021-02-08 at 14 32 19

### Before

This is the order where we can see the bug:

Screenshot_2021-02-08 Catering(1)

### After

And after the bug is fixed:
Screenshot_2021-02-08 Catering

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog 1.0.0].
 - The journey map now displays markup to include a step's answer in the specification template
 - users can answer questions which expect a number
 - checkbox questions no longer alter the capitalisation of options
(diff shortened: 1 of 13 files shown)


jacksonj04 (approved):
Neat solution to the deterministic-but-unexpected database ordering problem!

🚀

PR #162, opened 2021-02-10
(224) Questions can be hidden from the user

tahb:
A first part of [story 224](https://trello.com/c/59KpbPR8). Content users can configure a question to be hidden or shown within the task list and spec. We default to making the question visible as that's what the current behaviour is.

This change includes a narly change to scopes and eager loading to avoid a Bullet warning lock.

The next stage for this work will be to all config to be added to questions that will change the state of this field to make questions visible.

## Screenshots of UI changes

Screenshot 2021-02-10 at 17 07 45


app/models/step.rb
```diff
@@ -9,7 +9,9 @@ class Step < ApplicationRecord
   has_one :checkbox_answers
   has_one :number_answer
 
+  scope :viewable, -> { that_are_questions.that_are_not_hidden }
```
tahb, line 12: Should I remove these, YAGNI?

app/models/step.rb
```diff
@@ -9,7 +9,9 @@ class Step < ApplicationRecord
...
   scope :that_are_questions, -> { where(contentful_model: "question") }
+  scope :that_are_not_hidden, -> { where(hidden: false) }
```
erbridge, line 14: Are we missing a test for this scope?

app/services/create_journey_step.rb
```diff
@@ -99,6 +100,12 @@ def primary_call_to_action_text
     contentful_entry.primary_call_to_action
   end
 
+  def hidden
+    return false unless contentful_entry.respond_to?(:always_show_the_user)
+    return false if contentful_entry.always_show_the_user.nil?
+    !contentful_entry.always_show_the_user
```
erbridge, line 106: This makes me wonder if we should be reversing this property to be "included" or "visible" or something (the positive case).

app/models/step.rb
jacksonj04: I think this can go away, yes. If we do need to apply presentation logic for things that are/aren't questions these should probably happen more explicitly at a higher layer.

app/models/step.rb
tahb: Yes we were missing a test! I have decided to drop the whole commit because these scopes weren't used by the end.

app/services/create_journey_step.rb
tahb: I wondered a few times about naming after the positive case because of the extra conversion hop here. 

In the end I kept coming back to orientating around the negative since it should be by far the exceptional case. Most questions will always be shown/visible/normal so a field like `shown: true` would not be as clear as `hidden: false`, which is hopefully more explicit about the unusual state.

app/models/step.rb
tahb: I dropped it.

app/services/create_journey_step.rb
jacksonj04: I wonder if it's best this being more explicit about what it's doing (although possibly as part of Phase 2 of this work), and making it something like `has_conditionals`?

tahb: Phase 2 (as it has ended up) hasn't changed this code @jacksonj04, I did add an additional field called `additional_step_rule` which contains the rules, which starts to work in tandem with this `hidden` field. This seems like it might capture the "conditionals" part of your comment?

jacksonj04 (approved):

PR #166, opened 2021-02-15
(224) Show and hide additional questions based on previous answers

tahb:
Based on https://github.com/DFE-Digital/buy-for-your-school/pull/162

## Changes in this PR

external-content duckduckgo com

When the user answers a question they may be presented with `n` additional questions.

They can change their mind, and it will hide `n` additional questions.

They can also show or hide any additional question as we will follow any `show_additional_step_rule` for all associated steps incase a new chain breaks out. I think this is most complex part of this pull request and worthy of the most attention. I hope the screenshots and narrative are helpful to demonstrate how this works!

- we add a new filter to the journeys#show action to remove all steps that are in the hidden state, this should prevent them from being shown in the last list or the specification (I think Bullet was useful in the end in pushing us towards this)
- we refactor the code we had for saving the users' answers into a reusable service along with a generic result object. This enables us to present a simple interface back to the controller, keeping it smaller and with better unit test coverage. It also provides a natural place to extend with additional show/hide logic
- we got on a journey with the last 6 commits. Adding and showing _single_ additonal steps and then we do a little refactoring before coming back around and adding and showing _multiple_ additional steps with our good friend recursion (sorry!)

## Screenshots of UI changes

We start with a task list that has 2 hidden questions (locations marked with the red arrows).
Screenshot_2021-02-15 Catering

The first question is answered such that it matches the rule and unlocks the second question (highlighted).
Screenshot_2021-02-15 Catering(1)

The second question is also answered in a way that matches. _It also has an additional rule_ and so the second to last question is now shown. (We complete that too, ending the chain in this example)
Screenshot_2021-02-15 Catering(6)

We go back to the first question and change our mind! The 2 additional questions are once again hidden from view.
Screenshot_2021-02-15 Catering(7)

Finally we realise we were right the first time and switch our first answer back to the matching answer. The previous 2 questions (that have now been answered) appear back in the list as completed steps.
Screenshot_2021-02-15 Catering(8)


app/services/save_answer.rb
```diff
@@ -0,0 +1,25 @@
+class SaveAnswer
+  attr_accessor :answer, :step
+  def initialize(answer:)
+    self.answer = answer
+    self.step = answer.step
+  end
+
+  def call(answer_params: {}, checkbox_params: {})
```
tahb, line 8: It would be nice not to have to 2 parameters for parameters. I didn't spend much time figuring out the existing issue with the controller because I had enough on so I just focused on honouring the same interface.

jacksonj04: Yeah, this feels like we might need another layer (especially given that this will now need a third one to deal with dates being an exceptional case as well), but for now it'll hold up.

app/services/toggle_additional_steps.rb
```diff
@@ -36,9 +50,22 @@ def check_to_show_additional_step!
...
 
-  def check_to_hide_additional_step!
+  def check_to_hide_additional_steps!
     return if matching_answer?
 
-    additional_step.update(hidden: true)
+    recursively_toggle_additional_steps!(next_step_id: additional_step_id, hidden_state: true)
+  end
+
+  def recursively_toggle_additional_steps!(next_step_id:, hidden_state:)
```
jacksonj04, line 59: I heard you liked recursion in your recursion...

jacksonj04 (approved):

PR #167, opened 2021-02-15
(343) Fix date validation

tahb:
## Changes in this PR

Incorrect values for dates will return an error to the user, rather than causing [a crash](https://rollbar.com/dxw/dfe-buy-for-your-school/items/72/?item_page=0&item_count=100&#traceback). 

When the user adds `0, 0, 2` into the 3 fields Ruby would fail to parse this as a Date when assigning to the answer object. This happens before traditional model validation can take place, leaving us a bit stuck.

Using the [MOJ Form builder we can do this by adding a widget](https://github.com/ministryofjustice/gov_uk_date_fields#3-tell-your-models-attributes-are-gov_uk_dates) called `acts_as_gov_uk_date`. This does not appear to be present in the DfE Form builder which we are using, and that is very much part of our service (not something easily swapped out to).

Given that we have spent many days on hunting for the conventional approach to solving this problem, this solution takes a "functional over elegant" approach. The approach for this work is [based on the one the RODA team settled on](https://github.com/UKGovernmentBEIS/beis-report-official-development-assistance/blob/f94d7fdb18dcf8df15f8d3a36de7e9a67e2e822a/app/controllers/staff/activity_forms_controller.rb#L81) (as they use the same DfE gem) and should be equivalant.

The error copy says "real" which I confess is not the best use of language to convey an abstract "your date doesn't look quite right but we can't tell you exactly why". Any suggestions are of course welcome!

### Before
Screenshot 2021-02-15 at 15 34 45

### After
Screenshot 2021-02-15 at 15 27 48

### DIFF
--- a/app/controllers/answers_controller.rb
+++ b/app/controllers/answers_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class AnswersController < ApplicationController
+  include DateHelper
+
   def create
     @journey = Journey.find(journey_id)
     @step = Step.find(step_id)
@@ -9,7 +11,7 @@ class AnswersController < ApplicationController
     @answer.step = @step
 
     result = SaveAnswer.new(answer: @answer)
-      .call(checkbox_params: checkbox_params, answer_params: answer_params)
+      .call(checkbox_params: checkbox_params, answer_params: answer_params, date_params: date_params)
     @answer = result.object
 
     if result.success?
@@ -24,7 +26,7 @@ class AnswersController < ApplicationController
     @step = Step.find(step_id)
 
     result = SaveAnswer.new(answer: @step.answer)
-      .call(checkbox_params: checkbox_params, answer_params: answer_params)
+      .call(checkbox_params: checkbox_params, answer_params: answer_params, date_params: date_params)
     @answer = result.object
 
     if result.success?
@@ -50,5 +52,11 @@ class AnswersController < ApplicationController
 
   def checkbox_params
     params.require(:answer).permit(response: [])
+  end
+
+  def date_params
+    answer = params.require(:answer).permit(:response)
+    date_hash = {day: answer["response(3i)"], month: answer["response(2i)"], year: answer["response(1i)"]}
+    {response: format_date(date_hash)}
   end
 end
--- a/spec/models/single_date_answer_spec.rb
+++ b/spec/models/single_date_answer_spec.rb
@@ -11,6 +11,11 @@ RSpec.describe SingleDateAnswer, type: :model do
   end
 
   describe "validations" do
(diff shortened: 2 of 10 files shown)


app/controllers/answers_controller.rb
```diff
@@ -11,6 +13,10 @@ def create
     case @step.contentful_type
     when "checkboxes"
       @answer.assign_attributes(checkbox_params)
+    when "single_date"
```
tahb, line 16: Conflicts incoming https://github.com/DFE-Digital/buy-for-your-school/pull/166/files#diff-9a47a625588088d85a4a0251c15bc1d65bc9123e203ac9ce55759ac74db1a50dR11

app/controllers/answers_controller.rb
```diff
@@ -31,6 +37,10 @@ def update
...
+      answer = params.require(:answer).permit(:response)
+      date = {day: answer["response(3i)"], month: answer["response(2i)"], year: answer["response(1i)"]}
```
erbridge, line 42: This looks very similar to your `date` method. Is it actually different?

app/controllers/answers_controller.rb
```diff
@@ -60,4 +70,9 @@ def answer_params
   def checkbox_params
     params.require(:answer).permit(response: [])
   end
+
+  def date
```
erbridge, line 74: Could we make it clearer from the name that this comes from `params`, maybe?

spec/models/single_date_answer_spec.rb
```diff
@@ -11,6 +11,11 @@
   end
 
   describe "validations" do
-    it { is_expected.to validate_presence_of(:response) }
+    it "validates missing presence" do
```
erbridge, line 14:
```suggestion
    it "validates missing response" do
```

app/controllers/answers_controller.rb
```diff
@@ -60,4 +70,9 @@ def answer_params
...
+    response = params.require(:answer).permit(:response)
+    {day: response["response(3i)"], month: response["response(2i)"], year: response["response(1i)"]}
```
erbridge, line 76:
```suggestion
    answer = params.require(:answer).permit(:response)
    {day: answer["response(3i)"], month: answer["response(2i)"], year: answer["response(1i)"]}
```

app/controllers/answers_controller.rb
tahb: I left the method in by mistake 🙃

tahb: (I think I'll keep it on reflection)

jacksonj04 (approved):
As mentioned in the commit messages this _feels wrong_, but solves the problem, so 🚀

PR #169, opened 2021-02-16
Only reshow hidden questions that qualify

tahb:
* Before this change the user could show a question and it would go and show all dependent questions regardless of them having answers, or indeed answers that were required.
* This change turns the `matching_answer?` method into a functional method so it can be reused within the recursion as well as outside it
* We continue to hide all dependent child questions regardless of their answer

### DIFF
--- a/app/services/toggle_additional_steps.rb
+++ b/app/services/toggle_additional_steps.rb
@@ -34,37 +34,51 @@ class ToggleAdditionalSteps
     journey_steps.find_by(contentful_id: additional_step_id)
   end
 
-  def matching_answer?
-    expected_answer = additional_step_rule["required_answer"].downcase
-    case answer.response.class.name
+  def matching_answer?(a:, b:)
+    expected_answer = a.downcase
+    case b.class.name
     when "Array"
-      answer.response.map(&:downcase).include?(expected_answer)
+      b.map(&:downcase).include?(expected_answer)
     else
-      expected_answer == answer.response.downcase
+      expected_answer == b.downcase
     end
   end
 
   def check_to_show_additional_steps!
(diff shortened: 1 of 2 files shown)


app/services/toggle_additional_steps.rb
```diff
@@ -34,37 +34,51 @@ def additional_step
...
   def check_to_show_additional_steps!
-    return unless matching_answer?
-
-    recursively_toggle_additional_steps!(next_step_id: additional_step_id, hidden_state: false)
+    recursively_show_additional_steps!(current_step: step, next_step_id: additional_step_id)
   end
 
   def check_to_hide_additional_steps!
-    return if matching_answer?
+    return if matching_answer?(a: step.additional_step_rule["required_answer"], b: step.answer.response)
```
tahb, line 52: If it's not a match, we continue to hide every question down the chain regardless of their answers.

app/services/toggle_additional_steps.rb
```diff
@@ -34,37 +34,51 @@ def additional_step
...
+      )
+    end
   end
 
-  def recursively_toggle_additional_steps!(next_step_id:, hidden_state:)
+  def recursively_show_additional_steps!(current_step:, next_step_id:)
+    return unless current_step.answer && current_step.additional_step_rule
+
     if next_step_id
+      return unless matching_answer?(a: current_step.additional_step_rule["required_answer"], b: current_step.answer.response)
```
tahb, line 74: This is the magic line that fixes the problem. When showing questions again we always evaluate the answer exists and that it matches before showing it.

jacksonj04 (approved):

PR #170, opened 2021-02-16
(311) Parse markdown in help text

jacksonj04:
## Changes in this PR

- markdown in help text fields is now parsed into HTML

### Before

localhost_3000_journeys_a6ba1ea9-e1cd-4d50-bba5-28b4e45f2803_steps_1b4cf3c8-35a8-4af1-9ab1-26f2e2dc1531 (1)

### After

localhost_3000_journeys_a6ba1ea9-e1cd-4d50-bba5-28b4e45f2803_steps_1b4cf3c8-35a8-4af1-9ab1-26f2e2dc1531


Gemfile
```diff
@@ -66,3 +66,5 @@ group :test do
   gem "database_cleaner"
   gem "launchy"
 end
+
+gem "redcarpet", "~> 3.5"
```
tahb, line 70: This looks like it ought to belong earlier up the file with the other gems that don't sit within a group.

app/models/step.rb
```diff
@@ -31,4 +31,8 @@ def primary_call_to_action_text
   def options_list
     options.map { |hash| hash["value"] }
   end
+
+  def help_text_html
```
tahb, line 35: Did you wonder if this should or could live within the `StepPresenter`?

tahb (approved):
Looks good! Couple of suggestions and I think we need an entry in our changelog.

PR #171, opened 2021-02-16
Fix redis connection failures by depending only on REDIS_URL

tahb:
## Changes in this PR

The Heroku Redis addon controls the REDIS_URL environment variable. Sometimes this value changes[1] and the host and/or port changes leading to connection failures after a random interval, and without warning.

This commit changes our approach from making the database number configurable through ENV (using `REDIS_CACHE_URL` and `REDIS_SIDEKIQ_URL`) to being hardcoded into each service that uses Redis: Sidekiq and Redis for caching.

[1] https://data.heroku.com/datastores/49f12bfc-2341-465f-ae43-70f7c4edabcd#settings

> Please note that these credentials are not permanent.
Heroku rotates credentials periodically and updates applications where this datastore is attached.

Screenshot 2021-02-16 at 17 10 23

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,8 +10,7 @@ ROLLBAR_ACCESS_TOKEN=ROLLBAR_ACCESS_TOKEN
 ROLLBAR_ENV=development
 
 DATABASE_URL=postgres://postgres@localhost:5432/buy-for-your-school-development
-REDIS_CACHE_URL=redis://localhost:6379/1
-REDIS_SIDEKIQ_URL=redis://localhost:6379/2
+REDIS_URL=redis://localhost:6379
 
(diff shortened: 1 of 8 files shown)


jacksonj04 (approved):
My only thought on this would be if it's worth making the db number _also_ defined in an env variable, eg `REDIS_PRIMARY_DB_NUMBER` and `REDIS_SIDEKIQ_DB_NUMBER`, so it's just a bit more explicit.

But that is a very low concern.

tahb:
I hadn't thought of making ENV vars only for the db number. Given the time pinch I think it'll be better to merge them with them hardcoded and move on swiftly from this fix.

PR #174, opened 2021-02-17
(342) Users can answer using the extended checkbox pattern

tahb:
## Changes in this PR

Users can now be asked to provide further information for the choice they've made, if Contentful has specified that behaviour for the given answer

There should be no surprises here since we are following the precedent set by the extended-radio pattern. Fortunately the [DfE Form builder gem supports this "Other" mechanism](https://govuk-form-builder.netlify.app/form-elements/checkboxes/#generating-checkboxes-with-a-conditionally-revealed-text-field) in the same way. 

### Before

Screenshot 2021-02-17 at 15 07 35

### After

Screenshot 2021-02-17 at 15 07 40

With multiple options we can save multiple bits of further info:
Screenshot 2021-02-18 at 17 40 14

### DIFF
--- a/app/helpers/step_helper.rb
+++ b/app/helpers/step_helper.rb
@@ -6,4 +6,19 @@ module StepHelper
       OpenStruct.new(id: option.downcase, name: option)
     }
   end
+
+  def monkey_patch_form_object_with_further_information_field(
+    form_object:,
+    associated_choice:
+  )
+    existing_further_information = form_object&.further_information&.fetch(
+      "#{associated_choice}_further_information", nil
+    )
+
+    form_object.define_singleton_method("#{associated_choice}_further_information") do
+      existing_further_information
+    end
+
+    form_object
+  end
 end
--- a/spec/features/visitors/anyone_can_complete_a_journey_spec.rb
+++ b/spec/features/visitors/anyone_can_complete_a_journey_spec.rb
@@ -165,15 +165,22 @@ feature "Anyone can start a journey" do
           start_journey_from_category_and_go_to_question(category: "extended-checkboxes-question.json")
 
           check("Yes")
-          fill_in "answer[further_information]", with: "The school needs the kitchen cleaned once a day"
+          fill_in "answer[yes_further_information]", with: "The first piece of further information"
+
+          check("No")
+          fill_in "answer[no_further_information]", with: "A second piece of further information"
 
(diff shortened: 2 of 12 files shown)


spec/features/visitors/anyone_can_complete_a_journey_spec.rb
```diff
@@ -159,6 +159,23 @@
...
+
+          check("Yes")
+          fill_in "answer[further_information]", with: "The school needs the kitchen cleaned once a day"
+
+          click_on(I18n.t("generic.button.next"))
+
+          click_first_link_in_task_list
+
+          expect(page).to have_checked_field("Yes")
+          expect(find_field("answer-further-information-field").value)
```
jacksonj04, line 175: I think it's worth having a test here to make sure that if we have _multiple_ extended checkbox fields then they all behave according to plan, ie each one can store its response differently.

jacksonj04 (changes requested):
A question to resolve around how this behaves with multiple extended options in a single checkbox set

spec/features/visitors/anyone_can_complete_a_journey_spec.rb
tahb: Oh this is a major blunder! We can't store more than 1 piece of `further_information` in the same column, whilst it's a string at least. This is going to require more work.

tahb:
@jacksonj04 Ready for another look. Certainly not proud of this one but I can't see another way to do this without Javascript trickery.

app/helpers/step_helper.rb
```diff
@@ -6,4 +6,19 @@ def checkbox_options(array_of_options:)
       OpenStruct.new(id: option.downcase, name: option)
     }
   end
+
+  def monkey_patch_form_object_with_further_information_field(
```
jacksonj04, line 10: This is possibly one of my favourite method names to have ever reviewed, and is also entirely descriptive.

🔧 🐵

jacksonj04 (approved):
Approve for _answering_, but we don't currently expose this in the answers variables. It might be worth pushing this into [this card](https://trello.com/c/Km749nb6/373-for-extended-radio-and-checkbox-questions-distinguish-between-the-provided-answer-which-was-chosen-and-the-extended-answer-which), which handles dealing with exposing answers in arrays and things.

PR #176, opened 2021-02-22
(409) Multiple optional questions can be shown and hidden

tahb:
## Changes in this PR

Before this change we allowed the content user to "bolt on" **one** additional question if the answer matched an expected pattern.

This change allows them to show as many additional questions as they like. Though the current need is for 2 additonal questions. We do this by accepting an array of identifiers rather than a single string.

The amount of calculation required to determine and follow the question branches through to their resolution is not ideal and we know this. A better approach would be to keep track of the state of the tree of questions and make the required database updates with fewer database queries. This would require more time than we have in the last sprint so we are opting for something that works and accepting the technical debt will need to be repaid.

## Screenshots of UI changes

For this example we have a tree structure as follows. We have one question that is "Always shown" and 3 hidden questions with another subtree of the first and third hidden question being connected:

Screenshot 2021-02-22 at 11 34 19

We start with a single question. We only need to pay attention to the first section in this example:
Screenshot_2021-02-22 Catering

We answer it with a non-matching answer (so no new questions are shown)
Screenshot_2021-02-22 Catering(1)

We change our answer to the matching answer (and now expect 2 new questions to be shown)
Screenshot_2021-02-22 Catering(2)

We change our original answer and check both new questions are hidden again:
Screenshot_2021-02-22 Catering(3)

We check the subtree by answering the first hidden question with a matching answer and expect the third question to be revealed:
Screenshot_2021-02-22 Catering(4)

Finally we change our original answer again and check that all 3 connected questions are hidden:
Screenshot_2021-02-22 Catering(5)

## Next steps

Update Contentful entries that have optional logic already and change their value into an array of one item, rather than a string.

### DIFF
--- a/app/services/toggle_additional_steps.rb
+++ b/app/services/toggle_additional_steps.rb
@@ -26,12 +26,12 @@ class ToggleAdditionalSteps
     step.additional_step_rule
   end
 
-  def additional_step_id
+  def additional_step_ids
     additional_step_rule["question_identifier"]
   end
 
   def additional_step
-    journey_steps.find_by(contentful_id: additional_step_id)
+    journey_steps.find_by(contentful_id: additional_step_ids)
   end
 
   def matching_answer?(a:, b:)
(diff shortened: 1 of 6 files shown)


app/services/toggle_additional_steps.rb
```diff
@@ -45,41 +45,43 @@ def matching_answer?(a:, b:)
...
-    recursively_hide_additional_steps!(current_step: step, next_step_id: additional_step_id)
+    recursively_hide_additional_steps!(next_step_ids: additional_step_ids)
   end
 
-  def recursively_hide_additional_steps!(current_step:, next_step_id:)
-    if next_step_id
-
-      next_step = journey_steps.find_by(contentful_id: next_step_id)
-      next_step.update(hidden: true)
+  def recursively_hide_additional_steps!(next_step_ids:)
```
tahb, line 57: `current_step` wasn't used as a parameter so it was removed.

jacksonj04 (approved):
LGTM

PR #180, opened 2021-02-23
Fix cases where empty help text would cause errors

jacksonj04:
Since we weren't correctly guarding against empty `help_text` for some question types, when the `help_text` was empty we would try call methods on it that did not exist.

To make this easier to understand and cleaner for people in future, this also spends some time making sure the individual step views make use of `StepPresenter` instead of the `Step` object directly and moves the `help_text_html` into this presentation layer.

### DIFF
--- a/app/views/steps/_new_form_wrapper.html.erb
+++ b/app/views/steps/_new_form_wrapper.html.erb
@@ -1,5 +1,5 @@
-<%= content_for :title, @step.title %>
+<%= content_for :title, @step_presenter.title %>
 <%= form_for @answer, as: :answer, url: journey_step_answers_path(@journey, @step), method: "post" do |f| %>
   <%= yield f %>
-  <%= f.submit @step.primary_call_to_action_text, class: "govuk-button" %>
+  <%= f.submit @step_presenter.primary_call_to_action_text, class: "govuk-button" %>
 <% end %>
--- a/app/views/steps/currency.html.erb
+++ b/app/views/steps/currency.html.erb
@@ -1,7 +1,7 @@
 <%= render layout: layout do |f| %>
   <%= f.govuk_text_field :response,
-    label: { text: @step.title, size: "l" },
+    label: { text: @step_presenter.title, size: "l" },
     prefix_text: "£",
-    hint: -> { @step.help_text_html }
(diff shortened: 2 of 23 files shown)


app/views/steps/currency.html.erb
```diff
@@ -2,6 +2,6 @@
   <%= f.govuk_text_field :response,
     label: { text: @step.title, size: "l" },
     prefix_text: "£",
-    hint: -> { @step.help_text_html.html_safe }
+    hint: -> { @step.help_text_html.html_safe if @step.help_text.present? }
```
tahb, line 5: I would be thinking about using the `StepPresenter` for this to make the change in one place, remove the guard logic from the views and remove the need to remember to add it to future views:

```
Views
     hint: -> { @step.help_text_html } 
```

```
class StepPresenter
  def help_text_html
    return unless help_text.present?

    Redcarpet::Markdown.new(Redcarpet::Render::HTML)
      .render(step.help_text_html)
      .html_safe
  end
```
  
Usually I'd test this myself before throwing code around but I can't install redcarpet locally at the moment..

app/views/steps/radios.html.erb
```diff
@@ -3,7 +3,7 @@
   <%= f.govuk_radio_buttons_fieldset(:response, legend: { size: "l", text: @step.title }) do %>
     <% if @step.help_text.present? %>
```
tahb, line 4: Ah I guess we need this guard still because of the `div`, shame!

app/views/steps/_new_form_wrapper.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= content_for :title, @step.title %>
```
tahb, line 1: If we also swapped `@step` to `@step_presenter` here too, I wonder if `@step` could become local variable in the controller.

tahb (approved):

app/views/steps/_new_form_wrapper.html.erb
jacksonj04: Only if we also push in the path in place of `journey_step_answer_path`, but this is a `@step` I missed out in swapping to use the presenter!

PR #183, opened 2021-02-23
Checkbox info fields no longer try and create their own label

tahb:
## Changes in this PR

..When one is not provided

We instead don't show a label at all.
### Before

Screenshot 2021-02-23 at 15 11 34

### After

Screenshot 2021-02-23 at 15 11 48


jacksonj04 (approved):
Tentative approve - I'm not sure how well this will fare for accessibility in having an unlabelled text box kicking around, but I guess we can fix should it actually flag as an issue.

tahb:
Good point Nick. I've gone back and made sure the label is still there but hidden using the gem's options https://www.rubydoc.info/gems/govuk_design_system_formbuilder/GOVUKDesignSystemFormBuilder/Builder#govuk_text_field-instance_method

Screenshot 2021-02-23 at 16 43 25

tahb:
Another edit. If label text isn't provided, format the hidden value with a default "Optional further information" to be explicit (and potentially repetitive) for screen readers.

Screenshot 2021-02-23 at 17 02 26

PR #185, opened 2021-02-23
Create an extended further information text box option

LornaHarwood:
## Changes in this PR

Radio answers and checkbox answers now have the option to either display a `single` or `long` text box for their further information text box

## Screenshots of UI changes

This is the `single` option:

### Radio answer
Screen Shot 2021-02-23 at 11 40 25

### Checkbox answer
Screen Shot 2021-02-24 at 15 01 14

This is the `long` option:

### Radio answer
Screen Shot 2021-02-23 at 11 30 58

### Checkbox answer
Screen Shot 2021-02-24 at 15 49 40

### DIFF
--- a/app/views/steps/radios.html.erb
+++ b/app/views/steps/radios.html.erb
@@ -8,13 +8,13 @@
     <% end %>
 
     <% @step.options.each do |option| %>
-      <% if option["display_further_information"] == true %>
         <%= f.govuk_radio_button :response, option["value"].downcase, label: { text: option["value"] }, hint: { text: option["help_text"] } do %>
-          <%= f.govuk_text_area :further_information, rows: 1, label: { text: option["further_information_help_text"] } %>
+          <% if option["display_further_information"] == "single" || option["display_further_information"] == true %>
+            <%= f.govuk_text_field :further_information, label: { text: option["further_information_help_text"] } %>
+          <% elsif option["display_further_information"] == "long" %>
+            <%= f.govuk_text_area :further_information, rows: 6, label: { text: option["further_information_help_text"] } %>
+          <% end %>
         <% end %>
-      <% else %>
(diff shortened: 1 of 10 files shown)


app/views/steps/radios.html.erb
```diff
@@ -8,10 +8,14 @@
...
 
     <% @step.options.each do |option| %>
-      <% if option["display_further_information"] == true %>
+      <% if option["display_further_information"] == "single" || option["display_further_information"] == true %>
         <%= f.govuk_radio_button :response, option["value"].downcase, label: { text: option["value"] }, hint: { text: option["help_text"] } do %>
           <%= f.govuk_text_area :further_information, rows: 1, label: { text: option["further_information_help_text"] } %>
         <% end %>
+      <% elsif option["display_further_information"] == "long" %>
+        <%= f.govuk_radio_button :response, option["value"].downcase, label: { text: option["value"] }, hint: { text: option["help_text"] } do %>
+          <%= f.govuk_text_area :further_information, rows: 6, label: { text: option["further_information_help_text"] } %>
```
tahb, line 17: Is the only difference in what this conditional can return the rows count? If so, we could refactor this further with something like:

```
<%= f.govuk_text_area :further_information, rows: row_count(option: option), label: { text: option["further_information_help_text"] } %>
```

```
class StepHelper
  def row_count
    return 6 if option["display_further_information"] == "long"
    1
  end
end
```

Or perhaps something more expressive

```
def row_count
  case option["display_further_information"]
  when "single" then 1
  when "long" then 6
  else 
    1
  end
end
```

tahb: After further thought! Should we actually use the `govuk_text_field` rather than `govuk_text_area`, this feels more correct? In which case my previous idea for refactoring isn't applicable but some DRYing up could still be done I think.

jacksonj04 (approved):
LGTM

PR #186, opened 2021-02-24
Brakeman runs on CI again

tahb:
## Changes in this PR

- add this as a testing steip to script/test in keeping with the 'One script to rule them all' pattern
- disregard the output so brakeman doesn't open an interactive prompt, we could add the report to a file but the container is ephemeral

We have an existing security warning to address as part of this work now as we had a period of time where we weren't running these checks.

CI will now explode (with a level 1 error and a fake level 3) Screenshot 2021-02-24 at 11 44 23

### DIFF
--- a/app/controllers/journeys_controller.rb
+++ b/app/controllers/journeys_controller.rb
@@ -45,7 +45,7 @@ class JourneysController < ApplicationController
     respond_to do |format|
       format.html
       format.docx do
-        render docx: "specification.docx", content: @specification_html, layout: "specficiation"
+        render docx: "specification.docx", content: @specification_html
       end
     end
   end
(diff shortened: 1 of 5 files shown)


app/controllers/journeys_controller.rb
```diff
@@ -40,12 +40,12 @@ def show
...
 
     @answers = GetAnswersForSteps.new(visible_steps: @visible_steps).call
-    specification_html = @specification_template.render(@answers)
+    @specification_html = @specification_template.render(@answers)
 
     respond_to do |format|
       format.html
       format.docx do
-        render docx: "specification.docx", content: specification_html, layout: "specficiation"
+        render docx: "specification.docx", content: @specification_html, layout: "specficiation"
```
jacksonj04, line 48: One tiny `specficiation` typo whilst we're in this line anyway.

jacksonj04 (approved):
One teeny tyop, otherwise yay for analysis!

app/controllers/journeys_controller.rb
tahb: I went to fix the typo and found this `layout` option to be redundant so it, and it's typo have gone in the bin!

PR #187, opened 2021-02-24
Optimise expensive database queries

tahb:
## Changes in this PR

Refactor our implementation of `step.answer`.

Instead of trying each association in order we are selective about the query we expect to make. This changes the outcome of method, reducing the maximum of 7 database queries (if the step is currency) to always being at most 1 query. To do this we interogate the `contentful_type` field.

## Screenshots of UI changes

For example, this is the database output when asking a checkbox step for it's answer (and how I spotted this issue). You can see we make 4 redundant SELECT queries before checking for a checkbox answer: 

### Before

Screenshot 2021-02-24 at 11 38 42

### After

We now only make one query we always expect to retrieve something or raise a RecordNotFound:

Screenshot 2021-02-24 at 12 17 42

### DIFF
--- a/app/models/step.rb
+++ b/app/models/step.rb
@@ -14,13 +14,15 @@ class Step < ApplicationRecord
 
   def answer
     @answer ||=
-      radio_answer ||
-      short_text_answer ||
-      long_text_answer ||
-      single_date_answer ||
-      checkbox_answers ||
-      number_answer ||
-      currency_answer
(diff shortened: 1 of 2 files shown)


jacksonj04 (approved):
This is almost _but not quite_ at the point where doing this kind of thing iteratively would make more sense.

Almost.

Looks good, yay for reducing DB load!

PR #188, opened 2021-02-24
Move specification rendering into a service object

jacksonj04:
Previously, all rendering work for the specification was done in the journey controller. This moves the job of combining the template and the answers into a rendered output into a new service object, which gives us a clearer separation of which bit of code is responsible for the rendering.

This was originally planned to be a first step in adding an 'incomplete spec' warning message to the .docx output, but that work has had its priorities changed and there's no point this (entirely functional) refactoring languishing around unmerged.

### DIFF
--- a/app/services/specification_renderer.rb
+++ b/app/services/specification_renderer.rb
@@ -1,0 +1,12 @@
+class SpecificationRenderer
+  def initialize(template:, answers:)
+    @template = Liquid::Template.parse(
+      template, error_mode: :strict
+    )
+    @answers = answers
+  end
+
+  def to_html
+    @template.render(@answers).html_safe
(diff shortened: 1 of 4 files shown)


app/services/specification_renderer.rb
```diff
@@ -0,0 +1,13 @@
+class SpecificationRenderer
```
tahb, line 1: I would think about adding a new spec for this class and repeat the part of the feature spec that is currently covering it, for completeness.

I see some value in more coverage that would help catch regressions if that existing feature spec coverage changes and in reducing the friction (and encouraging more unit tests) when extending this class further.

tahb (approved):

PR #189, opened 2021-02-24
Checkbox answers are now human readable in a spec

tahb:
## Changes in this PR

Our test didn't have an accurate case for how we are storing answers. We store a machine readable `morning_break` in  the `response` field (so we can reliably set field labels etc). Our test set the database value as "Morning break" which is close but let through this bug where all values are formatted in the machine readable format, rather than a human readable one.

We now strip the underscores from the value if it as multiple words during presentation.

### Before

Screenshot 2021-02-24 at 16 55 30

### After
Screenshot 2021-02-24 at 17 10 19


jacksonj04 (approved):
Good catch

PR #192, opened 2021-02-25
(373) Extended options for Radio and Checkbox answers are now available for use in the specification

tahb:
Based atop https://github.com/DFE-Digital/buy-for-your-school/pull/189

## Changes in this PR

The liquid template has a new variable available to it for every Radio and Checkbox answer, called `extended_answer_x`. This is _aswell as_ the previous `answer_x`.

```
# Radio
answer_x => "Yes"
extended_answer_x => [
  { 
    "response" => "Yes",
    "further_information" => "Some further information to tell you about" 
  }
]

# Checkboxes
answer_x => "Yes, no"
extended_answer_x => [
  { 
    "response" => "Yes",
    "further_information" => "Some further information to tell you about" 
  }
  { 
    "response" => "No",
    "further_information" => "" 
  }
]

```

The hope is that by making both available the content team will have the flexiblity to list the checklist answers only OR start to mixin the further information fields too, where relevant.

There is a small piece of breaking behaviour. Before `answer_x` for Radios would render an interpolation eg. "Yes - Further info". This change means that now only "Yes" will be rendered and the content team will need to reevaluate the template and add in this additional logic to access the further information.

This is an example of the interface to Liquid we can now support (which was used to generate the screenshot):

```
<!-- Get the response only as we could before (which will be a comma separated list for checkboxes) -->
{{ answer_7p7URWwZbkjqqzo3E0k1v4 }}

<!-- To get further_information we make a single new variable with a prefix -->
<ul>
  {% for extended_answer in extended_answer_7p7URWwZbkjqqzo3E0k1v4 %}
    <li>{{ extended_answer["response"] }}</li>
    {% if extended_answer["further_information"] != "" %}
      <li>{{ extended_answer["further_information"] }}</li>
    {% endif %}
  {% endfor %}
</ul>
</article>
```

We use new helper methods to centralise our logic for reading and writing the further_information keys to and from human readable.

## Screenshots of UI changes

Screenshot 2021-02-25 at 12 39 28

### DIFF
--- a/app/services/get_answers_for_steps.rb
+++ b/app/services/get_answers_for_steps.rb
@@ -22,12 +22,21 @@ class GetAnswersForSteps
       end
 
       if answer.respond_to?(:further_information)
-        hash["extended_answer_#{step.contentful_id}"] = [
-          {
-            "response" => answer.response,
-            "further_information" => answer.further_information
-          }
-        ]
+        hash["extended_answer_#{step.contentful_id}"] = if answer.further_information.is_a?(Hash)
+          answer.further_information.each_pair.each_with_object([]) do |(key, value), array|
+            array << {
+              "response" => key.gsub("_further_information", "").capitalize,
+              "further_information" => value
+            }
+          end
+        else
+          [
+            {
+              "response" => answer.response,
+              "further_information" => answer.further_information
+            }
+          ]
+        end
       end
 
       hash["answer_#{step.contentful_id}"] = answer.response.to_s
--- a/app/services/get_answers_for_steps.rb
+++ b/app/services/get_answers_for_steps.rb
@@ -21,6 +21,15 @@ class GetAnswersForSteps
       else raise UnexpectedAnswer.new("Trying to present an unknown type of answer: #{step.answer.class.name}")
       end
 
+      if answer.respond_to?(:further_information)
+        hash["extended_answer_#{step.contentful_id}"] = [
+          {
+            "response" => answer.response,
+            "further_information" => answer.further_information
+          }
+        ]
+      end
+
       hash["answer_#{step.contentful_id}"] = answer.response.to_s
     }
   end
--- a/app/services/get_answers_for_steps.rb
+++ b/app/services/get_answers_for_steps.rb
@@ -1,6 +1,8 @@
 class GetAnswersForSteps
   class UnexpectedAnswer < StandardError; end
 
+  include AnswerHelper
+
   attr_accessor :visible_steps
   def initialize(visible_steps:)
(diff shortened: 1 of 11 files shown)


app/services/get_answers_for_steps.rb
```diff
@@ -22,12 +22,21 @@ def call
       end
 
       if answer.respond_to?(:further_information)
-        hash["extended_answer_#{step.contentful_id}"] = [
-          {
-            "response" => answer.response,
-            "further_information" => answer.further_information
-          }
-        ]
+        hash["extended_answer_#{step.contentful_id}"] = if answer.further_information.is_a?(Hash)
```
tahb, line 27: I had a note to refactor the responsibility of this new logic. Apart from the vertical height and that this method is doing a fair few things now, the fact that it only works for the answers that respond to `further_information` is a smell to me. I'd like to know if you can see any way to improve this as I've been a bit blinded by the detail for the moment!

jacksonj04:
I think we can simplify the radio output, since they only have one possible condition:

``` ruby
{
  answer_x => "Yes",
  extended_answer_x => "Some further information to tell you about"
}
```

But on the flip side I'm wondering if for checkboxes we actually need to be even more complex to allow for cleaner spec templates:

``` ruby
{
  answer_x => ["Yes", "No"],
  extended_answer_x => [
    yes => {
      response => "Yes",
      extended_answer => "Some further information to tell you about"
    }
  ]
}
```

So that we can do `{% if answer_x contains "yes" %}`, but also `{% if answer_x.yes %}`, as well as `{% for answer in extended_answer_x %}`.

tahb:
Thanks for the feedback! 

My idea was to make the same pattern for storing and accessing the variables work for both patterns but I take your point that for radios it is only ever going to be 1!

What's your position on what we should do now as opposed to later?

jacksonj04:
I think we could make this two PRs - one has the change to stop combining radio answers, and to add a new `extended` with the longer response, so pretty much the first two commits with that tweak so you just get a string back instead of an object for `extended`.

The second we can then think about checkboxes more clearly, although I'm erring towards shipping as is at this point and leaving the option of 'tidying up' for later.

tahb:
I can see that being able to express `{% if extended_answer_x.yes %}` could indeed be valuable to the content team!

Given that we are collecting all the bits of information we can migrate the data structures later if there's a need. I'd be happy to merge and come back around on it quickly if needs be.

jacksonj04 (approved):
✅  for what's here right now - happy for it to go and then we look at improvements later as/when they become necessary

PR #193, opened 2021-02-25
(354) Radio questions can be configured to show an OR separator

tahb:
## Changes in this PR

This change allows Radio questions within Contentful to specify a new attribute within it's existing `option` field called `separated_by_or`.

If omitted or set to false, no separator will be shown.

```
[
    {
        "value": "Yes"
    },
    {
        "value": "No",
        "separated_by_or": true
    }
]
```

### Before

Screenshot 2021-02-25 at 16 42 28

### After

Screenshot 2021-02-25 at 16 42 25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ The format is based on [Keep a Changelog 1.0.0].
 - users can be presenting with forking/branching question chains based on any one answer
 - create an extended further information text box option for radio answers
 - create extended further information text box option for checkbox answers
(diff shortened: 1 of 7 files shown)


tahb:
FYI: When reviewing this you may notice the same bug I did. When you click between radio options that don't have `further_information` configured the page will jump up and down as the gem inserts and toggles an unused container element. I'll fix this seperately now.

LornaHarwood (approved):
LGTM 👍

PR #194, opened 2021-02-25
Fix page jump for radio questions without further info

tahb:
## Changes in this PR

When the radio question does NOT have a further_information of any sorts then when the user selects an option, the radio elements will jump slightly on the page.

This is caused by a new HTML element being added by the gem when `f.govuk_radio_button` is passed an empty block. In order to avoid this jump, we default back to the old (slightly repetitive) version of the code that has an else condition that uses this method without passing any block. This fixes the issue.

>  block (Block) — Any supplied HTML will be wrapped in a conditional container and only revealed when the radio button is picked [1]

[1] https://www.rubydoc.info/gems/govuk_design_system_formbuilder/GOVUKDesignSystemFormBuilder/Builder#govuk_radio_button-instance_method

The same pattern is still used on checkboxes so there is no similar issue there.

### Before

Note the increase in pixels between both radio options.

Screenshot 2021-02-25 at 16 57 32

Here we can see the new elements being added and shown and hidden as the user toggles their answer:
Screenshot 2021-02-25 at 16 57 53

### After

Screenshot 2021-02-25 at 17 04 13

### DIFF
--- a/app/views/steps/radios.html.erb
+++ b/app/views/steps/radios.html.erb
@@ -11,6 +11,7 @@
       <% if option.fetch("separated_by_or", nil) == true %>
         <div class="govuk-radios__divider">or</div>
       <% end %>
+      <% if option["display_further_information"] %>
         <%= f.govuk_radio_button :response, option["value"].downcase, label: { text: option["value"] }, hint: { text: option["help_text"] } do %>
           <% if option["display_further_information"] == "single" || option["display_further_information"] == true %>
             <%= f.govuk_text_field :further_information, label: { text: option["further_information_help_text"] } %>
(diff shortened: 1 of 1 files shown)


LornaHarwood (approved):
👍 TIL 
🚀

jacksonj04 (approved):

jacksonj04 (commented):
Does this also need replicating in the checkboxes?

tahb:
I think we're good as [checkboxes are still using the non-block form when no further info is present](https://github.com/DFE-Digital/buy-for-your-school/blob/develop/app/views/steps/checkboxes.html.erb#L28) 🤞

PR #197, opened 2021-03-02
(419) Checkbox questions can be explicitly skipped/set as irrelevant

tahb:
The user need here is for a user to "complete" a step where no provided options are relevant to them. We need to be able to distinguish a user who has not yet answered and one who has chosen none are relevant. This allows us to mark the spec builder journey as complete and stop prompting the user to complete a step by showing an "complete" label in the task list.

There is no pattern for this checkbox form in the Design system.

[We had found one in DfE Claim](https://trello.com/c/Tdrjhtde/419-none-of-the-above-radio-option-on-checkbox-lists) but it doesn't appear to work properly. Crucially you cannot deselect the radio option once selected. Label clicks for the radio input also toggle checks to the check box selections too. We decided that it was too much work to try and salvage the pattern now, and a risk to maintaining it in future along with probably needing to use Javascript.

Pete suggested we use a link instead of a button here. I was unable to force a link to send a POST request and got into the muddy territory of creating new controller actions to support it. On future reflection I think a link is conventionally for navigation purposes and buttons are more transactional. Buttons feel correct and were the cleaner implementation 🤞 

## Changes in this PR

- a new `skipCallToAction` field exists for all Contentful questions and sits alongside `PrimaryCallToAction`, both provide copy for buttons
- If this option is present we will store it on the `Step`. We are then able to determine if any `Step` is should be skippable or not via the presence or omission of the button text (this is a bit mysterious but saves us having 2 fields to control this behaviour?)
- checkboxes are the only questions we know of where a skip option is necessary so we hard code this in as an option that is (currently) only something that will affect a checkbox question
- we have to do a little more work to turn validations on for the case where no checkbox or the skip button is used

## Screenshots of UI changes

We introduce a new secondary button, configured with copy from Contentful. We can use the OR separator to provide some vertical grouping of the 2 relevant parts of the form step:
Screenshot 2021-03-02 at 11 05 38

We continue to validate that _either_ one or more options are checked *or* the skip button is used. (sadly I couldn't see how to easily make the form errors wrap both form groups)
Screenshot 2021-03-02 at 11 51 39

Once an answer is created (either by skipping or by a normal choice) the question appears as "Completed" in the task list:
Screenshot 2021-03-02 at 12 01 06

If the user decides to initially store some answers we want to check that a later skip resets this:
Screenshot 2021-03-02 at 11 51 51

After pressing the skip button, we go back in and see that it has been reset.
Screenshot 2021-03-02 at 11 51 58

## Next steps

- [ ] Update the Confluence documentation
- [ ] Let the Content team know of the changes when available

### DIFF
--- a/app/controllers/answers_controller.rb
+++ b/app/controllers/answers_controller.rb
@@ -53,6 +53,8 @@ class AnswersController < ApplicationController
   end
 
   def checkbox_params
+    return {skipped: true, response: nil, further_information: nil} if skip_answer?
+
     answer_params = params.require(:answer)
 
     if @step.options
@@ -73,5 +75,9 @@ class AnswersController < ApplicationController
     answer = params.require(:answer).permit(:response)
     date_hash = {day: answer["response(3i)"], month: answer["response(2i)"], year: answer["response(1i)"]}
     {response: format_date(date_hash)}
+  end
+
+  def skip_answer?
+    params.fetch("skip", false)
   end
 end
--- a/app/models/checkbox_answers.rb
+++ b/app/models/checkbox_answers.rb
@@ -2,7 +2,9 @@ class CheckboxAnswers < ActiveRecord::Base
   self.implicit_order_column = "created_at"
   belongs_to :step
 
-  validates :response, presence: true
+  validates :response,
+    presence: true,
+    unless: proc { |answer| answer.step.skippable? && answer.skipped }
 
   def response=(args)
     return if args.blank?
--- a/spec/services/create_journey_step_spec.rb
+++ b/spec/services/create_journey_step_spec.rb
@@ -133,6 +133,21 @@ process around March.")
       end
     end
 
+    context "when no 'skipCallToAction' is provided" do
+      it "default copy is used for the button" do
+        journey = create(:journey, :catering)
+        fake_entry = fake_contentful_step(
+          contentful_fixture_filename: "steps/skippable-checkboxes-question.json"
+        )
+
+        step, _answer = described_class.new(
+          journey: journey, contentful_entry: fake_entry
+        ).call
+
+        expect(step.skip_call_to_action_text).to eq("None of the above")
+      end
+    end
+
     context "when no 'alwaysShowTheUser' is provided" do
       it "default hidden to true" do
         journey = create(:journey, :catering)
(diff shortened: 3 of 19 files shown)


spec/services/create_journey_step_spec.rb
```diff
@@ -133,6 +133,21 @@
       end
     end
 
+    context "when no 'primaryCallToAction' is provided" do
```
jacksonj04, line 136: The context and description for this test should refer to `skipCallToAction`, and that we're testing the button appears with the given text

app/models/checkbox_answers.rb
```diff
@@ -2,7 +2,7 @@ class CheckboxAnswers < ActiveRecord::Base
   self.implicit_order_column = "created_at"
   belongs_to :step
 
-  validates :response, presence: true
+  validates :response, presence: true, unless: proc { |answer| answer.step && answer.step.skippable? }
```
jacksonj04: Can we ever have an answer without a step?

app/controllers/answers_controller.rb
```diff
@@ -53,6 +53,8 @@ def answer_params
   end
 
   def checkbox_params
+    return {skipped: true, response: [""], further_information: {}} if skip_answer?
```
jacksonj04, line 56: Do we still need to pass a `response` and `further_information` if an answer is skipped, or could we `nil` those out?

app/models/checkbox_answers.rb
```diff
@@ -2,7 +2,9 @@ class CheckboxAnswers < ActiveRecord::Base
   self.implicit_order_column = "created_at"
   belongs_to :step
 
-  validates :response, presence: true
+  validates :response,
+    presence: true,
+    unless: proc { |answer| answer.step && answer.step.skippable? && answer.skipped }
```
jacksonj04, line 7: Is the `answer.step` needed here? We can't have an answer without a step, or if we do then it probably should throw an error.

jacksonj04 (approved):
An elegant solution which neatly sidesteps all the JavaScript woes.

Just a couple of quick non-blocking bits, and one question mark on the naming of a test.

app/models/checkbox_answers.rb
tahb: This was added because of the test suite. For the Shoulda helpers to continue working they would be built without an associated step (the step would be added on `create`). I've remove the Shoulda matchers now and gone manual as you're right, we should never have an answer without a step, in real life.

PR #198, opened 2021-03-02
(426) Fix bug where radio questions with multiple expanded options only saves response to last in sequence

jacksonj04:
## Changes

If there are multiple options within a radio question where expanded information is switched on, only the value from the last option is saved. This is the case even if the input is hidden, which means that unless the user chooses the last option in the list it will appear as their input has not been saved.

We fix this for presenting questions to the user by treating extended radio options like checkboxes and storing a hash of `further_information`. This also means that if a user changes their radio selection at a later date, their original response to further information will also be stored should they change back in future.

For rendering the result in the specification, as we prepare variables for the liquid template we only use the `further_information` value for the key which matches the selected option.

## Screenshots

We check that we can provide answers to both options:
Screenshot 2021-03-16 at 12 33 46
Screenshot 2021-03-16 at 12 33 53

We go back tot he question and check that we open it with the right selected answer _and_ that when we click between the top 2 options, both bits further information are there:
Screenshot 2021-03-16 at 12 35 55
Screenshot 2021-03-16 at 12 34 05

We open the spec and make sure that the variable is available with only the chosen answer in the `further_information` key.
Screenshot 2021-03-16 at 12 35 47

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ The format is based on [Keep a Changelog 1.0.0].
 - further information for options that include special characters is now available to the specification
 - incomplete specification documents include a warning within the document as well as the file name
 - users return to the same place in the task list after answering a question
+- fix extended radio questions so that further_information can be remembered when it can be provided through multiple fields
 
 ## [release-005] - 2021-1-19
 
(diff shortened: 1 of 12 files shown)


tahb (approved):
[Nick has reviewed and said 👍 ](https://ukgovernmentdfe.slack.com/archives/C01R7UF21MG/p1615903667013800?thread_ts=1615890834.005800&cid=C01R7UF21MG)

PR #201, opened 2021-03-03
(423) Link to a new planning guidance page

tahb:
## Changes in this PR

We replicate 2 pages in the prototype with this change:

1. [Our existing start page now includes a link to a new static](https://dfe-hubs-prototype.herokuapp.com/beta/phase-4/start)
2. [A new planning page exists within the service](https://dfe-hubs-prototype.herokuapp.com/beta/phase-4/plan/plan-catering-procurement)

The existing start page in the prototype does not match the existing content we have within the service. I've checked with Nick and Faye that the service is the most up to date version and have added the link into a similar position.

There is quite a lot of content on this page so I've only tested for the accordion buttons (which required some funky xpath code as a `click_on` wouldn't work) and the first paragraph. This felt like good enough confidence but does allow for regressions. Given we aren't expecting to make many changes here at all this felt like a fair trade off. Happy to add more assertions if my confidence isn't shared on this one!

## Screenshots of UI changes

Screenshot 2021-03-03 at 16 01 13
Screenshot 2021-03-03 at 16 01 18
Screenshot_2021-03-03 Catering services


tahb:
I've now learnt these pages need to be linked the other way around so this is back to a WIP

tahb:
I've clarified with Pete. We're back on with this as is!

jacksonj04 (approved):
🚀

PR #209, opened 2021-03-08
(452) Part 1 - Checkbox answers are not automatically joined by commas

tahb:
## Changes in this PR

Instead of performing this operation in Rails, we leave the presentational logic to Contentful.

The same logic can be achieved by defining the following in the `specification_template`:

```
{{ options | join: ", " }}
```

This enables the user to present the options in different ways. Not only as a single paragraph but as a list of options.

## Next steps

- [] Migrate 13 existing checkbox answers to using the new join syntax in Contentful to replicate the previous intention
- [] Document this change in [Confluence](https://dfedigital.atlassian.net/wiki/spaces/GHBFS)

### DIFF
--- a/app/services/get_answers_for_steps.rb
+++ b/app/services/get_answers_for_steps.rb
@@ -42,7 +42,7 @@ class GetAnswersForSteps
         end
       end
 
-      hash["answer_#{step.contentful_id}"] = answer.response.to_s
+      hash["answer_#{step.contentful_id}"] = answer.response
     }
   end
 end
--- a/spec/presenters/checkboxes_answer_presenter_spec.rb
+++ b/spec/presenters/checkboxes_answer_presenter_spec.rb
@@ -2,10 +2,10 @@ require "rails_helper"
 
 RSpec.describe CheckboxesAnswerPresenter do
   describe "#response" do
-    it "returns all none nil values, capitalised as a comma separated string" do
-      step = build(:checkbox_answers, response: ["yes", "no", "morning_break", ""])
+    it "returns all none nil values as an array" do
+      step = build(:checkbox_answers, response: ["Yes", "No", "Morning break", ""])
       presenter = described_class.new(step)
-      expect(presenter.response).to eq("Yes, No, Morning break")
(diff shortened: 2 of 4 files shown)


app/services/get_answers_for_steps.rb
```diff
@@ -42,7 +42,7 @@ def call
         end
       end
 
-      hash["answer_#{step.contentful_id}"] = answer.response.to_s
```
tahb, line 45: This was coercing an array into a string which failed some tests. It's no longer needed as we accept that we will start to return different data types (in this case an array).

spec/presenters/checkboxes_answer_presenter_spec.rb
```diff
@@ -2,10 +2,10 @@
 
 RSpec.describe CheckboxesAnswerPresenter do
   describe "#response" do
-    it "returns all none nil values, capitalised as a comma separated string" do
-      step = build(:checkbox_answers, response: ["yes", "no", "morning_break", ""])
+    it "returns all none nil values as an array" do
+      step = build(:checkbox_answers, response: ["Yes", "No", "Morning break", ""])
```
tahb, line 6: Update the options for realism. We checked recently and found that the `reponse` should contain the human readable/presented strings, rather than sluggified machine readable ones.

tahb:
[Ugh apparently Liquid's `contain` method can only be called on a string, not an array](contains can only search strings. You cannot use it to check for an object in an array of objects). 

I think it might be too far and awkward to expect the spec to also parse the array into a string before being able to use `contains`?:

```
    {% assign selections = answer_7p7URWwZbkjqqzo3E0k1v4 | join: ", " %}
    {% if selections contains "extra" %}
      <p>when there is an answer but no selection was made, this should show</p>
    {% endif %}
```

Perhaps we need to include the joined string and the array :/

tahb:
Mmm I think this may be superceded by the second part of 452 after refactoring the whole approach of how **both** `@answer_x` and `@extended_answer_x` are passing information to Liquid. [Our Slack thread is here for reference](https://dxw.slack.com/archives/C01FBPG2VB7/p1615206489005800)!

PR #210, opened 2021-03-08
(452) Part 1 - Change how information about all answers is shared with the specification builder

tahb:
## Changes in this PR

The variables and their data structure that we make available to the specification through the template has breaking changes.

We remove the concept of `extended_answer` by merging all of its responsibility into `answer` by changing it from a `string` to a `hash`.

We go from:
```
answer_123 => "The option the user picked"
extended_answer_123["response"] => "the_option_the_user_picked"
extended_answer_123["further_information"] => "extra information"
```

To:
```
answer_123 => {
  response: ["Foo", "Bar"],
  concatenated_response: "Foo, Bar",
  skipped: false,
  selected_answers: [
    {
      machine_value: :foo,
      human_value: "Foo",
      further_information: "More yes info"
    },
    {
      machine_value: :bar,
      human_value: "Bar",
      further_information: nil
    }
  ]
}
```

This addresses several issues:

1. the distinction between the human readable text (that the user sees in the browser and selects) and the machine version (used for mapping the `further_information` fields) is clear
2. when no further_information is provided we return `nil` rather than a missing object in the `furhter_information` field
3. provide a structure that allows us to add more configuration that Liquid templating can use, such as `answer["skipped"] == true`

Whilst the `CheckboxAnswerPresenter` class receives the bulk of the changes, every `AnswerPresenter` is also changed with similar (but simpler) changes to return a hash of only the `response`. Offering the same interface for all question types should make it simpler for users in Liquid and provides us an easier way to extend again in future.

There are 32 references to `extended` in the specification template right now. Merging this change will break the spec until they are all converted over.

## Next steps

- [ ] Convert all spec references of old @extended_ variables across to the new format

### DIFF
--- a/app/presenters/checkboxes_answer_presenter.rb
+++ b/app/presenters/checkboxes_answer_presenter.rb
@@ -2,4 +2,33 @@ class CheckboxesAnswerPresenter < SimpleDelegator
   def response
     super.reject(&:blank?)
   end
+
+  def to_param
+    {
+      response: response,
+      skipped: skipped,
+      selected_answers: selected_answers
+    }
+  end
+
+  private def concatenated_response
+    return nil if response.empty?
+
+    response.join(", ")
+  end
+
+  private def selected_answers
+    return [] if response.empty?
+
+    response.each_with_object([]) do |human_readable_choice, array|
+      machine_readable_key = human_readable_choice.parameterize.to_sym
+      matching_further_information = further_information&.fetch("#{machine_readable_key}_further_information", nil)
+
+      array << {
+        machine_value: machine_readable_key,
+        human_value: human_readable_choice,
+        further_information: matching_further_information
+      }
+    end
+  end
 end
--- a/app/presenters/checkboxes_answer_presenter.rb
+++ b/app/presenters/checkboxes_answer_presenter.rb
@@ -6,6 +6,7 @@ class CheckboxesAnswerPresenter < SimpleDelegator
   def to_param
     {
       response: response,
+      concatenated_response: concatenated_response,
       skipped: skipped,
       selected_answers: selected_answers
     }
--- a/app/presenters/checkboxes_answer_presenter.rb
+++ b/app/presenters/checkboxes_answer_presenter.rb
@@ -1,8 +1,5 @@
 class CheckboxesAnswerPresenter < SimpleDelegator
   def response
-    super.reject(&:blank?).map { |answer|
-      answer.tr!("_", " ")
-      answer.capitalize!
-    }.join(", ")
+    super.reject(&:blank?)
   end
 end
--- a/spec/fixtures/contentful/categories/extended-checkboxes-question.json
+++ b/spec/fixtures/contentful/categories/extended-checkboxes-question.json
@@ -39,6 +39,6 @@
             }
           }
         ],
-        "specification_template": "<article id='specification'><h1>Liquid {{templating}}</h1><p>{{answer_extended-checkboxes-question}}</p><ul>{% for extended_answer in extended_answer_extended-checkboxes-question %}<li>{{ extended_answer['response'] }}</li><li>{{ extended_answer['further_information'] }}</li>{% endfor %}</ul></article>"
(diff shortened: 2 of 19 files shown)


app/presenters/checkboxes_answer_presenter.rb
```diff
@@ -6,6 +6,7 @@ def response
   def to_param
     {
       response: response,
+      concatenated_response: concatenated_response,
```
jacksonj04, line 9: This makes sense and we should probably keep it in so we have a backwards-compatible string available for the existing template logic, but I think you _can_ use `contains` on an array - [the documentation](https://shopify.github.io/liquid/basics/operators/#contains) certainly suggests so.

spec/presenters/long_text_answer_presenter_spec.rb
```diff
@@ -16,4 +16,12 @@
       end
     end
   end
+
+  describe "#to_param" do
+    it "returns a hash of long_text_answer" do
+      step = build(:long_text_answer, response: "First line\r\nSecond line")
+      presenter = described_class.new(step)
+      expect(presenter.to_param).to eql({response: "<p>First line\n<br />Second line</p>"})
```
jacksonj04, line 24: TIL we run `simple_format` on long text responses.

spec/services/get_answers_for_steps_spec.rb
```diff
@@ -97,6 +97,21 @@
...
     end
 
+    context "when the answer is of type currency_answer" do
+      it "returns the answer information in a hash" do
+        answer = create(:currency_answer, response: 100.01)
+
+        result = described_class.new(visible_steps: [answer.step]).call
+        assertion = {
+          "answer_#{answer.step.contentful_id}" => {
+            response: "£100.01"
```
jacksonj04, line 94: Future thought, not worth worrying about now: we should also include this value directly as a number.

spec/services/get_answers_for_steps_spec.rb
```diff
@@ -112,6 +112,21 @@
...
     end
 
+    context "when the answer is of type number_answer" do
+      it "returns the answer information in a hash" do
+        answer = create(:number_answer, response: 2)
+
+        result = described_class.new(visible_steps: [answer.step]).call
+        assertion = {
+          "answer_#{answer.step.contentful_id}" => {
+            response: "2"
```
jacksonj04, line 109: As with currency, a future plan should be to include this as a raw integer value as well.

spec/fixtures/contentful/categories/extended-checkboxes-question.json
```diff
@@ -39,6 +39,6 @@
             }
           }
         ],
-        "specification_template": "<article id='specification'><h1>Liquid {{templating}}</h1><p>{{answer_extended-checkboxes-question}}</p><ul>{% for extended_answer in extended_answer_extended-checkboxes-question %}<li>{{ extended_answer['response'] }}</li><li>{{ extended_answer['further_information'] }}</li>{% endfor %}</ul></article>"
+        "specification_template": "<article id='specification'><h1>Liquid {{templating}}</h1><ul>{% for checkbox_answer in answer_extended-checkboxes-question['selected_answers'] %}<li>{{ checkbox_answer['human_value'] }}</li><li>{{ checkbox_answer['further_information'] }}</li>{% endfor %}</ul></article>"
```
jacksonj04, line 42: Would it be better to use the liquid dot notation (`answer_extended-checkboxes-question.selected_answers`)? Or at least to make sure that both continue to work.

jacksonj04 (approved):
A couple of tiny tweaks, some notes for future so they're not just stuck in my brain (trello tasks also created for these), but otherwise let's ship it! :shipit:

app/presenters/checkboxes_answer_presenter.rb
tahb: I tried to `["Yes", "No"] contains "Yes"` but it wasn't having any of it!

spec/fixtures/contentful/categories/extended-checkboxes-question.json
tahb: Good idea. I've now added one both types of accessor into this change so we check for both.

PR #211, opened 2021-03-08
Terraform Cloud Foundry GPaas

Stretch96:
## Changes in this PR

* Adds a Terraform backend using S3
* Adds the `cloudfoundry` provider
* Adds intial Readme for Terraform usage


erbridge:
Should this have an accompanying ADR (or two)?

terraform/app/.terraform.lock.hcl
```diff
@@ -0,0 +1,23 @@
+# This file is maintained automatically by "terraform init".
+# Manual edits may be lost in future updates.
+
+provider "registry.terraform.io/cloudfoundry-community/cloudfoundry" {
+  version     = "0.13.0"
```
tahb, line 5: Is this meant to match or be compatible with what we have in .terraform-version which is `0.14.7`?

terraform/app/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+- go to the application `/terraform` directory
+- install [tfenv](https://github.com/tfutils/tfenv)
+- run `tfenv install` to install the vesion of terraform specified in `.terraform-version`
+- set the `cloudfoundry` provider credentials (GPaas username/password):
+```
+$ export CF_USER="username"
+$ export CF_PASSWORD="password"
+```
+- get the terraform state S3 Bucket AWS credentials
```
tahb, line 19: Would it be possible to give the reader some direction here on how best to discover this? I imagine it'd be one of the following:

- look in the dxw 1password vault for this project
- ask another dev
- look in the CF or GPaaS console(?)

terraform/app/backend.tf
```diff
@@ -0,0 +1,7 @@
+terraform {
+  backend "s3" {
+    bucket = "paas-s3-broker-prod-lon-b2db1a0a-5a37-488b-858f-5e850ff4fdfb"
+    key    = "terraform"
+    region = "eu-west-2"
```
tahb, line 5: I'd be very surprised if all of the DfE stuff was in London. I did a [quick check and other services are using eu-west-2](https://github.com/search?q=org%3ADFE-Digital+eu-west-2&type=code), this may be a default for the ADR

tahb (commented):
I do agree that a short ADR explaining our why we are moving to GPaaS would serve as good documentation.

How about:

```
## Context

The early beta was originally hosted on dxw's Heroku to get delivering quickly whilst access to GPaaS could be set up.

Access to GPaaS with approved billing has now been confirmed so we are migrating the service to its longer term home.

## Decision

Move the service from Heroku to GPaaS for all environments except the ephemeral pull request review environments.

## Consequences

- the service will be easier to maintain in the long term if it is grouped with the rest of DfE's digital services inside GPaaS
- the service will need to be Terraformed as other DfE digital services
- as there are no real users of the service there is less risk to migrating the service now rather than later
- the pull request environments will no longer have strong parity with the live service on an infrastructure level. This can be changed in future if we have move time to also the approach taken by Teaching Vacancies
```

terraform/app/.terraform.lock.hcl
Stretch96: This is the `cloudfoundry` provider version - https://registry.terraform.io/providers/cloudfoundry-community/cloudfoundry/latest/docs

Currently at v0.13.0

terraform/app/README.md
Stretch96: I've put in there 

> (GPaas username/password):

Perhaps before that line we could explain that a GPaas account is required

terraform/app/backend.tf
Stretch96: That region is just where the S3 bucket is, not the terraform aws region
We're not currently using the aws provider (which is where we would specify the region as a variable)

terraform/app/README.md
tahb: Oh, is it the case that the Bucket credentials come from the GPaaS console? I think we should include that an account is required and to ask a DfE Digital org manager for an invitation 👍

tahb (approved):
LGTM once the ADR is in.

terraform/app/README.md
Stretch96: The Bucket credentials are gained from Cloud Foundry, logged in using credentials for a GPaas account, by running:
```
$ cf target -s sct-research
$ cf service-key terraform-state terraform-state-key
```

PR #215, opened 2021-03-09
Checkbox and radio answers store the human readable choice

LornaHarwood:
## Changes in this PR

When a user selects the human readable `Pots, plans or cooking utensils` as a checkbox or radio answer we persist `pots,_pans_or_cooking_utensils` in the database instead of `Pots, plans or cooking utensils`.

`f.govuk_check_box :response` is now set to `option["value"]`, which saves the human readble value, `Pots, plans or cooking utensils`.

This fixes a problem where branching required the content user to pass in a machine readable value, often requiring the content user to have to guess the formatting of this value.

### DIFF
--- a/app/views/steps/checkboxes.html.erb
+++ b/app/views/steps/checkboxes.html.erb
@@ -11,21 +11,21 @@
       <% if option.fetch("separated_by_or", nil) == true %>
         <div class="govuk-radios__divider">or</div>
       <% end %>
-      
+
       <% f.object = monkey_patch_form_object_with_further_information_field(form_object: f.object, associated_choice: machine_value) %>
 
       <% if option["display_further_information"] == true || option["display_further_information"] == "single" %>
-        <%= f.govuk_check_box :response, machine_value, label: { text: option["value"] } do %>
+        <%= f.govuk_check_box :response, option["value"], label: { text: option["value"] } do %>
           <%= f.govuk_text_field "#{machine_value}_further_information",
             label: { text: option.fetch("further_information_help_text", "Optional further information"), hidden: !option.keys.include?("further_information_help_text") } %>
         <% end %>
       <% elsif option["display_further_information"] == "long" %>
-        <%= f.govuk_check_box :response, machine_value, label: { text: option["value"] } do %>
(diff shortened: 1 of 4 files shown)


app/views/steps/checkboxes.html.erb
```diff
@@ -11,21 +11,21 @@
       <% if option.fetch("separated_by_or", nil) == true %>
         <div class="govuk-radios__divider">or</div>
       <% end %>
-
```
LornaHarwood, line 14: I cannot seem to get rid of this spacing change. It appeared after I did a `git pull` and is not something I did myself. If anyone has any suggestions of how to remove this change, please let me know

tahb: Looks like this change is correct and you've found a line that had a different type of invisible characters on it!

If you turn on 'show invisibles' in your Atom settings:
Screenshot 2021-03-09 at 17 45 08

You can see this on `develop`:

Screenshot 2021-03-09 at 17 44 59

We could remove it from this change but I think it's a helpful tidy up :+1:

spec/features/visitors/anyone_can_edit_their_answers_spec.rb
```diff
@@ -59,7 +59,7 @@
   end
 
   context "when the question is checkbox_answers" do
-    let(:answer) { create(:checkbox_answers, response: ["breakfast", "lunch", ""]) }
+    let(:answer) { create(:checkbox_answers, response: ["Breakfast", "Lunch", ""]) }
```
tahb, line 62: I think we can update this in the default Factory values too to remove a future problem for someone https://github.com/DFE-Digital/buy-for-your-school/blob/e5a92f4160da5e1e2948013d5f59999bfc766630/spec/factories/answer.rb#L30

tahb (approved):
LGTM if we can update the factory 👍

spec/features/visitors/anyone_can_edit_their_answers_spec.rb
LornaHarwood: 👍

PR #217, opened 2021-03-09
Terraform Infrastructure

Stretch96:


terraform/workspace-variables/preview.tfvars.example
```diff
@@ -0,0 +1 @@
+environment = "preview"
```
erbridge, line 1: Can we make these files the actual files and make them import a local file for sensitive stuff that has an example?

Stretch96: I've found it easier to be able to keep tfvars in a signle file, or not at all (and use `TF_VAR_` environment variables instead)
The main reason I've done it this way is it follows the pattern that DfE are using on teacher-vacancy-service

terraform/workspace-variables/research_app_env.yml.example
```diff
@@ -0,0 +1,3 @@
+---
+RAILS_ENV: prod
```
tahb, line 2: Should be `research` I think?

terraform/app/web-app.tf
```diff
@@ -0,0 +1,19 @@
+resource "cloudfoundry_app" "web_app" {
+  name                       = "${local.project_name}-${local.environment}"
+  space                      = data.cloudfoundry_space.space.id
+  instances                  = local.web_app_instances
+  disk_quota                 = local.web_app_disk_quota
+  timeout                    = local.web_app_timeout
+  docker_image               = local.web_app_docker_image
+  strategy                   = "blue-green-v2"
```
tahb, line 8: Nice

terraform/workspace-variables/prod.tfvars.example
```diff
@@ -0,0 +1 @@
+environment = "prod"
```
tahb, line 1: Given we default `postgres_class` to the small tier, should we include the "medium" equivalant here in case it's forgotten.

tahb (approved):
From my rusty Terraform eyes this looks good to me, concise 👍

terraform/workspace-variables/research_app_env.yml.example
Stretch96: Ah yep ...

terraform/workspace-variables/prod.tfvars.example
Stretch96: Yup, forgot about that

I'll put in a conditional for the 'prod' environment

terraform/workspace-variables/research_app_env.yml.example
tahb: Wait, no I'm wrong! The `RAILS_ENV` should always be `production` for all live environments as this app only has "development" or "production" as environments.

terraform/app/README.md
```diff
@@ -35,3 +35,36 @@ $ export AWS_REGION=aws_region
...
+  to: `$PROJECT_PATH/terraform/workspace-variables/app/<environment>_app_env.yml`
+- ask the Tech Lead for the environment's variables for `tfvars` and `app_env.yml` files
+- change to the relevant terraform workspace:
+```
+$ terraform workspace list
+$ terraform workspace select <environment>
+```
+- without any changes, ensure that this does not invoke any infrastructure changes:
+```
+$ terraform plan -var-file=<environment>.tfvars
```
tahb, line 57: This should probably be `$PROJECT_PATH/terraform/workspace-variables/app/<environment>.tfvars`

terraform/app/README.md
```diff
@@ -35,3 +35,36 @@ $ export AWS_REGION=aws_region
...
 $ export AWS_SECRET_ACCESS_KEY=aws_secret_access_key
 ```
+- initialise terraform to test the local environment:
+```
+terraform init
+```
+
+### Deploying
+
+- copy: `$PROJECT_PATH/terraform/workspace-variables/<environment>.tfvars.example`
```
tahb, line 45: If the plan is to copy each config file out of the /workspace-variables directory and into app/ before they can be used, it might be better to name it /workspace-variable-examples. I missed the subtle introduction of /app here

Stretch96: Yep seems a better name for it
This does however currently mirror what DfE have named it in teacher-vacancies, but won't be too far out of what they have 👍

.gitignore
```diff
@@ -108,3 +108,7 @@ terraform.rc
 *.sql
 *.csv
 *.xls
+
+# App env yaml files
+*_app_env.yml
```
tahb, line 113: This didn't ignore my app env file when copied to terraform/app/staging_app_env.yml

PR #219, opened 2021-03-10
Radio further information fields no longer auto-generates it's own label

LornaHarwood:
## Changes in this PR

When `further_information_help_text` isn't provided, it now no longer auto-generates it's own label. 
Instead it doesn't show a label at all.

### Before

Screen Shot 2021-03-10 at 12 40 56

### After

Screen Shot 2021-03-10 at 12 43 40


jacksonj04 (approved):
Bug: squished!

PR #222, opened 2021-03-10
(434) Specification is shown on its own page

tahb:
## Changes in this PR

Move the specification, the warning and the download button away from the task list and onto its own page.

Include a back link so the user can return to the task list, using the pattern we added for the planning start page.

## Screenshots of UI changes

Screenshot_2021-03-10 Catering
Screenshot_2021-03-10 Catering(1)


jacksonj04 (approved):
I'm happy with this, but Brakeman isn't! It just needs the ignore rule moving over (since we do the rendering of XSS-possible content in a new place (we're happy that this isn't a risk, since the content originates from a controlled space).

PR #223, opened 2021-03-11
(465) Further information is now saved for options with commas (or other special characters)

tahb:
## Changes in this PR

We were accidentally blocking further information `params` as the key didn't match an expected pattern. This caused them not to be saved on the record and the choice appeared to be ignored because we don't raise when an unexpected param was found (maybe we should?).

Flesh out the feature test to cover this regression with the "Other, please specify" option that's used a lot on live.

## Screenshots of UI changes

This value is now being persisted and is visible when returning to the question:

Screenshot 2021-03-11 at 11 56 15
Screenshot 2021-03-11 at 11 56 22


jacksonj04 (approved):
Good to get this hole in the tests plugged alongside the bug fix!

PR #227, opened 2021-03-15
(398) Specification document includes incomplete warnings

tahb:
## Changes in this PR

We change the filename and include a simple paragraph at the top of the document which cannot intentionally or accidentally  reordered by the `specification_template`.

[The language used is taken from this document that the content team prepared](https://docs.google.com/document/d/11BOARCY9u6p846oS_NGq0hm3RE2EOOcBzFUyjarejt0/edit).

I'm unhappy with the feature test and the subsequant lack of complete coverage we have. [I've explained the different options I went through on the commit](https://github.com/DFE-Digital/buy-for-your-school/pull/227/commits/d4bab9cef9c8df183afb7e7756f3a9467c023e78). If there's a way anyone can think to improve this without changing the testing stack to include a non headless config I'd love to hear it!

### Before

Note the filename doesn't include "incomplete":

Screenshot 2021-03-15 at 15 36 14

No warning at the top of the file:

Screenshot 2021-03-15 at 15 36 22

### After

Screenshot 2021-03-15 at 15 02 44

Screenshot 2021-03-15 at 15 13 27

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ The format is based on [Keep a Changelog 1.0.0].
 - the specification lives on it's own page, separate to the task list
 - fix checkbox questions where further information couldn't be saved for an option that included a special character
 - further information for options that include special characters is now available to the specification
+- incomplete specification documents include a warning within the document as well as the file name
 
 ## [release-005] - 2021-1-19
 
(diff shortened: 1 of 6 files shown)


jacksonj04 (approved):
On the testing front I think a .docx file is actually a zip of the XML and associated assets. We could unzip it and peek inside, but at that point we're pretty much just testing to see if the gem is capable of doing its job of rendering HTML.

I think I'm fine with this as-is. It's a testing hole that isn't ideal, but at the same time we could (theoretically) have all kinds of user-generated content which could cause unwanted behaviour in the .docx renderer and we can't realistically test for them all. We just trust that the gem, if called with HTML, will turn it into a document.

PR #231, opened 2021-03-17
Add MIT Licence

tahb:
DfE work in the open and MIT is the default license [1]

[1] https://github.com/DFE-Digital/technical-guidance/blob/master/source/documentation/standards/licencing-software-or-code.html.md.erb#L15

### DIFF
--- a/LICENCE
+++ b/LICENCE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Crown Copyright (Department for Education)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


jacksonj04 (approved):
Should we be copyright 2021 instead?

tahb:
@jacksonj04 I started with that but changed it after [closer inspection of DfE's guidance](https://github.com/DFE-Digital/technical-guidance/blob/master/source/documentation/standards/licencing-software-or-code.html.md.erb#L41), we open sourced in 2020 so I picked that year instead. We could do 2020-2021 but it's not clear to me that that's required

jacksonj04:
I think we probably have enough updates to justify 2020-2021, but that then just adds a bit of maintenance burden so I really don't feel that strongly.

Ship it!

PR #233, opened 2021-03-17
Flash messages can be shown as notification banners

tahb:
## Changes in this PR

- We add the existing (but unused) messages partial into the layout
- Flash messages can now be set by controllers that will appear as an undissmisable notification banner
- Use design system styling for the notificaiton component

Screenshot 2021-03-17 at 16 38 24

### DIFF
--- a/app/views/layouts/_messages.html.erb
+++ b/app/views/layouts/_messages.html.erb
@@ -1,9 +1,16 @@
-<% # Rails flash messages styled for Bootstrap 4.0 %>
 <% flash.each do |name, msg| %>
   <% if msg.is_a?(String) %>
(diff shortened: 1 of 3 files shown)


app/views/layouts/_messages.html.erb
```diff
@@ -1,9 +1,16 @@
...
+          <%= name.capitalize %>
+        </h2>
+      </div>
+      <div class="govuk-notification-banner__content">
+        <h3 class="govuk-notification-banner__heading">
+          <%= content_tag :div, msg, id: "flash_#{name}" %>
+        </h3>
+      </div>
     </div>
+    <% end %>
```
jacksonj04, line 15: There looks to have been an unexpected indentation change here.

jacksonj04 (approved):
One whitespace change I don't think should happen, but otherwise LGTM.

PR #234, opened 2021-03-17
(477) Users have to sign in with DfE Sign-in

tahb:
There isn't documentation from DSI on how we should implement an interface with them that follows the OpenID Connect standard. I've used my [previous work on Teaching Vacancies](https://github.com/DFE-Digital/teaching-vacancies/pull/335) and the [good work done on Apply](https://github.com/DFE-Digital/apply-for-teacher-training/pull/507) as a guide. Alongside this, generally pushing buttons in the DfE Sign-in console and trying to configure it without exact guidance on what each means.

## Changes in this PR

- refactor documentation to record how DfE Sign-in is configured
- introduce a basic concept of a user that only records the DSI UID (as that's all we need)
- users have to sign in with DfE Sign-in when they start the journey
- users can't access pages apart from the start pages without authentication
- machines can still access the `health_check` endpoint without authentication
- we can use Omniauth without DfE Sign-in in development to get started quickly (we also don't have another environment in DSI to configure off of)

⚠️ The preview environment will break after merging this change. Shall we make a task in Trello to figure out how to set up a new DSI environment or do it before meging this change? I don't think preview is being used a lot now all the questions are in so I'd opt for the latter

⚠️ I've ignored a high Brakeman warning because I think it's a false positive, [please double-check my commit for my thinking](https://github.com/DFE-Digital/buy-for-your-school/pull/234/commits/cbd61c94acfbffccb96b60fa216b6ac0a0800351)

⚠️ I've left DSI configured to localhost to allow you to review this PR with a note to change it to staging before/after merging. It's mentioned else where but you can toggle it on or off locally with `DFE_SIGN_IN_ENABLED`

## Screenshots of UI changes

Screenshot_2021-03-22 Create a specification to procure a catering service for your school
Screenshot_2021-03-22 DfE Sign-in

If you click sign out

Screenshot_2021-03-22 Catering

You are signed out of our service and DSI 

Screenshot_2021-03-22 Create a specification to procure a catering service for your school

You have to sign back into DSI

Screenshot_2021-03-22 DfE Sign-in

---

In development we have a different journey that includes a button suffix to remind everyone:

Screenshot_2021-03-22 Create a specification to procure a catering service for your school(2)
Screenshot_2021-03-22 User Info(1)

## Next steps

- [x] Set `APPLICATION_URL` on all environments
- [x] Change DfE Sign in test-integration environment to using the staging environment
- [x] Invite the team to DSI test integration environment so they can sign into staging post merge


jacksonj04 (approved):
This looks solid to me.

I agree that the Brakeman warning is safe to ignore, since we only do some very specific things with the incoming data in a controlled manner.

For the time being the scopes look good, although in future we may have to extend this to capture email addresses if we need to start emailing users. For the time being though, best to not even worry about them.

PR #235, opened 2021-03-18
Fix checkbox answer to save choices after previously skipping

LornaHarwood:
## Changes in this PR

When a checkbox answer was skipped, any later updates to this answer were not saving. `further_information_params` was only setting the skipped value to true. It now reassigns the skipped value to false when any checkbox choices are later selected.


jacksonj04 (approved):
Solution looks good!

I'd be tempted to reword the commit message so that it describes the problem and the solution separately, eg:

> The `skipped` value was being set to `true` when a checkbox question was skipped, but the opposite case didn't set it to `false`. This meant that once a question had been skipped once it would never become not skipped.
>
> This commit adds setting `skipped` to `false` in `further_information_params` for checkbox questions.

PR #238, opened 2021-03-22
Add a new dashboard page with ability to create new specifications

LornaHarwood:
## Changes in this PR

After sign in, users are now sent to the Specification Dashboard (instead of starting a new spec), from where they can choose to create a new specification.

A subsequent PR will add the ability to view existing specifications.

## Screenshots of new UI

localhost_3000_dashboard (1)

### DIFF
--- a/app/views/journeys/show.html.erb
+++ b/app/views/journeys/show.html.erb
@@ -1,5 +1,6 @@
 <%= content_for :title, @journey.category.capitalize %>
 
+<%= link_to I18n.t("generic.button.back"), dashboard_path, class: "govuk-back-link" %>
 <h1 class="govuk-heading-xl"><%= I18n.t("specifying.start_page.page_title") %></h1>
 
 <%= render "resume_journey_notice" %>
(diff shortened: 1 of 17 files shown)


tahb (commented):
Looks good to me.

I notice that when we click the link to start a journey that there's no way back to the dashboard to start again. Before this we were able to click the service name in the header to get back but now we have a new page in the journey that requires 2 clicks to get back to. Should we add some more back navigation, what do you think?

app/views/journeys/show.html.erb
```diff
@@ -1,5 +1,6 @@
 <%= content_for :title, @journey.category.capitalize %>
 
+<%= link_to I18n.t("generic.button.back"), dashboard_path, class: "govuk-back-link" %>
```
tahb, line 3: Should we have a feature test for this?

LornaHarwood: I have added a test for this within the `anyone_can_view_a_list_of_tasks_spec`, as this feels like the most logical place to put it

tahb (approved):
LGTM

PR #239, opened 2021-03-22
Existing specifications can be viewed from the dashboard

jacksonj04:
**This is based atop #238, which should be merged first.**

## Changes in this PR

* There is a new view which lists all existing specifications
* The dashboard now has a new link to the view of existing specifications
* Every new journey now belongs to a a single User
* Users can only see the journeys (and associated steps and answers) that they belong to
* Improve `CreateJourney` service so it fails loudly if for some reason it can't create a journey we can use

## Screenshots of UI changes

Log in as User A:
Screenshot_2021-03-22 User Info(1)
View a single attached journey:
Screenshot_2021-03-29 Existing specifications

Log in as User B
Screenshot_2021-03-29 User Info
View does not contain journey from User A
Screenshot_2021-03-29 Existing specifications(1)

## Next steps

* It should be easier to identify which specification is which
* Distinguish between 'in progress' and 'completed' specifications

### DIFF
--- a/spec/features/visitors/anyone_can_sign_in_with_dfe_sign_in_spec.rb
+++ b/spec/features/visitors/anyone_can_sign_in_with_dfe_sign_in_spec.rb
@@ -22,17 +22,15 @@ feature "Anyone can sign in with DfE Sign-in" do
 
       user_exists_in_dfe_sign_in(dsi_uid: user.dfe_sign_in_uid)
       user_starts_the_journey
-
       expect(page).to have_content(I18n.t("specifying.start_page.page_title"))
-
-      visit root_path
 
       # Undo the OmniAuth stub to check we don't require it again
       OmniAuth.config.mock_auth[:dfe] = OmniAuth::AuthHash.new(foo: :bar)
 
-      answer = create(:short_text_answer)
-      visit edit_journey_step_path(answer.step.journey, answer.step)
-      expect(page).to have_content(answer.step.title)
+      journey = create(:journey, user: user)
+      step = create(:step, :radio, journey: journey)
+      visit journey_step_path(journey, step)
+      expect(page).to have_content(step.title)
     end
   end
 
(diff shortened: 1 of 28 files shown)


spec/features/visitors/anyone_can_sign_in_with_dfe_sign_in_spec.rb
```diff
@@ -22,17 +22,15 @@
...
-
       # Undo the OmniAuth stub to check we don't require it again
       OmniAuth.config.mock_auth[:dfe] = OmniAuth::AuthHash.new(foo: :bar)
 
-      answer = create(:short_text_answer)
-      visit edit_journey_step_path(answer.step.journey, answer.step)
-      expect(page).to have_content(answer.step.title)
+      journey = create(:journey, user: user)
+      step = create(:step, :radio, journey: journey)
+      visit journey_step_path(journey, step)
```
jacksonj04, line 32: This has changed from `edit_journey_step_path` to `journey_step_path`. It doesn't have any material effect on the intent of the test, this is just a note for those who come after.

spec/requests/authorisation_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+RSpec.describe "Authorisation", type: :request do
+  describe "Users can only see journeys they belong to" do
+    context "when the journey IS theirs" do
+      it "returns 200" do
+        journey = create(:journey)
+        user_is_signed_in(user: journey.user)
+
+        get journey_path(journey)
+
+        expect(response).to have_http_status(200)
```
jacksonj04, line 12: Could we go with `:ok` instead of `200` so this matches the negative case?

tahb (approved):
[Nick has reviewed and approved](https://dxw.slack.com/archives/D019B564VT8/p1617100019000300). Since they started this pull request (and I continued it) I have to be the one to click the approve button on their behalf.

PR #241, opened 2021-03-23
(469) Cache invalidation end point so caching can be left on

tahb:
At the moment the service is dangerously close to taking 60 seconds when the user starts the journey. This has started to lead to timeouts which prevents anyone from using the service. 

We do have a caching feature to help mitigate against the high number of requests we need to make to Contentful, however this has been deliberately turned off. Caching was disabled to allow the content team so they could see their updates (made in Contentful) within the service straight away. The need for quick feedback was mostly to do with checking branching logic which cannot be checked within the preview environment.

If we can ensure that changes to entries are updated quickly within the service, we will be able to keep caching enabled and solve both problems.

To do this we think we can create a new Contentful webhook to send us a message when any entry is changed. We'll then look for that entry ID in our cache and delete it if found.

## Changes in this PR

- a new endpoint which recieves a webhook from Contentful when any of its' entries are changed in any way and invalidates any matching cache entry
- the endpoint is named generically as we may want to do more in future when this type of event occurs. It doesn't have to be closely coupled with cache invalidation and the controller is open to extend with more functionality later if needed
- basic auth added as a minimal way to prevent unfettered access to this endpoint as it would allow infinite requests to `Redis.del(params)` which could cause a Redis DOS or make us vulnerable should the endpoint be asked to do more in the future

⚠️ Removing entries from the cache made me think about what happens if we replace that entry with one that is invalid. What would it break? How would we recover? I think there are a few points to discuss:

1. The content team aren't using the preview environment to test branching, they are using staging which leaves it in a potentially broken state. Can/should we investigate how to help them using the Preview environment for this? When moving to production the impact of getting it wrong is increased
2. We don't validate entries before we cache them. An entry with invalid JSON for any of the advanced options can cause odd behaviour. For example if "previous_identifiers" is spelt wrong then branching will silently stop working as it will be ignored.

I don't think this code introduces this problem, it allows us to turn caching on which is good but we should be careful this isn't viewed as papering over those risks.

## Next steps

- [ ] Add the API key into each environment
- [ ] Enable caching in all environments


tahb:
@LornaHarwood I made those amendments to the variable names and included a better PR description as it was definitely lacking the context of the Trello card.

lozette (approved):
Makes sense to me!

PR #251, opened 2021-03-26
Load REDIS_URL and other GPaaS variables before initialisation

tahb:
* The previous implementation had the vcap_parser called in an initialiser. Rails calls these files in alphabetical order [1] which meant it was being run after the redis initialiser. Causing REDIS_URL to be missing
* The Vcap parser only needs to run in the live environment as that's the only place `ENV["VCAP_SERVICES"]` will ever exist.
* I could have numbered the files and ordered vcap before session store but this felt fragile. Someone could easily change the order without realising the consequences.

[1] https://guides.rubyonrails.org/configuring.html#using-initializer-files

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
+# Load environment variables that are created by GPaaS
+require_relative "../../lib/vcap_parser"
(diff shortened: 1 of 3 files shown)


Stretch96 (approved):
LGTM 👍

PR #254, opened 2021-03-30
(490) Automatically remove stale Journey records

tahb:
Everytime a user starts a specification journey within the service it creates 100-200 database rows. Each row represents a copy of Contentful content, most commonly a question. We make this copy to ensure that the journey remains future proof if the content in Contentful changes (versioning) and if the Contentful service becomes unavailable.

We are finding that we are creating approximately 10,000 a week on staging on our Heroky hobby database which is the limit. Whilst our live database will have millions of rows the problem remains that many thousands of rows will become occupied over time and there is a big potential that many of them will serve no value. There is no option or need for users to delete old specifications we present in the dashboard so for now the service is going to slowly fill up and undoubtedly in 6months/1year's time the database limit will be hit, long after we have stopped developing.

The starting definition is the Journey isn't association with a _single_ answer *and* it hasn't been worked on in the last month. The time period is arbitrary so we can increase it to become more defensive if we are concerned we might accidentally delete journeys where users have invested their time and effort. I think with the stale check, 1 month is fine. If the user comes back they are likely going to create a new journey, if they look for an old one and don't find it, they will hopefully find the new call to action and not be frustrated since no effort is lost on their part.

## Changes in this PR

- add some state to the Journey to avoid ugly and expensive database queries to find all qualifying journeys
- add a service that will contain the definition of what stale means
- add a new job that runs nightly to remove all stale journeys (this could be run less frequently but I think running it often while the dev team is on hand is sensible for relatively little cost in having it run every night)

### DIFF
--- a/app/models/journey.rb
+++ b/app/models/journey.rb
@@ -13,6 +13,7 @@ class Journey < ApplicationRecord
 
   def freshen!
     attributes = {}
+    attributes[:last_worked_on] = Time.zone.now
     attributes[:started] = true unless started == true
 
     update(attributes)
--- a/app/models/journey.rb
+++ b/app/models/journey.rb
@@ -10,4 +10,11 @@ class Journey < ApplicationRecord
   def all_steps_completed?
     visible_steps.all? { |step| step.answer.present? }
   end
+
+  def freshen!
+    attributes = {}
+    attributes[:stale] = false unless stale == false
+
+    update(attributes)
+  end
 end
--- a/app/models/journey.rb
+++ b/app/models/journey.rb
@@ -13,7 +13,7 @@ class Journey < ApplicationRecord
 
   def freshen!
     attributes = {}
-    attributes[:stale] = false unless stale == false
+    attributes[:started] = true unless started == true
 
     update(attributes)
   end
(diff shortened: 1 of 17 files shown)


app/models/journey.rb
```diff
@@ -9,4 +9,9 @@ class Journey < ApplicationRecord
   def all_steps_completed?
     visible_steps.all? { |step| step.answer.present? }
   end
+
+  def fresh!
```
lozette, line 13: We have a query about the name `fresh!` - this implies it's a query, when what this is doing is "freshening" the record. Maybe `freshen` or `touch` is better

db/migrate/20210330105313_add_stale_state_to_journey.rb
```diff
@@ -0,0 +1,6 @@
+class AddStaleStateToJourney < ActiveRecord::Migration[6.1]
```
lozette, line 1: We think the name `stale` doesn't accurately represent the state of the journey. Maybe something like `has_answers` or `started` is better. `stale` implies it hasn't been worked on for a while, but that's represented by the `last_worked_on` property from a later commit

lozette (changes requested):
A couple of comments!

tahb:
@lozette those changes are done and I added a new commit to make the date range configurable for staging.

lozette (approved):

tahb:
I suspect that this job won't run on staging until this error is properly understood https://github.com/DFE-Digital/buy-for-your-school/pull/279.

Holding pattern ✈️

PR #258, opened 2021-03-31
Debug GitHub Action deployment failure

tahb:
Staging deployments aren't working with GitHub Actions https://github.com/DFE-Digital/buy-for-your-school/runs/2235529241?check_suite_focus=true

This looks to be because the new docker image is not set: `- docker_image               = "***:latest" -> null`

Try using the syntax found in this documentation to access the docker image name of the image we built in the previous job: https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#using-workflow-commands-to-access-toolkit-functions

I see there's a warning about this variable which may also be to blame that GitHub Actions thinks it's a secret (which it isn't) 
Screenshot 2021-03-31 at 10 37 45

This points to the case where it GitHub may think it contains a secret. Some or all of the variable we are setting is probably being included in `.secrets` somewhere so we can share that value between jobs. Add echo commands in to see which piece of information that contributes to the construction of `TF_VAR_docker_image` is being accidentally obfuscated.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -25,9 +25,15 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v2
       - name: Set Docker tag environment variable
-        run: echo "DOCKER_TAG=${GITHUB_RUN_ID}-${GITHUB_SHA}" >> $GITHUB_ENV
+        run: |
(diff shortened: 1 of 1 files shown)


jacksonj04 (approved):
🐛 🔍

PR #265, opened 2021-04-01
Add Sections as a model in the application

lozette:
Currently, only Journeys & Steps are models in the application, and Sections
are stored as a json blob `section_groups` on Journeys.

We want to add a new model type to Contentful & the application between Section
and Step ("Tasks") and to do so it makes sense to add Sections as a concrete
model in the application, and later add Tasks as a model as well.

This PR is the first step in the process. Sections are added as a model,
attached to a Journey. Eventually we will add tasks as a `has_many` association
to Sections, and Steps as a `has_many` association to Tasks.

### DIFF
--- a/app/models/journey.rb
+++ b/app/models/journey.rb
@@ -1,6 +1,7 @@
 class Journey < ApplicationRecord
   self.implicit_order_column = "created_at"
   has_many :steps
+  has_many :sections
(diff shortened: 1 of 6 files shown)


LornaHarwood (approved):
LGTM 👍 🚀

PR #266, opened 2021-04-01
Add cdn-routes

Stretch96:
## Changes in this PR

Adds cdn routes for the `prod` and `staging` environments

The cdn-routes had already been created, and they have now been imported into terraform

Routes have been added to point:
`prod`: `get-help-buying-for-schools.education.gov.uk`
`staging`: `staging-get-help-buying-for-schools.education.gov.uk`


terraform/app/locals.tf
```diff
@@ -5,6 +5,9 @@ locals {
   cloudfoundry_org           = var.cloudfoundry_org
   cloudfoundry_space         = local.environment == "prod" ? "sct-production" : "sct-${local.environment}"
   shared_cloudfoundry_domain = var.shared_cloudfoundry_domain
+  custom_cloudfoundry_domain = var.custom_cloudfoundry_domain
+
+  custom_hostname = local.environment == "prod" ? "get-help-buying-for-schools" : "${local.environment}-get-help-buying-for-schools"
```
tahb, line 10: Since we only have prod and staging domains, when we deploy this to the `research` or `preview` will they or their default hostname break and become unaccessible?

tahb (approved):
Something could be wrong with the `custom_hostname`? I gather this has already been deployed and I can see the new domain works so if you think there's nothing to worry about then we can merge.

PR #270, opened 2021-04-07
Add noindex,nofollow meta tag to all pages

lozette:
As per Gov.uk guidelines, all pages on the application should have a "robots"
meta tag with the content "noindex,nofollow"

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -10,6 +10,7 @@
     <meta name="theme-color" content="#0b0c0c">
 
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="robots" content="noindex,nofollow">
 
     <%= favicon_link_tag %>
     <%= favicon_link_tag 'govuk-mask-icon.svg', rel: 'mask-icon', type: 'image/svg+xml', color: '#0b0c0c' %>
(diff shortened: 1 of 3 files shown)


app/views/layouts/application.html.erb
```diff
@@ -10,6 +10,7 @@
     <meta name="theme-color" content="#0b0c0c">
 
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="robots" content="noindex,nofollow">
```
tahb, line 13: I forget how this works! How does this work in relation to robots.txt, is this change enough or do we also need an entry in robots.txt?

lozette: No, as per the guidance: 

```
Telling search engines not to index pages
You can either:

serve a meta tag on every page with the value: noindex, nofollow
use a X-Robots-Tag response header with the value: noindex, nofollow
Do not use a robots.txt file to prevent crawling of your service. This will prevent the crawler from seeing the noindex directive, so pages on your service may still appear in search results.
```

https://www.gov.uk/service-manual/technology/get-a-domain-name

tahb: Makes sense 👍 

I think including those guidelines would be useful context to keep in commit message too.

tahb (approved):

app/views/layouts/application.html.erb
lozette: Added

PR #271, opened 2021-04-07
Do not allow access to the Preview endpoints on non-preview environment

lozette:
The "preview" endpoints are for the Preview environment only. Hitting this
endpoint on the non-preview environment creates cruft entries in the database.

Resolve this by checking on the sole preview controller that we are on the 
Preview environment. If we are not, return `not_found`

### DIFF
--- a/app/controllers/preview/entries_controller.rb
+++ b/app/controllers/preview/entries_controller.rb
@@ -1,4 +1,6 @@
 class Preview::EntriesController < ApplicationController
+  before_action :check_app_is_running_in_preview_env
+
   def show
     @journey = Journey.create(
       category: "catering",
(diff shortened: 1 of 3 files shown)


app/controllers/preview/entries_controller.rb
```diff
@@ -19,4 +21,9 @@ def show
   def entry_id
     params[:id]
   end
+
+  def preview_environment?
+    return if ENV["CONTENTFUL_PREVIEW_APP"].eql?("true")
```
tahb, line 26: `return` if not given anything will return `nil` which is fine as a guard however the method name is using the `?` convention which implied to me a `true` or `false` would come back. In reality it returns `nil` or `render` instead.

The behaviour is right so I suspect a method rename here would do. Perhaps `check_app_is_running_in_preview_environment` or something less long!

lozette: Amended!

tahb (approved):

PR #275, opened 2021-04-08
In development DSI is disabled by default

tahb:
DSI doesn't work in development unless we have DSI environment configured to work with localhost.

So that the app starts for new devs with the minimum amount of fuss, disable this so the OmniAuth developer mode is used for signing in.


lozette (approved):
👍

PR #276, opened 2021-04-09
(568) Sanitise user answers before persisting and on rendering

tahb:
## Changes in this PR

- refactor how params are passed to `SaveAnswer` depending on the different types of `Answer` before we begin extending the `SaveAnswer` with sanitisation logic
- strip unsafe tags from all user answers so that we only store safe values in the database. If an answer.response is used with `.html_safe` then no script will be executed in the clients browser

User input given to LongTextAnswer was being unintentionally escaped by using `.to_param` within 
[GetAnswersForSteps](https://github.com/DFE-Digital/buy-for-your-school/blob/develop/app/services/get_answers_for_steps.rb#L26).

User input given to `ShortTextAnswer` or `further_information` fields within `Radios` or `CheckboxAnswers` was not being stripped by `to_param` in the same way so scripts were executed when viewing the spec. 

`CurrencyAnswer` and `NumberAnswer` have validation for numerical input only so were safe.

This change does introduces several new changes:

1. unsafe tags like `<a>`, `<script>` and `` are stripped out along with anything else except those on our whitelist (more on this next)
1. tags `<p>`, `<b>` and `<i>` are allowed since they are stored for `LongTextAnswer` to render line breaks - we want to keep this

### Before

I add `<script>alert();</script>` into `LongTextAnswer`, `ShortTextAnswer` and `CheckboxAnswers` (with further info).

We can see that before _only_ long text was escaped by `.to_param`:

Screenshot 2021-04-14 at 16 17 28
Screenshot 2021-04-14 at 16 17 32

### After

Now all question types strip the <script> tag and output the contents safely:
Screenshot 2021-04-14 at 16 18 31

### DIFF
--- a/app/services/string_sanitiser.rb
+++ b/app/services/string_sanitiser.rb
@@ -9,15 +9,24 @@ class StringSanitiser
       if value.is_a?(Array)
         value = value.map do |sub_value|
           if sub_value.is_a?(String)
-            sub_value = Rails::Html::FullSanitizer.new.sanitize(sub_value)
+            sub_value = sanitize(sub_value)
           end
           sub_value
         end
       elsif value.is_a?(String)
-        value = Rails::Html::FullSanitizer.new.sanitize(value)
+        value = sanitize(value)
       end
 
       value
     end
+  end
+
+  private def sanitize(value)
+    safe_list_sanitizer = Rails::Html::SafeListSanitizer.new
+    safe_list_sanitizer.sanitize(value, tags: allowed_html_tags)
+  end
+
+  private def allowed_html_tags
+    %w[p b i]
   end
 end
--- a/app/services/string_sanitiser.rb
+++ b/app/services/string_sanitiser.rb
@@ -1,0 +1,23 @@
+class StringSanitiser
+  attr_accessor :args
+  def initialize(args:)
+    self.args = args
+  end
+
+  def call
+    args.deep_merge(args) do |_, _, value|
+      if value.is_a?(Array)
+        value = value.map do |sub_value|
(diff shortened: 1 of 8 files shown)


tahb:
Confirmed that normal text inputs were not vulnerable but further information fields were: 
Screenshot 2021-04-14 at 14 38 53

Will work on a fix for this to sanitise this on the way out too, as well as what we have already here for making it safe on the way in.

app/services/string_sanitiser.rb
```diff
@@ -9,15 +9,24 @@ def call
...
     end
   end
+
+  private def sanitize(value)
+    safe_list_sanitizer = Rails::Html::SafeListSanitizer.new
+    safe_list_sanitizer.sanitize(value, tags: allowed_html_tags)
+  end
+
+  private def allowed_html_tags
+    %w[p b i]
```
jacksonj04, line 30: I'm happy with these tags being allowed, although the chances of users adding their own `<b>` etc is slim!

The "easy to read" purist in me wonders if `["p", "b", "i"]` is a more obvious way to express this without relying on knowing what `%w[]` does, but that's very weakly held.

jacksonj04 (approved):
📦

app/services/string_sanitiser.rb
tahb: Agree that <b> is slim but for the sake of one letter we don't have to come back and add it later. 

We do use the %w[] syntax in 3 other places at present so if it's a styling change we want for readability and consistency then I think it's fair to do this later as a sweeping change. As neither feel strongly I'll let this one go for now 👍

PR #283, opened 2021-04-13
(569) Fix Sidekiq connection failure due to NotImplementedError

tahb:
The [stacktrace from the error message](https://rollbar.com/dxw/dfe-buy-for-your-school/items/131/) for NotImplementedError wasn't helpful so this change is based on trial and error. 

I was able to confirm that Sidekiq was not connecting to Redis. On the Heroku research environment this was further confused by the fact the worker wasn't enabled at all. I stripped the various configuration parts out until finding that removing `sidekiq_options` fixed the issue. This is surprising because [Sidekiq include this in the documentation for setting up a job with ActiveJob](https://github.com/mperham/sidekiq/wiki/Active-Job#customizing-error-handling).

I checked that the Sidekiq patch upgrade that was applied 5 days didn't introduce this issue.

There isn't a suggested way to configure retry limits so with this removal the default retry limit will be the Sidekiq default which is 28 days.

We should probably consider Sidekiq health checks. Leaving the Rollbar info debugging in for now so we can be confident it's fixed on gpaas too.

## You can test this by:

```
# ssh to research where this commit is currently present
$ heroku run rails c -a dfe-bfys-beta-research

# Check the cache in Redis 1
$ Redis.new(url: "#{ENV['REDIS_URL']}/1").keys

# Clear any cached keys if they exist so we start a fresh
$ Redis.new(url: "#{ENV['REDIS_URL']}/1").flushdb

# Run the task asynchronously
$ WarmEntryCacheJob.perform_later

# Check to see if there are lots of nice new cache keys
$ Redis.new(url: "#{ENV['REDIS_URL']}/1").keys
```

PR #287, opened 2021-04-15
(490) Delete stale journey does not raise NotImplementedError

tahb:
## Changes in this PR

We saw recently on our first job that including sidekiq_options threw a `NotImplementerError`. We fixed it by removing the retry configuration but the root cause remains a mystery. https://github.com/DFE-Digital/buy-for-your-school/pull/283

The same error happens here so to quickly get moving we remove the retry count and use the Sidekiq defaults. https://rollbar.com/dxw/dfe-buy-for-your-school/items/107/occurrences/160269956473/. If we really do need retry in the future we can invest more time in figuring out why this happens. It is presumably a problem with sidekiq or some conflicting configuration we have that we haven't found yet.

This change should mean this asynchronous starts running at 2am again.

### DIFF
--- a/app/jobs/delete_stale_journeys_job.rb
+++ b/app/jobs/delete_stale_journeys_job.rb
@@ -3,5 +3,7 @@ class DeleteStaleJourneysJob < ApplicationJob
 
   def perform
     DeleteStaleJourneys.new.call
+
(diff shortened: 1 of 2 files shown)


jacksonj04 (approved):
Yet another mystery!

PR #288, opened 2021-04-15
Add header and footer information for feedback and data requests

LornaHarwood:
## Changes in this PR

Add header and footer information for feedback and data requests

## Screenshots of UI changes

Header

Screen Shot 2021-04-16 at 14 58 31

Footer

Screen Shot 2021-04-16 at 17 09 49

## Next Steps

- [ ] Add ENV variable to Staging
- [ ] Add ENV variable to Research
- [ ] Add ENV variable to Production

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,6 +17,8 @@ REDIS_URL=redis://localhost:6379
 # Miscellaneous
 DAYS_A_JOURNEY_CAN_BE_INACTIVE_FOR=30
 
+SUPPORT_EMAIL=email@example.gov.uk
+
 # Contentful
 CONTENTFUL_URL=cdn.contentful.com
 CONTENTFUL_SPACE=
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -15,7 +15,8 @@ en:
   banner:
     beta:
       tag: "beta"
-      message: "This is a new service - your feedback will help us to improve it."
+      message:
+          This is a new service – your <a href="mailto:%{support_email}" class="govuk-link">feedback</a> will help us to improve it.
     preview:
       tag: "preview"
       message: "This environment is only for previewing Contentful changes before publishing."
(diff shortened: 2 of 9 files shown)


config/locales/en.yml
```diff
@@ -15,7 +15,8 @@ en:
   banner:
     beta:
       tag: "beta"
-      message: "This is a new service - your feedback will help us to improve it."
+      message:
+          This is a new service – your <a href="mailto:schools.digital@education.gov.uk" class="govuk-link">feedback</a> will help us to improve it.
```
jacksonj04, line 19: I'd be tempted to make this email address a global variable somewhere, since we use it in a few places and it might change in future.

jacksonj04 (changes requested):

jacksonj04:
Other drive-by thought: this should probably have a changelog entry!

LornaHarwood:
> Other drive-by thought: this should probably have a changelog entry!

I had forgotton to do a final git push with the changelog entry 🤦 -  it is there now

config/locales/en.yml
LornaHarwood: Good point - do you have an idea of where it could sit?

.env.example
```diff
@@ -17,6 +17,8 @@ REDIS_URL=redis://localhost:6379
 # Miscellaneous
 DAYS_A_JOURNEY_CAN_BE_INACTIVE_FOR=30
 
+SUPPORT_EMAIL=email@example.gov.uk
```
jacksonj04, line 20: This will also need adding to the files in `/terraform/workspace-variables` so it's available for GPaaS deployments

jacksonj04 (approved):
One comment to make sure you include the ENV variables in the Terraform files (and then set them in the environments when this is merged), otherwise 👍🏼

PR #289, opened 2021-04-15
(573) Support specification templates that exceed 50k characters

tahb:
## Changes in this PR

The field coming back from Contentful for a Category called `specification_template` has a problem in that we have added more than 50,000 characters to it. The team still need to add more but we think it is unlikely it will ever need to be more than 100k.

Given Categories are generic and may be reused for Catering as well as future categories like Energy, we should keep our field naming approach generic too. For instance, "specification_template_food_section" would not make sense when creating a Energy section and vice versa.

There are 27 unique sections of the specification at the moment. This is far too many to make a single field per section as every new section would require a code change. Eg. `specification_template_section_28` etc.

The solution we've gone for here is to allow the content team to split their content across multiple fields as they see fit. We'll add `specification_template_part_2` and make it easy to add `specification_template_part_3` (or more) in the future _with only_ changes required to Contentful.

Our approach is to intercept the category _as soon as it comes back from Contentful_ and merge it there before it goes any further into the code base. This means we don't have to make more changes to things like our Liquid validator or how we create the Journey record.

### Before

Screenshot 2021-04-12 at 15 27 03

### After 
Screenshot 2021-04-15 at 17 26 45
Screenshot 2021-04-15 at 17 26 31

## Next steps

- [ ] Update Content Model in master environment with new fields

### DIFF
--- a/app/services/get_category.rb
+++ b/app/services/get_category.rb
@@ -1,3 +1,7 @@
+class Contentful::Entry
+  attr_accessor :combined_specification_template
+end
+
 class GetCategory
   class InvalidLiquidSyntax < StandardError; end
 
@@ -14,8 +18,12 @@ class GetCategory
       raise
     end
 
+    # INFO: Due to a 50k character limit within Contentful we check to see if
+    # we need to combine this value from multiple fields set on a Contentful Category.
+    category.combined_specification_template = combined_specification_templates(category: category)
+
     begin
-      validate_liquid(template: category.specification_template)
+      validate_liquid(template: category.combined_specification_template)
     rescue Liquid::SyntaxError => error
       send_rollbar_error(message: "A user couldn't start a journey because of an invalid Specification", entry_id: category_entry_id)
       raise InvalidLiquidSyntax.new(message: error.message)
@@ -38,5 +46,23 @@ class GetCategory
 
   def validate_liquid(template:)
     Liquid::Template.parse(template, error_mode: :strict)
+  end
+
+  def combined_specification_templates(category:)
+    specification_template_array = []
+
+    # Allow a new `specification_template_part_x` field to be added in Contentful
+    # without requiring an additional code change.
+    all_specification_fields = (category.public_methods - Object.methods)
+      .grep(/^specification_template(_part_[0-9]+)*(?<!=)$/)
+      .sort
+
+    all_specification_fields.each do |specification_field|
+      if category.respond_to?(specification_field)
+        specification_template_array << category.send(specification_field)
+      end
+    end
+
+    specification_template_array.compact.join("\n")
   end
 end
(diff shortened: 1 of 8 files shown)


app/services/get_category.rb
```diff
@@ -39,4 +47,22 @@ def send_rollbar_error(message:, entry_id:)
...
     Liquid::Template.parse(template, error_mode: :strict)
   end
+
+  def combined_specification_templates(category:)
+    specification_template_array = []
+
+    # Allow a new `specification_template_part_x` field to be added in Contentful
+    # without requiring an additional code change.
+    all_specification_fields = (category.public_methods - Object.methods)
+      .grep(/^specification_template.*(?<!=)$/)
```
tahb, line 57: `*(?<!=)` Makes sure we don't return `specification_template` and `specification_template=`.

tahb: After sitting on this for a few days.. Is supporting `specification_template_part_3` and beyond something we want keep this code around in order to support it?

jacksonj04: I think supporting an arbitrary number of fields is useful - I do think the regex could perhaps be a bit tighter, perhaps `^specification_template(_part_[0-9]+)*(?<!=)$` so that it requires `_part_` followed by a number?

app/services/get_category.rb
```diff
@@ -39,4 +47,22 @@ def send_rollbar_error(message:, entry_id:)
...
+      .sort
+
+    all_specification_fields.each do |specification_field|
+      if category.respond_to?(specification_field)
+        specification_template_array << category.send(specification_field)
```
jacksonj04, line 62: Can we always guarantee that the `specification_template`s come back in the right order? I _think_ the Contentful API does it in the order in the UI, but that feels like it might be an assumption that will bite us later.

jacksonj04 (approved):
Happy for this to go as it is since it does the job, just left a couple of tweaks which we could make.

app/services/get_category.rb
tahb: I was worried about the same. I [tried to sort the accessors alphabetically](https://github.com/DFE-Digital/buy-for-your-school/pull/289/files#diff-bcc05eaf1d158c408a782dbce6062c6616b512f330b3531d5b8a6e96a7804d50R58) so that we ask the `category` for the information in the correct order.

app/services/get_category.rb
tahb: Too much flexibility could become a problem. Tightening would allow us to present a simpler instruction: 'To add more add another field with `part_n`"

tahb:
https://dfedigital.atlassian.net/wiki/spaces/GHBFS/pages/2441445390/Managing+content#Configuring-very-long-specifications-(50k%2B)%5BhardBreak%5D

PR #291, opened 2021-04-19
release-007

tahb:
- Add `noindex,nofollow` meta tag to all pages, as per Gov.UK guidance
- fix API auth by switching mechanism from Basic to Token
- remove `Returning to this specification` URL from task list
- Add Tasks to the database when iterating through Sections from Contentful
- fix XSS vulnerability by sanitising all user answers
- Terraform a new CDN to support the custom host name

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog 1.0.0].
 
 ## [Unreleased]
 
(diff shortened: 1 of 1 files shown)


tahb:
CI is disabled now to prevent us pushing during the CDN install. When this has a thumbs up we can merge with admin priviledge.

jacksonj04 (approved):
LGTM!

🚀

PR #293, opened 2021-04-19
Add log stream service

Stretch96:
## Changes in this PR

* Creates a user-provided service, and binds it to the applications
* `<ENV>_TF_VAR_syslog_drain_url` GitHub secrets need to be created for
  this to launch


tahb (approved):
LGTM

PR #298, opened 2021-04-20
(573) Fix fixtures to align with Contentful for multiple specification fields

tahb:
## Changes in this PR

If you ask for a category over the Contentful API you get `specificationTemplate` rather than `specification_template`. Fortunantely this hasn't caused a problem so far as Contentful makes the method `specification_template` automatically available as a getter on `Contentful::Model`. 

- update all existing fixtures with `specification_template` set from snake case to camel case, making it more accurate which should reduce the chances that this causes a problem in future, but should cause no functional change
- update `specification_template_part_2` to `specificationTemplatePart2` in our fixtures
- fix the regex to account for how Contentful defines getting methods with numbers. Instead of the expected `specification_template_part_2` it actually makes `specification_template_part2` available. Update our calls and stubs to reflect this

### DIFF
--- a/spec/services/get_category_spec.rb
+++ b/spec/services/get_category_spec.rb
@@ -70,7 +70,7 @@ RSpec.describe GetCategory do
 
         # It should leave the original fields as they were
         expect(result.specification_template).to eql("<article id='specification'>\n  <section>\n   <p>Part 1</p>\n  </section>\n</article>")
-        expect(result.specification_template_part_2).to eql("<article id='specification'>\n  <section>\n    <p>Part 2</p>\n  </section>\n</article>")
+        expect(result.specification_template_part2).to eql("<article id='specification'>\n  <section>\n    <p>Part 2</p>\n  </section>\n</article>")
 
         # The result object can be asked for the combined specification template
         expect(result.combined_specification_template)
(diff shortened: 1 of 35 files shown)


spec/services/get_category_spec.rb
```diff
@@ -70,7 +70,7 @@
 
         # It should leave the original fields as they were
         expect(result.specification_template).to eql("<article id='specification'>\n  <section>\n   <p>Part 1</p>\n  </section>\n</article>")
-        expect(result.specification_template_part_2).to eql("<article id='specification'>\n  <section>\n    <p>Part 2</p>\n  </section>\n</article>")
+        expect(result.specification_template_part2).to eql("<article id='specification'>\n  <section>\n    <p>Part 2</p>\n  </section>\n</article>")
```
jacksonj04, line 73: Is this still relying on the magic getter the gem provides, and we could swap to `result.specificationTemplatePart2` for added precision (along with a matching regex swap)?

tahb: At this point `result` is an instance of `Contentful::Entry` which has the magic getter that is set for us by the gem. I think this is accurate given the [translation that is happening here](https://github.com/DFE-Digital/buy-for-your-school/pull/298/files#diff-1d6e47596d10e7df1de8aa46e9b07a428512265b4e69af24f3e1bbac59afee3eR145)?

jacksonj04: Yup. Brain fart from me!

jacksonj04 (approved):
🚀

PR #305, opened 2021-04-21
(590) Force SSL in production

tahb:
This does several things that make our app more secure:

1. Redirects any HTTP traffic to HTTPS
1. Enables [HSTS](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) headers which ensure user agents change HTTP into HTTPS before the request is made
1. Marks cookies as secure so they are not sent over the wire in the clear

https://github.com/rails/rails/blob/794252d70990162d14da17ee854ad353e74297ee/actionpack/lib/action_dispatch/middleware/ssl.rb


lozette (approved):
🚀

PR #306, opened 2021-04-21
Create Task pattern; group Steps into Tasks

lozette:
## Changes in this PR

Trello epic: https://trello.com/c/OyNb9jRN/306-implement-grouping-multiple-questions-into-one-task

The initial intention of this PR was to show the new Task model on the Journey page. However, to do this
we have had to break the existing direct association between Journey and Steps, and create a new hierarchy
of Journey -> Section -> Task -> Step. This necessitated a large refactor of the underlying code.

This PR:
- Creates a Task model; creates Tasks in the database from Contentful and associates them with Sections
- Associates Steps with Tasks
- Displays the Tasks on the Journey page, but *only* if a Task contains multiple steps. If it contains only one step, the step is shown instead to reduce clicks for the end user
- Adds a new Task show page
- Maintains the illusion of a direct association of Journey -> Step, so that the existing fixtures still work with the tests. Eventually the fixtures will need to be updated and the workarounds can be removed.


jacksonj04 (approved):
I think this is a new record for "enormous change which doesn't actually slice neatly"!

🚀 

*Still to do:*

- [ ] Add the appropriate content types to Staging
- [ ] Migrate existing steps into tasks

PR #307, opened 2021-04-21
(591) Prevent simultaneous logins

tahb:
## Changes in this PR

- Use Redis as a session store instead of the default Rails Cookie store. Moving this data to server side will give us an opportunity to delete other user sessions, which will sign them out.
- Introduce a concept called `RedisSessionLookup` to keep track of which DfE Sign in UID is connected with each redis session keys. This allows us to delete the right session data for the specific user who has signed in twice

To test this:

- fire up the server locally
- visit localhost:3000 in 2 new browsers, either in private mode or using different browsers
- sign both in on both using the same UID, let's say "22"
- refresh the page of the user you signed in as **first** and check you're now signed out and can see the message
- refresh the page of the user you signed into **second** and check you're still signed in

We can't configure a `Redis::Namespace` with the Redis `session_store` which introduces an unfortunante inconsistency. `RedisSessions` replicates this by not using `Redis::Namespace` but connects straight to Redis. This makes it difference to the config of our `RedisCache`.

## Screenshots of UI changes

Screenshot 2021-04-21 at 17 24 03
Screenshot 2021-04-21 at 17 24 07


lozette (approved):
Works great!

PR #315, opened 2021-04-26
RSpec now runs on all examples when run by CI

tahb:
## Changes in this PR

- CI has been running RSpec but RSpec hasn't had any files to test in 1 month https://github.com/DFE-Digital/buy-for-your-school/pull/245. [For example this is a feature 11 days ago with missing rspec ourput](https://github.com/DFE-Digital/buy-for-your-school/runs/2354797149?check_suite_focus=true#step:4:40) (action history doesn't go back that far)
- Fix this issue by copying spec files into the image (remember this used to be `COPY . .` so this change isn't conceptually new)
- Fix missed docker-compose redis configuration problem we now get when running rspec again

Please investigate the Test action for this pull request and verify that the output of RSpec is actually testing our ~359 examples!


jacksonj04 (approved):
Tasty tests ✅

PR #317, opened 2021-04-26
Don't notify on every failed build, only those during deployments

tahb:
The slack channel is going to get very noisy if we notify on test failures for pull request test runs. The dev should already be looking at the pull request to go green before moving it into review so an additional Slack message isn't necessary, I don't think.


jacksonj04 (approved):
Agree, CI telling us something which we're going to check anyway is pointless.

PR #318, opened 2021-04-26
Text input field width is now to set to `full` to fit the screen width

LornaHarwood:
## Changes in this PR

Fixes text input fields width and now displays to full width

### Before

Screen Shot 2021-04-27 at 17 14 07

### After

Screen Shot 2021-04-27 at 17 13 46

### DIFF
--- a/app/views/steps/long_text.html.erb
+++ b/app/views/steps/long_text.html.erb
@@ -2,7 +2,7 @@
   <%= f.govuk_text_area :response,
     label: { text: @step_presenter.title, size: "l" },
     hint: -> { @step_presenter.help_text_html },
-    width: "one-third",
+    width: "full",
     rows: 9
   %>
 <% end %>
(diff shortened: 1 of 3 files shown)


app/views/steps/long_text.html.erb
```diff
@@ -2,7 +2,7 @@
   <%= f.govuk_text_area :response,
     label: { text: @step_presenter.title, size: "l" },
     hint: -> { @step_presenter.help_text_html },
-    width: "one-third",
+    width: "full",
     rows: 9
```
tahb, line 6: The width looks good but the height seems to have an existing issue. I wonder why this isn't 9 rows high?

tahb: Oh I think I see what's confused me. This change is for long-text but screenshots from short-text have been used?

tahb (approved):
LGTM after initial screenshot confusion. Needs a line in our CHANGELOG, then this can be merged.

app/views/steps/long_text.html.erb
LornaHarwood: Yes, sorry, I should have been more explicit at what the screenshots are showing: this is just the short-text display

PR #324, opened 2021-04-27
Cache CI runs to reduce run times by ~5 minutes

tahb:
(Opening this while I wait for all my rebasing jobs to succeed to make this process easier)

This work is being copied across from our Rails Template. It was added after this project was created [1].

We are already using Docker Compose in CI on this project. However for context there is another pending change on the Rails Template[2] that includes a screenshot of how caching can behave.

The approach here is that we create a shared area on disk called /tmp/docker-save. We load this in at the start of every test run. It will make a past docker image for this app available to the build context. Docker can then use this image and it's layers to optimise the build steps, opting for the cached version rather than rebuilding from scratch.

If the cache is not hit (meaning the Dockerfile or gem.lock file changed) then it will add a new image into the cache at the end of the run. In this situation the builds will still take as long as they do now however this should be much less frequent.

[1] https://github.com/dxw/rails-template/pull/147
[2] https://github.com/dxw/rails-template/pull/213

## Before

Screenshot 2021-04-27 at 12 19 23

## After

Screenshot 2021-04-27 at 15 50 46

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -13,7 +13,22 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v2
+      - id: cache-docker
+        uses: actions/cache@v2
+        with:
+          path: /tmp/docker-save
+          key:
+            docker-save-${{ hashFiles('Dockerfile', 'Gemfile.lock',
+            'package-lock.json') }}
+      - name: Load cached Docker image
+        run: docker load -i /tmp/docker-save/snapshot.tar || true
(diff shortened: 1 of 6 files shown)


jacksonj04 (approved):
👏🏼

tahb:
Trying to work out why this isn't caching as it should do, and does do on the Rails Template. Trying to bring parity with the additional Dockerfile changes.

tahb:
I've changed the Dockerfile to use multi stage builds and caching works a charm. Adding screenshots for before and after.

PR #328, opened 2021-04-27
(610) Error pages are styled

tahb:
## Changes in this PR

The default Rails errors were still being served on live environments beacuse they were in public/500.html etc.

Make these error pages styled.

In future we could make static files that are self contained with styles to be served by a CDN if the application becomes unavailable.

### Before

Screenshot 2021-04-27 at 10 51 30

### After

Screenshot 2021-04-27 at 18 35 54
Screenshot 2021-04-27 at 18 35 59
Screenshot 2021-04-27 at 18 36 05


spec/requests/errors_spec.rb
```diff
@@ -0,0 +1,30 @@
+require "rails_helper"
+
+RSpec.describe "Errors", type: :request do
+  describe "internal_server_error" do
+    it "the health_check endpoint is not authenticated" do
```
jacksonj04, line 5: These test descriptions are incorrect

jacksonj04 (changes requested):
Needs changes to the test descriptions to reflect reality (copy/paste error)

jacksonj04 (approved):

PR #330, opened 2021-04-28
Inline scripts now use nonce's

LornaHarwood:
## Changes made

Expandable text fields were visible by default

When we switched on the content security policy, we didn't allow inline scripts. We now use the built in helper to generate a nonce for these so the scripts can be loaded.

## Before

Screen Shot 2021-04-29 at 17 41 20

Screen Shot 2021-04-29 at 17 41 14

## After

Screen Shot 2021-04-29 at 17 39 49

Screen Shot 2021-04-29 at 17 39 38

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -20,6 +20,8 @@ Rails.application.config.content_security_policy do |policy|
   # policy.report_uri "/csp-violation-report-endpoint"
 end
 
+Rails.application.config.content_security_policy_nonce_generator = ->(request) { SecureRandom.base64(16) }
+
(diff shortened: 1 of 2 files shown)


config/initializers/content_security_policy.rb
```diff
@@ -20,6 +20,8 @@
   # policy.report_uri "/csp-violation-report-endpoint"
 end
 
+Rails.application.config.content_security_policy_nonce_generator = ->(request) { SecureRandom.base64(16) }
```
jacksonj04, line 23: We should consider adding this (even if commented out) to our rails template, for the next weary travellers!

jacksonj04 (approved):

PR #332, opened 2021-04-29
Make breadcrumb journey match the user journey

LornaHarwood:
## Changes in this PR

The back button journey now matches the user journey

### DIFF
--- a/spec/features/school_buying_professionals/view_a_list_of_tasks_spec.rb
+++ b/spec/features/school_buying_professionals/view_a_list_of_tasks_spec.rb
@@ -23,9 +23,59 @@ feature "Users can view the task list" do
     end
   end
 
-  scenario "user can navigate back to the dashboard" do
+  scenario "user can navigate back to the task list from a task view" do
+    start_journey_with_tasks_from_category(category: "section-with-multiple-tasks.json")
+
+    within(".app-task-list") do
+      click_on "Task with multiple steps"
+    end
+
+    click_on(I18n.t("generic.button.back"))
+
+    expect(page).to have_content(I18n.t("specifying.start_page.page_title"))
+  end
+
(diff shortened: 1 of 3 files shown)


spec/features/school_buying_professionals/view_a_list_of_tasks_spec.rb
```diff
@@ -26,6 +26,7 @@
   scenario "user can navigate back to the dashboard" do
```
jacksonj04, line 26:
```suggestion
  scenario "user can navigate back to the dashboard from a question" do
```

jacksonj04: This could be more specific about what the test is actually doing.

For completeness we should actually have a series of tests that each validate one possible 'back' action:

- user can navigate back to the list of journeys from a task list
- user can navigate back to the task list from a task view
- user can navigate back to the task list from a question (where the task has 1 question)
- user can navigate back to the task view from a question (where the task has >1 questions)

spec/features/school_buying_professionals/view_existing_journeys_spec.rb
```diff
@@ -36,7 +36,6 @@
       start_journey_from_category(category: "radio-question.json")
 
       click_on(I18n.t("generic.button.back"))
-      click_on(I18n.t("dashboard.existing.link"))
```
jacksonj04, line 39: Good behaviour change, I hadn't even considered that this would be the expected flow.

jacksonj04 (changes requested):
I think we can change the test behaviour to give us more confidence that the 'back' actions are working correctly under more situations, rather than validating that a whole chain of actions works at once.

jacksonj04 (approved):

PR #336, opened 2021-05-04
Add script/update and script/bootstrap

jacksonj04:
To closer follow the Scripts To Rule Them All pattern, these scripts will automate updating dependencies and performing migrations. They are also run automatically as part of script/test.

## Changes in this PR

`script/update` and by extension `script/bootstrap` are run automatically as part of `script/test`.

Since these actions are idempotent they do not interfere with the setup actions already taken by the `Dockerfile` during CI.

### DIFF
--- a/script/bootstrap
+++ b/script/bootstrap
@@ -1,0 +1,64 @@
+#!/bin/sh
+
+# script/bootstrap: Resolve all dependencies that the application requires to
+#                   run.
+
+set -e
+
+cd "$(dirname "$0")/.."
+
+if [ -f .gitmodules ]; then
+  echo "==> Updating Git submodules..."
+  git submodule update --init
+fi
+
+if [ -z "$CI" ]; then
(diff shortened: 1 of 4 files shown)


script/bootstrap
```diff
@@ -0,0 +1,64 @@
...
+if [ -f .gitmodules ]; then
+  echo "==> Updating Git submodules..."
+  git submodule update --init
+fi
+
+if [ -z "$CI" ]; then
+  if [ -f Brewfile ] && [ "$(uname -s)" = "Darwin" ]; then
+    if ! brew bundle check >/dev/null 2>&1; then
+      echo "==> Installing Homebrew dependencies..."
+      brew bundle install --verbose --no-lock
```
tahb, line 19: I think we should add our [own Brewfile](https://github.com/dxw/rails-template/blob/develop/Brewfile) to support this by adding the various bits and bobs.

tahb (approved):
Probably needs a Brewfile, but otherwise LGTM.

PR #342, opened 2021-05-10
Add backend functionality to record events

jacksonj04:
## Changes in this PR

Implement a new model for storing event log entries.

This is a timestamped record, along with:

- Internal user ID
- Internal journey ID
- Contentful category ID
- Contentful task ID
- Contentful section ID
- Contentful question ID
- String to describe the action being taken
- JSONB for additional payload needed to describe context

In all cases we are storing strings and not relationships, since journeys in particular may be purged from the database after inactivity, but we still may want to derive information.

### DIFF
--- a/app/services/record_action.rb
+++ b/app/services/record_action.rb
@@ -1,0 +1,72 @@
+class RecordAction
+  class UnexpectedActionType < StandardError; end
+
+  attr_accessor :action_type, :journey_id, :user_id, :contentful_category_id, :contentful_section_id, :contentful_task_id, :contentful_step_id, :data
+
+  ALLOWED_ACTION_TYPES = %w[
+    begin_journey
+  ].freeze
+
+  def initialize(
+    action:,
+    journey_id:,
+    user_id:,
+    contentful_category_id: nil,
+    contentful_section_id: nil,
+    contentful_task_id: nil,
+    contentful_step_id: nil,
+    data: nil
+  )
+    self.action_type = action
+    self.journey_id = journey_id
+    self.user_id = user_id
+    self.contentful_category_id = contentful_category_id
+    self.contentful_section_id = contentful_section_id
+    self.contentful_task_id = contentful_task_id
+    self.contentful_step_id = contentful_step_id
+    self.data = data
+  end
+
+  def call
+    if unexpected_action_type?
+      send_rollbar_warning
+      raise UnexpectedActionType
+    end
+
+    ActivityLogItem.create(
+      action: action_type,
+      journey_id: journey_id,
+      user_id: user_id,
+      contentful_category_id: contentful_category_id,
+      contentful_section_id: contentful_section_id,
+      contentful_task_id: contentful_task_id,
+      contentful_step_id: contentful_step_id,
+      data: data
+    )
+  end
+
+  private
+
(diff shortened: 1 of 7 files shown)


tahb (commented):
A couple of suggestions but this looks like a decent foundation. 

The approach is coupled to the presence of a journey and Contentful. I think this is ok as we cannot see the service performing any other type of transaction. If we can foresee this we may want to consider a more generic approach with polymorphic target records etc which could be more easily reused.

I imagine we'd have an asynchronous job to tackle the creation of these records? Throughout the code we could then have generic method that was something like `track_user_action(user, action, contentful_params)`, is that the idea? If that is the idea, unless we want to inject user records into all our services this will be a controller only method. This would be fine so long as we steer clear of using the same mechanism for machine events (see other comment).

Some suggestions:

1. Name of the table
1. Indexes for how we expect to pull this data out
1. We may wish to also store a session identifier so we can determine if a user performed those actions within the same or different sessions. That could be an interesting way to slice the data
1. Do we want database validations for presence to provide better integrity
1. An ADR to explain our approach to event tracking, avoiding GA, approach to user vs machine events

tahb:
@jacksonj04 I can see that my points 1, 2, 5 have caused some changes (and lots of indexes!) 🙌 

What are your thoughts on 3, 4 and the potential for adding a generic job for adding this to the database? I wonder if the plan is to we wait until we track the first event to add that in which is one way to do it but I understood we'd be adding this type of plumbing 🚰 in this change so thought I'd check.

app/services/record_action.rb
```diff
@@ -0,0 +1,72 @@
+class RecordAction
+  class UnexpectedActionType < StandardError; end
+
+  attr_accessor :action_type, :journey_id, :user_id, :contentful_category_id, :contentful_section_id, :contentful_task_id, :contentful_step_id, :data
+
+  ALLOWED_ACTION_TYPES = %w[
+    begin_journey
```
tahb, line 7: This will be a joy to extend by adding new types 🎉

app/services/record_action.rb
```diff
@@ -0,0 +1,72 @@
...
+  ].freeze
+
+  def initialize(
+    action_type,
```
tahb, line 11: You can still use a named parameter here and leave the value blank like `action_type:,`. 

Doing this keeps the advantage of giving us more meaningful errors when we don't pass things in correctly. It'll say "you forgot to pass an 'action_type'" rather than it blowing up as it's `nil` further down the stack when it's used for the first time. It'd have a more obscure error right now if this was called with nil:

```
ActivityLogItem.create(action: nil) => 💥
```

app/services/record_action.rb
```diff
@@ -0,0 +1,72 @@
...
+    journey_id: nil,
+    user_id: nil,
```
tahb, line 13: I can see how nil defaults makes testing easier but I wonder if there will ever be a case where we care about an activity that doesn't have a user or journey? Could this be premature safeguarding that may open to door for (and suggest to the user to use) this class to track non-user activities.

tahb (approved):
Looks solid to me. A couple of small code design suggestions for you that I think are worth considering before merge.

PR #346, opened 2021-05-10
(631) Sections are shown in the order set by Contentful

tahb:
## Changes in this PR

The order of sections that are presented to the user does not always match the order specified by the content team in Contentful.

[We had already fixed this](https://github.com/DFE-Digital/buy-for-your-school/pull/156/files) but [it seems to have been undone doing the work to introduce `task` into the associations](https://github.com/DFE-Digital/buy-for-your-school/pull/326/commits/d4dc93b3a42ae421c5d6536c39cfa1d5f90ed1de).

- we set a `default_scope` on the section as `implicit_order_column` does not do what we thought it did
- we add a new `order` column to section, along with indexes, to store the order it comes back from Contentful
- we present sections using the `order` value rather than the `created_at` value


jacksonj04 (approved):
🔢 

LGTM

PR #350, opened 2021-05-11
(632) Order steps as defined in Contentful

tahb:
The original user story was to fix steps for hidden questions. After digging into the problem I believe the fact it was hidden was a coincidence and the same random ordering bug exists, [as seen on sections](https://github.com/DFE-Digital/buy-for-your-school/pull/346).

## Changes in this PR

- reorganise the test suite in a few places to make these changes easier to add
- all Steps store the order they were received back from Contentful in
- all Steps are shown to users in the right order by setting a default scope


spec/features/school_buying_professionals/view_a_list_of_tasks_spec.rb
```diff
@@ -163,6 +163,40 @@
         expect(page).to_not have_content("Task with a hidden step")
       end
     end
+
+    context "when that step becomes visible" do
+      it "appears directly underneath the question that made it visible" do
```
jacksonj04, line 168: I don't think this is the best description; we're making sure the step appears in the order it was given in Contentful, but this isn't necessarily immediately underneath the question.

jacksonj04 (approved):
One tweak to a test description which I don't think quite matches the intent.

Other than that, the purist in me wonders if this could be split into a test tidying PR and a bug fixing PR, but that's not a strongly held feeling.

PR #353, opened 2021-05-12
Fix redis connections on randomly failing on CI

tahb:
## Changes in this PR

Failures - particularly on CI - have been randomly occurring recently. Sometimes the test suite runs before Redis is ready and a large part of the test suite fails due to failing Redis connections.

The Debian `wait-for-it` package is used to wait for Redis to be ready _before_ running the tests. This follows this Docker guidance [1]. We give Redis a generous 60 seconds to start up.

Wait for it is only installed in the test stage of the Dockerfile so it should not be included in the deployed image.

[1] https://docs.docker.com/compose/startup-order/

Screenshot 2021-05-12 at 12 28 55

### DIFF
--- a/script/test
+++ b/script/test
@@ -20,12 +20,12 @@ if [ -n "$TEST_FILE" ]; then
   echo "==> Running the tests matching '$TEST_FILE'..."
   bundle exec rspec --pattern "$TEST_FILE"
 else
-  if [ -n "$CI" ]; then
-    echo "==> Linting Ruby..."
-    bundle exec standardrb
-  else
+  if [ -n "$AUTOMATICALLY_FIX_LINTING" ]; then
     echo "==> Linting Ruby in fix mode..."
     bundle exec standardrb --fix
+  else
+    echo "==> Linting Ruby..."
+    bundle exec standardrb
   fi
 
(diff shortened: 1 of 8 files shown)


script/test
```diff
@@ -20,7 +20,7 @@ if [ -n "$TEST_FILE" ]; then
   echo "==> Running the tests matching '$TEST_FILE'..."
   bundle exec rspec --pattern "$TEST_FILE"
 else
-  if [ -n "$CI" ]; then
+  if [ -n "$AUTOMATICALLY_FIX_LINTING" ]; then
```
jacksonj04, line 23: Is this actually fixing? I think it's linting, but not doing a fix, so maybe `LINT_CODEBASE` is a better name?

(also, what's the significance of this variable? Shouldn't we be linting regardless?)

jacksonj04 (approved):
Approved, just one query about the nature of `AUTOMATICALLY_FIX_LINTING`.

script/test
tahb: I think I got the -n flag the wrong way round so I have switched the code around.

Now if `"$AUTOMATICALLY_FIX_LINTING" is present then we run standard with `--fix`

PR #357, opened 2021-05-14
(654) Use GitHub environment rather than repository secrets

tahb:
## Changes in this PR

This avoids the [limit of 100](https://docs.github.com/en/actions/reference/encrypted-secrets#limits-for-secrets) `repository` secrets that we just ran into.

By specifying the `environment` within the Github Action job, only the variables within that [Github environment](https://docs.github.com/en/actions/reference/environments#environment-secrets) will be made available.

Given that the environment can only be declared at the top of a `job`, we move  our 4 deployment proceses from `job steps` into individual `jobs`.

I've tested this on the reserch environment. I've removed all RESEARCH keys from the repository secret list and added them all into the environment called `research`. You can take a [peek inside this deployment](https://github.com/DFE-Digital/buy-for-your-school/runs/2582959652?check_suite_focus=true) that the deploy job still gets all the environment variables as previously configured.

The same `APP_ENV_RESEARCH_` prefix is still used even though this change introduces isolation of secrets so that each job should only get ENV for that environment. It is possible that these prefixes could be removed in future but we would need to do more investigation into the [Terraform deployment process that does complex key parsing](https://github.com/DFE-Digital/buy-for-your-school/blob/develop/script/deploy-terraform#L28) in order to pass the variables to the newly deployed containers.

I'll leave the rest of the environment variables on staging, production and preview unchanged until this proposal is reviewed. Once we're happy I'll change the rest of the variables from repoistory variables before we merge.

### Before

Screenshot 2021-05-12 at 14 57 48

### After

Screenshot 2021-05-14 at 11 14 18

Screenshot 2021-05-14 at 11 03 24

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,7 +10,7 @@ on:
 env:
   GHCR_USERNAME: ${{ secrets.GHCR_USERNAME }}
   GHCR_PASSWORD: ${{ secrets.GHCR_PASSWORD }}
-  GHCR_REPO: "ghcr.io/dfe-digital/buy-for-your-school"
+  GHCR_REPO: 'ghcr.io/dfe-digital/buy-for-your-school'
   CF_USER: ${{ secrets.CF_USER }}
   CF_PASSWORD: ${{ secrets.CF_PASSWORD }}
   AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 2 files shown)


jacksonj04 (approved):
I'm not a massive fan of the duplication, but I can't think of a more elegant solution to this.

tahb:
Preview, staging and prod have had their secrets moved into environment secrets. Testing staging deploys OK still and then the next step will be to deploy to `main` before we're done 🎉

PR #361, opened 2021-05-18
release-013

tahb:
- prevent random CI failures due to missing Redis connections
- GitHub action that allows manual cloning from master to staging with one-click for content testing
- steps are listed in the order set in Contentful rather than database insert order
- fix github secret limit by switching to environment variables


jacksonj04 (approved):
🚀

PR #362, opened 2021-05-19
(652) Replace GetStepsFromSection service that was still used

tahb:
## Changes in this PR

We recently did a bunch of work to introduce Tasks into the data hierarchy so that steps were expected to come from Tasks, rather than Sections

Contentful still has section data that references all the steps. This part of the migration wasn't done so things still somehow work in live environments but the data needs cleaning up to reduce confusion when changes need to be made.

This work tackles:

- chore: remove duplicate caching effort
- chore: iterate the cache warming approach so it reliably warms everything, every night. Rather than only adding missing entries 
- switches Cache warming logic to get Step data via Tasks rather than Sections
- switches Journey map logic to get Step data via Tasks rather than Sections, this fixes this feature too
- remove now unused GetStepsFromSection service


app/jobs/warm_entry_cache_job.rb
```diff
@@ -32,4 +36,32 @@ def cache
       ttl: ENV.fetch("CONTENTFUL_ENTRY_CACHING_TTL", 60 * 60 * 72)
     )
   end
+
+  def backup_old_cache
+    move_cached_items(
+      keys: RedisCache.redis.keys("contentful:entry:*"),
+      old_key_prefix: "contentful:entry:",
```
jacksonj04, line 45: Since these prefixes are being used in a few places is it worth defining them as constants somewhere?

app/services/get_steps_from_task.rb
```diff
@@ -9,11 +11,28 @@ def call
     return [] unless task.respond_to?(:steps)
     step_ids = []
```
jacksonj04, line 12: I wonder if this check should actually apply to the entire category, rather than just a task?

jacksonj04 (approved):
📦 🚀

app/services/get_steps_from_task.rb
jacksonj04: Actually scrap that (certainly for this PR), because then we've got the faff of maintaining the list of tasks for the entire category ingest process rather than it just being scoped to this service.

app/jobs/warm_entry_cache_job.rb
tahb: Good idea! Will see what I can do.

tahb: One constant added!

PR #363, opened 2021-05-19
(639) Answering a task step takes user to the next step

tahb:
Before this change the user would click on one of their tasks and it would present a list of questions. The user would click into a question, answer it and be taken back to the task page.

This is quite a fair bit of back and forth for a user who wants to progress through a related task in the same session. This change will change the redirection in this scenario so that they now only return to the task when we think they're done.

There are more stories to come after this that will change this task page to a check your answers pattern.

## Changes in this PR

- If the task only has 1 step then answering it returns the user to the task page, as before.
- If the step is the last question to be answered then it returns the user to the task page, as before
- Otherwise if there is a remaining unanswered question the user will be taken straight to it now, instead of the task page.

### 1️⃣We start

Screenshot 2021-05-19 at 16 18 26

### 2️⃣ We select the middle question

Screenshot 2021-05-19 at 16 18 31

### 3️⃣ We answer it and are taken to the first question

Screenshot 2021-05-19 at 16 18 36

### 4️⃣ We answer it and are taken to the last question

Screenshot 2021-05-19 at 16 18 44

### 5️⃣ We answer it and are taken to back to the task page having completed the task
Screenshot 2021-05-19 at 16 18 49

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -46,6 +46,10 @@ class Task < ApplicationRecord
 
     NOT_STARTED
   end
+
+  def all_steps_answered?
+    eager_loaded_steps.all?(&:answered?)
+  end
   private
 
   def eager_loaded_steps
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -46,4 +46,17 @@ class Task < ApplicationRecord
 
     NOT_STARTED
   end
+  private
+
+  def eager_loaded_steps
+    @eager_loaded_steps ||= steps.includes([
+      :short_text_answer,
+      :long_text_answer,
+      :radio_answer,
+      :checkbox_answers,
+      :currency_answer,
+      :number_answer,
+      :single_date_answer
+    ])
+  end
 end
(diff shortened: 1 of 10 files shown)


app/models/task.rb
```diff
@@ -46,6 +46,10 @@ def status
 
     NOT_STARTED
   end
+
+  def all_steps_answered?
+    eager_loaded_steps.all? { |step| step.answered? }
```
jacksonj04, line 51: TIL about `.all?`. Would be handy to replace the logic in `status` with a call to this method as well, so we only have this logic in one place.

jacksonj04 (approved):
LGTM!

My comment on `.all?` is really something that should be a separate refactor, so I've spun up a ticket for that.

app/models/task.rb
tahb: Having looked at that method again I wonder if this new addition actually needs to filter out hidden steps too! In which case I can and should probably refactor it for reuse.

tahb: I've made this change 👍

PR #365, opened 2021-05-20
Swap task view pattern for the check your answers pattern

LornaHarwood:


spec/features/school_buying_professionals/view_a_list_of_tasks_spec.rb
```diff
@@ -74,7 +74,10 @@
         click_on "Task with multiple steps"
       end
 
-      click_on "Everyday services that are required and need to be considered"
+      within(".govuk-summary-list") do
+        page.find_by_id("checkbox").click
```
jacksonj04, line 78: This `id` feels odd - `id` should normally be something that uniquely identifies an element, and `checkbox` seems pretty generic.

Should this actually be clicking a "change" link?

spec/features/school_buying_professionals/view_a_task_spec.rb
```diff
@@ -57,27 +57,49 @@
...
+    it "allows the user to click on a step to supply an answer, and returns to the change your answers page" do
       start_journey_with_tasks_from_category(category: "section-with-multiple-tasks.json")
 
       within(".app-task-list") do
         click_on "Task with multiple steps"
       end
 
-      click_on "Which service do you need?"
+      within(".govuk-summary-list") do
+        page.find_by_id("radio").click
```
jacksonj04, line 74: I think this is another case where `id` isn't the right solution (or if it is we're doing something strange with `id`s!)

jacksonj04 (changes requested):
Just a couple of things:

- Selecting elements by `id` for testing feels a bit weird for the IDs we're actually selecting by - there's probably a cleaner way to do this
- If we could improve the checkbox presenter so we don't get `[ Array, Of, Items ]` that would be great
- This needs a rebase to make sure it works with Tom's flow changes (although I think it should just drop straight in, so this might be a bit boilerplateish)

jacksonj04 (approved):

PR #367, opened 2021-05-20
Refactor CreateJourney to accept and store category ID

jacksonj04:
## Changes in this PR

- Change behaviour of `CreateJourney` to accept a category ID and derive the name from that, rather than accept a name and rely on the category ID being defined in the class.
- Store the Contentful category ID with newly created journeys, in case we need it later (eg for activity logging)

### DIFF
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -56,12 +56,12 @@ class Task < ApplicationRecord
   def eager_loaded_visible_steps
     @eager_loaded_visible_steps ||= visible_steps.includes(
       %i[short_text_answer
-         long_text_answer
-         radio_answer
-         checkbox_answers
-         currency_answer
-         number_answer
-         single_date_answer]
+        long_text_answer
+        radio_answer
+        checkbox_answers
+        currency_answer
+        number_answer
+        single_date_answer]
     ).ordered
   end
 end
--- a/db/migrate/20210520155735_add_category_id_to_journey.rb
+++ b/db/migrate/20210520155735_add_category_id_to_journey.rb
@@ -1,0 +1,5 @@
+class AddCategoryIdToJourney < ActiveRecord::Migration[6.1]
+  def change
+    add_column :journeys, :contentful_id, :string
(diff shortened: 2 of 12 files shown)


app/models/task.rb
```diff
@@ -63,12 +63,12 @@ def next_unanswered_step_id
   def eager_loaded_visible_steps
     @eager_loaded_visible_steps ||= visible_steps.includes(
       %i[short_text_answer
-         long_text_answer
-         radio_answer
-         checkbox_answers
-         currency_answer
-         number_answer
-         single_date_answer]
+        long_text_answer
```
tahb, line 66: I did this recently but CI was happy. Sorry! 👀

spec/services/create_journey_spec.rb
```diff
@@ -17,6 +17,7 @@
       expect { described_class.new(category_id: "contentful-category-entry", user: build(:user)).call }
         .to change { Journey.count }.by(1)
       expect(Journey.last.category).to eql("Catering")
+      expect(Journey.last.contentful_id).to eql("contentful-category-entry")
```
tahb, line 20: Now we're calling Journey.last twice we could extract that into a variable to remove one call to the database.

db/migrate/20210520155735_add_category_id_to_journey.rb
```diff
@@ -0,0 +1,5 @@
+class AddCategoryIdToJourney < ActiveRecord::Migration[6.1]
+  def change
+    add_column :journeys, :contentful_id, :string
```
tahb, line 3: We might want to consider adding presence validation for this column incase we ever try to create one without this important piece of data. 

We had/have `t.string "category", null: false` but it could be filled with an unrelated string by mistake and the `contentful_id` could be null. That would cause many weird things to happen I'm sure!

tahb (approved):
LGTM. Couple of small things. I think we should add validation in for the new column to improve data integrity.

PR #379, opened 2021-06-02
Feature/674 broaden feature specs for formatted answers

peterdavidhamilton:
## Changes in this PR

Add feature spec coverage for all responses in the task list.
This checks the answer presenters formatted output and view partials.

PR #380, opened 2021-06-02
Add missing number answer partial

peterdavidhamilton:
## Changes in this PR

Number responses now render and do not cause a 500 error

### Before

500 error

### After
Screenshot 2021-06-02 at 16 54 59

## Next steps

Spec coverage will be handled with ticket #674

### DIFF
--- a/app/views/tasks/_long_text_answer.html.erb
+++ b/app/views/tasks/_long_text_answer.html.erb
@@ -7,4 +7,4 @@
     <dd class="govuk-summary-list__value"><%= "-" %></dd>
     <dd class="govuk-summary-list__actions"><%= link_to I18n.t("generic.button.change_answer"), journey_step_path(@journey, step), class: "govuk-link", id: "long-text" %></dd>
   <% end %>
-</div>
+  </div>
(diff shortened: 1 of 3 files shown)


app/views/tasks/_long_text_answer.html.erb
```diff
@@ -7,4 +7,4 @@
     <dd class="govuk-summary-list__value"><%= "-" %></dd>
     <dd class="govuk-summary-list__actions"><%= link_to I18n.t("generic.button.change_answer"), journey_step_path(@journey, step), class: "govuk-link", id: "long-text" %></dd>
   <% end %>
-  </div>
+</div>
```
jacksonj04, line 10: Could we drop these tweaks in unrelated files, or bump them to a separate commit?

jacksonj04 (approved):

app/views/tasks/_long_text_answer.html.erb
peterdavidhamilton: Sure, I'll address the tidy up in #379

PR #384, opened 2021-06-04
Guard against use of back button after form validation errors

peterdavidhamilton:
## Changes in this PR

Should a user attempt to submit an answer and receive form validation errors, then use the back button, a routing error would occur. An additional route catches this and resets the form.

### Before

Clicking back here would raise an error.

Screenshot 2021-06-04 at 14 27 49

### After

Now the validations are cleared.

Screenshot 2021-06-04 at 14 34 27


jacksonj04 (approved):
Good catch!

PR #387, opened 2021-06-07
Implement dynamic validation defined in Contentful

peterdavidhamilton:
## Changes in this PR

Use an optional JSON field to define custom validations for questions.
<https://dfedigital.atlassian.net/wiki/spaces/GHBFS/pages/3261825072/Error+Messages>

- define a custom error message rendered to the user
- use an upper/lower bound to set a range the value must fall within to be considered valid

## Next steps

- cater for open-ended bounds (no upper limit for example)

### DIFF
--- a/app/services/create_step.rb
+++ b/app/services/create_step.rb
@@ -8,6 +8,7 @@ class CreateStep
 
   # Steps are either Questions or Statements
   #
+  # @return [Array<String>]
   ALLOWED_CONTENTFUL_MODELS = %w[
     question
     statement
@@ -15,6 +16,7 @@ class CreateStep
 
   # Contentful Question Model defined types
   #
+  # @return [Array<String>]
   ALLOWED_CONTENTFUL_QUESTION_TYPES = %w[
     long_text
     short_text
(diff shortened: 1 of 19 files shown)


peterdavidhamilton:
Need to talk to the designers about validations and get examples but I imagine number and currency types could be dealt with in the same way with an upper/lower bound.

spec/services/create_step_spec.rb
```diff
@@ -298,5 +298,38 @@
           .to raise_error(CreateStep::UnexpectedContentfulStepType)
       end
     end
+
+    # TODO: use this pattern above if not already to curtial duplication
```
peterdavidhamilton, line 302: @EdwinKruglov this is my take on it

AndrewCowle (approved):

app/services/create_step.rb
```diff
@@ -134,6 +135,16 @@ def options
     contentful_entry.extended_options
   end
 
+  # Custom question validation
+  #
+  # @example
+  #   criteria => { lower: "", upper: "", message: "" }
```
peterdavidhamilton, line 141: I need to correct my `@example` because I used symbols bitmoji

EdwinKruglov (approved):

PR #388, opened 2021-06-08
Refactor presenters and view partials

peterdavidhamilton:
## Changes in this PR

- instantiate a presenter in the controller not the views
- nest presenter objects
- consolidate creation of formatted answers into a single method
- move view logic into presenters
- dry up view partials
- add missing unit tests for presenters


jacksonj04 (approved):
LGTM! Just give these two a squish so we keep passing commits all the way, then 🚀

PR #390, opened 2021-06-08
Add user action tracking

jacksonj04:
Add recording of key user actions throughout the journey:

- Starting a journey
- Revisiting a previously started journey
- Viewing a task for the first time
- Returning to a task which is either in progress or completed
- Looking at a step for the first time
- Returning to look at a previously completed step
- Answering a question for the first time
- Updating a previously answered question


peterdavidhamilton (approved):
image

PR #395, opened 2021-06-16
Store a hash of step counts on the task

peterdavidhamilton:
## Changes in this PR

- record numbers of total, visible, hidden and answered steps
- save counters as an extensible hash in JSONB
- use callbacks when answers are committed or steps revealed
- use the persisted tally to calculate task state in other methods
- remove memoisation in Task#eager_loaded_visible_steps which caused issues
- remove dedicated counting methods
- enable the TaskController to find the next_unanswered_task
- add a guard clause to Task#next_unanswered_step_id
- reduce the use of expensive relation loading

## Next steps

- add utility predicate methods to the task to query state
- use answer model class inheritance

### DIFF
--- a/app/models/short_text_answer.rb
+++ b/app/models/short_text_answer.rb
@@ -3,4 +3,10 @@ class ShortTextAnswer < ActiveRecord::Base
   belongs_to :step
 
   validates :response, presence: true
+
+  after_commit :update_task_counters
+
+  private def update_task_counters
+    step.update_task_counters
+  end
 end
--- a/app/controllers/tasks_controller.rb
+++ b/app/controllers/tasks_controller.rb
@@ -4,9 +4,9 @@ class TasksController < ApplicationController
   def show
     @journey = current_journey
     @current_task = task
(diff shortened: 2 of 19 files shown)


peterdavidhamilton:
It shouldn't be fragile but if it turns out it is we should revisit:

```ruby
def next_unanswered_task
  current_journey.tasks.reject(&:all_steps_answered?).last
end
```

Which effects `rspec ./spec/features/school_buying_professionals/view_a_list_of_tasks_spec.rb:200`
The method uses `.last` to keep that spec passing.

app/models/short_text_answer.rb
```diff
@@ -3,4 +3,10 @@ class ShortTextAnswer < ActiveRecord::Base
   belongs_to :step
 
   validates :response, presence: true
+
+  after_commit :update_task_counters
```
AndrewCowle, line 7: I would refactor this method into a concern for DRy-ness

EdwinKruglov (approved):

AndrewCowle (approved):

PR #56, opened 2020-11-30
(117) Refactor technical naming for plans to journeys and questions to steps

tahb:
## Changes in this PR

- Refactor routing and modelling of "Plan" to "Journey" to abstract the concept in order to support plans *and* specifications
- Refactor routing and modelling of "Question" to "Step" to abstract the concept in order to support questions, static content *and* decisions.
- Fixtures for the contentful type `question` remains and so the fixtures have kept their names 
- `Journey#new` still connects to the concept of a "plan" through an environment variable. I imagine that when we want to add a different type of journey, the button a user clicks can send across some context for which journey they are attempting to start. The controller can then use that to use different `next_entry_id`.
- I assume that a `Step` can eventually be a Contentful type of `static_content`, which means our `Step` model would handle questions and content which require no associated `Answer` record.

PR #397, opened 2021-06-17
#56: Make the create spec button comply with accessibility needs

EdwinKruglov:
## Changes in this PR

- Move the **Create new specification** button to the top-left of the dashboard in order to comply with accessibility requirements by removing the containing `div` and changing the dashboard layout class to `govuk-grid-column-full`.
- Update the **Create new specification** button to be a link.
- Update the `view_a_dashboard_spec` scenario `user can view existing specifications` by checking for the presence of the button and ensuring its containing class is`govuk-grid-column-full`.
- Update other relevant tests now that the button is a link.

### Before

image

### After

image

### DIFF
--- a/app/views/dashboard/show.html.erb
+++ b/app/views/dashboard/show.html.erb
@@ -8,9 +8,10 @@
   </div>
 
   <% if @current_user.journeys.any? %>
-    <div class="govuk-grid-column-two-thirds">
-        <h2 class="govuk-heading-m"><%= I18n.t("dashboard.existing.header") %></h2>
-        <p class="govuk-body"><%= I18n.t("dashboard.existing.body") %></p>
+    <div class="govuk-grid-column-full">
+      <%= link_to I18n.t("dashboard.create.button"), new_journey_path, class: "govuk-button" %>
+      <h2 class="govuk-heading-m"><%= I18n.t("dashboard.existing.header") %></h2>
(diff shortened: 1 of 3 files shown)


AndrewCowle (approved):

app/views/dashboard/show.html.erb
```diff
@@ -8,6 +8,9 @@
   </div>
 
   <% if @current_user.journeys.any? %>
+    <div class="govuk-grid-column-full">
+      <%= button_to I18n.t("dashboard.create.button"), new_journey_path, class: "govuk-button", method: :get %>
+    </div>
     <div class="govuk-grid-column-two-thirds">
```
peterdavidhamilton, line 14: Perhaps this should be removed so everything is within column-full?

app/views/dashboard/show.html.erb
```diff
@@ -8,6 +8,9 @@
   </div>
 
   <% if @current_user.journeys.any? %>
+    <div class="govuk-grid-column-full">
+      <%= button_to I18n.t("dashboard.create.button"), new_journey_path, class: "govuk-button", method: :get %>
```
peterdavidhamilton, line 12: Can we change this to link_to and remove **_method: get_**, I realise now the styling is the same.

peterdavidhamilton:
Looks good, if you can address those comments, squash it ;)

peterdavidhamilton (approved):

PR #398, opened 2021-06-21
Add ordering to tasks

peterdavidhamilton:
## Changes in this PR

Add task ordering which is required to reliably find the next unanswered task.


peterdavidhamilton:
The failed spec is that last-minute link change to the prototype which needs to be resolved separately.

EdwinKruglov (approved):

AndrewCowle (approved):

PR #400, opened 2021-06-22
Refactored update_task_counters into concern

AndrewCowle:
Refactored update_task_counters callback from the answer models into a concern to 'DRY' them up and updated specs to used a shared concern.


peterdavidhamilton (approved):
image

EdwinKruglov (approved):

PR #401, opened 2021-06-22
#87: YARD dependency and initial documentation

EdwinKruglov:
## Changes in this PR

- Added `yard` and `yard-junk` dependencies.
- Added new Rake task to generate the documentation .
- Documented the models, controllers and services to evaluate the documentation.
- Updated `.gitignore` to ignore the generated documentation.

## Next steps

Document the rest of the codebase.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -55,6 +55,7 @@ group :development do
   gem "rails-erd"
   gem "rails_layout"
   gem "spring-commands-rspec"
+  gem "yard-junk"
 end
 
 group :development, :test do
--- a/app/models/activity_log_item.rb
+++ b/app/models/activity_log_item.rb
@@ -1,3 +1,8 @@
+# ActivityLogItem captures user actions with contextual information, such as `user_id`, `journey_id`, etc.
+#
+# This is used to keep track of actions taken on {Journey}s, {Task}s, {Step}s, and others.
+#
+# @see RecordAction
 class ActivityLogItem < ApplicationRecord
   self.table_name = "activity_log"
 end
--- a/app/models/checkbox_answers.rb
+++ b/app/models/checkbox_answers.rb
@@ -1,3 +1,4 @@
+# CheckboxAnswers is used to capture a checkbox value answer to a {Step}.
 class CheckboxAnswers < ApplicationRecord
   include TaskCounters
 
@@ -7,6 +8,11 @@ class CheckboxAnswers < ApplicationRecord
             presence: true,
             unless: proc { |answer| answer.step.skippable? && answer.skipped }
 
+  # Overridden response accessor that ensures no blank checkbox values are set.
+  #
+  # @param [Array] args
+  #
+  # @return [Array, nil] array of answers or nil.
   def response=(args)
     return if args.blank?
 
--- a/app/models/currency_answer.rb
+++ b/app/models/currency_answer.rb
@@ -1,3 +1,4 @@
+# CurrencyAnswer is used to capture a currency answer to a {Step}.
 class CurrencyAnswer < ApplicationRecord
   include TaskCounters
 
@@ -10,6 +11,11 @@ class CurrencyAnswer < ApplicationRecord
               message: "does not accept £ signs or other non numerical characters",
             }
 
+  # Overridden response accessor that ensures no commas are present in the currency value.
+  #
+  # @param [Float, String] value
+  #
+  # @return [Float]
   def response=(value)
     if value.is_a?(String)
       super(value.delete(","))
--- a/app/models/step.rb
+++ b/app/models/step.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+# A Step belongs to a {Task} and may have one of
+# {RadioAnswer}, {ShortTextAnswer}, {LongTextAnswer}, {SingleDateAnswer}, {CheckboxAnswers}, {NumberAnswer}, {CurrencyAnswer}.
 class Step < ApplicationRecord
   self.implicit_order_column = "order"
 
@@ -18,6 +20,9 @@ class Step < ApplicationRecord
 
   after_save :update_task_counters
 
+  # Returns the answer to this step.
+  #
+  # @return [Mixed]
   def answer
     @answer ||=
       case contentful_type
@@ -35,6 +40,10 @@ class Step < ApplicationRecord
     !answer.nil?
   end
 
+  # Returns the text for the primary call-to-action button.
+  # @see https://design-system.service.gov.uk/components/button/
+  #
+  # @return [String]
   def primary_call_to_action_text
     return I18n.t("generic.button.next") if super.blank?
 
@@ -45,10 +54,19 @@ class Step < ApplicationRecord
     skip_call_to_action_text.present?
   end
 
+  # This method fetches the {Journey} associated to this step's {Task}'s {Section}.
+  #
+  # @return [Journey]
   def journey
     task.section.journey
   end
 
+  # This method calls the owning {Task}'s save method to update its step tallies.
+  #
+  # This is used by the {TaskCounters} concern as a callback for when step answers
+  # are committed.
+  #
+  # @return [Boolean]
   def update_task_counters
     task.save
   end
--- a/app/models/task.rb
+++ b/app/models/task.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+# A Task belongs to a {Section} and consists of many {Step}s.
 class Task < ApplicationRecord
   self.implicit_order_column = "created_at"
 
@@ -16,6 +17,9 @@ class Task < ApplicationRecord
   IN_PROGRESS = 1
   COMPLETED = 2
 
+  # Returns all visible steps.
+  #
+  # @return [Step::ActiveRecord_AssociationRelation]
   def visible_steps
     steps.where(hidden: false)
   end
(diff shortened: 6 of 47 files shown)


Rakefile
```diff
@@ -3,9 +3,15 @@
 # Add your own tasks in files placed in lib/tasks ending in .rake,
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
+require 'yard'
```
peterdavidhamilton, line 6: Let's move this to a dedicated task file ala the comments. This way the Rake file just loads the app and sets the default.

app/models/step.rb
```diff
@@ -29,23 +40,40 @@ def answer
       end
   end
 
+  # Checks whether this step has been answered.
+  #
+  # @return [Boolean] `true` if this step has been answered; `false` otherwise.
```
peterdavidhamilton, line 45: Don't worry about annotating predicates like this. `@return [Boolean]` is clear enough.

Gemfile
```diff
@@ -53,6 +53,8 @@ group :development do
   gem "better_errors"
   gem "rails_layout"
   gem "spring-commands-rspec"
+  gem "yard"
+  gem "yard-junk"
```
peterdavidhamilton, line 57: We could leave off `yard` as being an implicit dependency of the junk plugin.

app/models/step.rb
```diff
@@ -29,23 +40,40 @@ def answer
...
   def answered?
     !answer.nil?
   end
 
+  # Returns the text for the primary call-to-action button.
+  #
+  # @return [String] the button text.
```
peterdavidhamilton, line 52: Maybe include references using [@see](https://rubydoc.info/gems/yard/file/docs/Tags.md#see), for example to https://design-system.service.gov.uk/components/button/

app/models/step.rb
```diff
@@ -29,23 +40,40 @@ def answer
...
 
+  # Returns the {Journey} this step belongs to.
+  # This method fetches the {Journey} associated to this step's {Task}'s {Section}.
+  #
+  # @return [Journey] the {Journey} this step belongs to.
   def journey
     task.section.journey
   end
 
+  # This method calls the owning {Task}'s save method to update its step tallies.
```
peterdavidhamilton, line 73: What I think would be valuable documentation here is that callbacks are being used. We can reference the new TaskCounters concern for example. Our aim is to add context, link to related code and add information about our code that might not be apparent.

peterdavidhamilton (commented):
@EdwinKruglov the intent of boilerplate Rails will be well known. Try tackling the service objects and what is being passed into classes and methods. That stuff is different in each project but where docs can be most valuable.

app/models/step.rb
```diff
@@ -16,6 +18,15 @@ class Step < ApplicationRecord
...
 
+  # Returns the answer to this step.
+  #
+  # @return [RadioAnswer] if this step has a radio button answer.
+  # @return [ShortTextAnswer] if this step has a short text answer.
+  # @return [LongTextAnswer] if this step has a long text answer.
+  # @return [SingleDateAnswer] if this step has a single date answer.
+  # @return [CheckboxAnswers] if this step has checkbox answers.
+  # @return [NumberAnswer] if this step has a number answer.
+  # @return [CurrencyAnswer] if this step has a currency answer.
```
peterdavidhamilton, line 29: Shorthand here could be `@return [Mixed]` (provided that isn't a genuine object) because it alerts you to it, or alternatively, we could adopt a `BaseAnswer` in a later refactor and that would suffice.

app/models/task.rb
```diff
@@ -9,33 +10,55 @@ class Task < ApplicationRecord
...
+  #
+  # @return [Integer] {NOT_STARTED}, {IN_PROGRESS} or {COMPLETED}
   def status
     return COMPLETED if all_steps_answered?
     return IN_PROGRESS if step_tally["answered"].positive?
 
     NOT_STARTED
   end
 
+  # Checks whether all steps within this task have been answered by comparing the number of visible steps
```
peterdavidhamilton, line 44: > Checks whether all steps within this task have been answered

The method name is unambiguous so I feel this isn't needed.

> by comparing the number of visible steps to the number of answered steps

The method is simple enough that this is easy to see, but if it were more complex, better to document the **what** and **why** not the **how**.

peterdavidhamilton (commented):
For good examples of succinct annotations check out the [dry-rb](https://dry-rb.org/) gems like [this one](https://github.com/dry-rb/dry-schema/blob/master/lib/dry/schema/key_map.rb).

app/models/step.rb
```diff
@@ -40,38 +36,36 @@ def answer
...
+  # @return [String]
   def primary_call_to_action_text
     return I18n.t("generic.button.next") unless super.present?
     super
   end
 
-  # Determines whether this step is skippable by checking for the presence of `skip_call_to_action_text`.
-  #
-  # @return [Boolean]
   def skippable?
```
peterdavidhamilton: For predicate methods just go with `@return [Boolean]` throughout, but if for example, the name seems odd, you can use `rails notes` to document that it should be revisited.

app/models/task.rb
```diff
@@ -1,4 +1,6 @@
-# This class defines the Task model. A Task belongs to a {Section} and consists of many {Step}s.
+# This class defines the Task model.
```
peterdavidhamilton, line 1: Can any relevant Contentful info be added here? I'll get you access now.

app/models/task.rb
```diff
@@ -10,23 +12,17 @@ class Task < ApplicationRecord
 
   attr_accessor
 
-  # The task has not been started.
   NOT_STARTED = 0
```
peterdavidhamilton: Unrelated to this PR, I like to use symbols here because they are more descriptive than integers.

app/models/task.rb
```diff
@@ -41,24 +37,22 @@ def status
...
   def visible_steps_with_answers
     eager_loaded_visible_steps.select(&:answered?)
   end
 
   # Returns the ID of the next unanswered step.
   #
-  # @return [String] ID of the next unanswered step; `nil` if all steps are answered.
+  # Returns `nil` if all steps are answered.
+  #
+  # @return [String]
```
peterdavidhamilton, line 55: Here the return value is either String or Nil, we should combine these three lines into one

app/models/activity_log_item.rb
```diff
@@ -1,3 +1,10 @@
+# This class defines the ActivityLogItem model.
```
peterdavidhamilton, line 1: I feel like this line is overly verbose. 
I'd prefer a header that conveys the purpose and intent of the class just like you've done below.

app/models/checkbox_answers.rb
```diff
@@ -8,6 +11,13 @@ class CheckboxAnswers < ActiveRecord::Base
     presence: true,
     unless: proc { |answer| answer.step.skippable? && answer.skipped }
 
+  # Overridden response accessor that ensures no blank checkbox values are set.
+  #
+  # Returns `nil` if `args` is blank.
+  #
+  # @param [Array] args
+  #
+  # @return [Array]
```
peterdavidhamilton, line 20: You can merge two of those lines `@return [Array, nil] array of answers or nil`

app/models/currency_answer.rb
```diff
@@ -6,6 +9,11 @@ class CurrencyAnswer < ActiveRecord::Base
 
   validates :response, presence: true, numericality: {greater_than_or_equal_to: 0, message: "does not accept £ signs or other non numerical characters"}
 
+  # Overridden response accessor that ensures no commas are present in the currency value.
+  #
+  # @param [Numeric/String] value
+  #
+  # @return [Numeric]
```
peterdavidhamilton, line 16: Just for giggles now we have Pry rather than IRB put `binding.pry` inside the top of the method and run the suite. See what value is and if it is always `Integer` and never `Float` for example.

peterdavidhamilton: We could put breakpoints in before but Pry is a great way to explore code if you're not already.

app/services/create_task.rb
```diff
@@ -9,6 +10,11 @@ def initialize(section:, contentful_task:, order:)
     @order = order
   end
 
+  # Creates and persists a new Task.
+  #
+  # This relies on the passed-in `contentful_task` to construct the object.
+  #
+  # @return [Task]
```
peterdavidhamilton, line 17: Please can you also use the [@raise](https://rubydoc.info/gems/yard/file/docs/Tags.md#raise) tag

app/services/create_task.rb
```diff
@@ -9,6 +10,11 @@ def initialize(section:, contentful_task:, order:)
     @order = order
   end
 
+  # Creates and persists a new Task.
+  #
+  # This relies on the passed-in `contentful_task` to construct the object.
```
peterdavidhamilton, line 15: What object is the `contentful_task`? Where did it come from?

app/services/create_task.rb
```diff
@@ -1,3 +1,4 @@
+# CreateTask service is responsible for constructing a {Task} for the given section.
```
peterdavidhamilton, line 1: In the example I linked to I think you'd see this parred down to:
`Build and persist a {Task}`
Note the same tense that we use in git commits with terse language. Short and sweet.

app/services/create_task.rb
```diff
@@ -9,6 +10,11 @@ def initialize(section:, contentful_task:, order:)
     @order = order
   end
 
+  # Creates and persists a new Task.
```
peterdavidhamilton, line 13: Same here, because it is a functional class this info is best kept at the top as a class description. So what other relevant info might a developer benefit from here? Who calls CreateTask#call for example?

peterdavidhamilton:
Before you commit please could you also include the @param for #initialize in the service objects.
bitmoji

peterdavidhamilton (approved):

AndrewCowle (approved):

PR #420, opened 2021-06-29
#112: Utility script for test suite in Docker

EdwinKruglov:
## Changes in this PR

- Add convenience script `script\spec` that runs our tests within Docker.
- Update documentation with `doc\clean-up.md` that shows how to kill and remove running containers and volumes.


peterdavidhamilton:
Rather than hard coding `rake`, can we have the script accept input that defaults to `rake`, but would allow passing `rspec`, `rubocop` or other commands. This way it is more flexible and we don't force the default task. One of the other scripts does this so you can copy that.

peterdavidhamilton (approved):

AndrewCowle (approved):

PR #421, opened 2021-07-01
#189: Create Category Model

EdwinKruglov:
## Changes in this PR

- Add new entity `Category`
- Update entity `Journey` with association to `Category`
- Add state to `Journey`: `initial`, `stale`, `archive`, `remove`
- Remove `liquid_template`, `last_worked_on`, and `contenful_id` columns from `Journey`
- Update the app to get liquid templates and contentful IDs for categories from `Category`, rather than `Journey`
- Update the app to use `Journey.updated_at` in place of `last_worked_on`
- Add rake task, `migrate_to_category`, to create a Catering category and associate any existing `Journeys` with it


AndrewCowle:
I'm reviewing this morning - give you a shout apres standup

app/models/journey.rb
```diff
@@ -8,8 +8,11 @@ class Journey < ApplicationRecord
   has_many :tasks, through: :sections, class_name: "Task"
   has_many :steps, through: :tasks, class_name: "Step"
   belongs_to :user
+  belongs_to :category, counter_cache: true
 
-  validates :category, :contentful_id, :liquid_template, presence: true
+  validates :contentful_id, :liquid_template, presence: true
+
+  enum state: { started: 0, abandoned: 1, completed: 2, archived: 3, stale: 4 }
```
AndrewCowle, line 15: Might be worth using state machine here so we can extend upon it later

https://github.com/aasm/aasm

db/migrate/20210630125559_create_categories.rb
```diff
@@ -0,0 +1,11 @@
+class CreateCategories < ActiveRecord::Migration[6.1]
+  def change
+    create_table :categories, id: :uuid do |t|
+      t.string :name, null: false
+      t.string :description
+      t.string :contentful_id, null: false
+      t.jsonb :liquid_template, null: false
+      t.timestamps
```
AndrewCowle, line 8: will also need a `t.integer :journeys_count` for the counter cache to work

AndrewCowle (changes requested):

spec/services/create_journey_spec.rb
```diff
@@ -56,21 +56,21 @@
...
-      described_class.new(category_id: "contentful-category-entry", user: build(:user)).call
+      described_class.new(category: build(:category, contentful_id: "contentful-category-entry"), user: build(:user)).call
 
       expect(Journey.last.liquid_template)
         .to eql("<article id='specification'><h1>Liquid {{templating}}</h1></article>")
     end
 
     context "when the journey cannot be saved" do
-      it "raises an error" do
+      xit "raises an error" do
```
EdwinKruglov, line 66: @AndrewCowle,  I've x'ed this one out for now. This test tries to assert that an exception is raised when a `Journey` is saved with `category = nil`. Now that `Category` is its own entity, `Journey.category` has been removed, so this test no longer applies. Contentful's `category.title` used to be assigned `Journey.category` and fixture `category-with-no-title` provides a category with no title. But now, since we're storing `Categories` in the DB, we're associating them to `Journeys` instead. This kind of test might be more appropriate for when we update our Contentful services to fetch all the categories and populate our DB with them.

app/models/journey.rb
```diff
@@ -2,14 +2,22 @@
 
 # A Journey belongs to a {User} and consists of {Section}s.
 class Journey < ApplicationRecord
+  include AASM
```
peterdavidhamilton, line 5: I don't think we should introduce a dependency with a new DSL this early on.

db/migrate/20210630125559_create_categories.rb
```diff
@@ -0,0 +1,11 @@
+class CreateCategories < ActiveRecord::Migration[6.1]
+  def change
+    create_table :categories, id: :uuid do |t|
+      t.string :name, null: false
```
peterdavidhamilton, line 4: I should've used `title` here so that it follows the same pattern as the other models. Can we swap it out please?

app/services/create_journey.rb
```diff
@@ -18,15 +18,16 @@ def initialize(category_id:, user:)
...
     journey = Journey.new(
-      category: category.title,
-      contentful_id: category.id,
+      category: category,
+      contentful_id: category_contentful_id,
       user: user,
       started: true,
       last_worked_on: Time.zone.now,
-      liquid_template: category.combined_specification_template,
+      liquid_template: contentful_category.combined_specification_template,
```
peterdavidhamilton, line 30: Shouldn't this be in the Category @AndrewCowle and do we need the `contentful_id` in the Journey now we are associated with a category?

app/controllers/journeys_controller.rb
```diff
@@ -58,7 +59,21 @@ def show
...
       journey_id: @journey.id,
       user_id: current_user.id,
-      contentful_category_id: @journey.contentful_id,
+      contentful_category_id: @journey.category.contentful_id,
     ).call
   end
+
+private
+
+  def get_category(category_id = nil)
```
peterdavidhamilton, line 68: If we make the exisiting category id the default param we can simplify this method:
```ruby
  def get_category(category_id = ENV["CONTENTFUL_DEFAULT_CATEGORY_ENTRY_ID"])
    Category.find_or_create_by!(contentful_id: category_id) do |category|
      contentful_category = GetCategory.new(category_entry_id: category.contentful_id).call
      category.title = contentful_category.title
      category.liquid_template = contentful_category.combined_specification_template
    end
  end
```

app/models/journey.rb
peterdavidhamilton: I know we have _state_ on the `Journey`, now I have thought about it, these states are wrong. We should have a quick chat to identify the correct _states_.

app/models/journey.rb
```diff
@@ -8,8 +8,9 @@ class Journey < ApplicationRecord
   has_many :tasks, through: :sections, class_name: "Task"
   has_many :steps, through: :tasks, class_name: "Step"
   belongs_to :user
+  belongs_to :category, counter_cache: true
 
-  validates :category, :contentful_id, :liquid_template, presence: true
+  enum state: { deleted: 0, archived: 1, stale: 2 }
```
peterdavidhamilton, line 13: Can we change this so the default isn't `deleted` and escalates in severity using the present tense?
It would be good to add comments to clarify the reasoning.

```ruby
{  default: 0, stale: 1, archive: 2, delete: 3 }
```

- default: I'm sure a better word exists...
- stale: automatic flag the journey as being untouched for a while and something we could maybe purge
- archive: user-defined flag so it could be hidden from their dashboard
- delete: user-defined flag so it can be safely deleted (the _soft delete_ functionality)

app/models/journey.rb
```diff
@@ -8,8 +8,9 @@ class Journey < ApplicationRecord
   has_many :tasks, through: :sections, class_name: "Task"
   has_many :steps, through: :tasks, class_name: "Step"
   belongs_to :user
```
peterdavidhamilton, line 10: Please could we move the `belongs_to` above the `has_many` with a line in between? 
bitmoji

AndrewCowle (approved):

peterdavidhamilton:
When you finally merge it @EdwinKruglov be sure to have a nice tidy commit message that lists all the changes.

PR #422, opened 2021-07-02
Next un-answered question

AndrewCowle:
Something weird with the ordering going on - will wait to pair with Peter


peterdavidhamilton:
Add this to `Task` to maintain the order.

```ruby
  self.implicit_order_column = "order"

  default_scope { order(:order) }
```

PR #424, opened 2021-07-05
Ignore tool-verions file

AndrewCowle:
Ignore .tool-verions (preferred file convention for ASDF tooling)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -118,4 +118,5 @@ terraform.rc
 # ERD file
 doc/erd.pdf
 
+# File where software versions are managed by asdf
 .tool-versions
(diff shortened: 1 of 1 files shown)


EdwinKruglov (approved):

.gitignore
```diff
@@ -117,3 +117,5 @@ terraform.rc
 
 # ERD file
 doc/erd.pdf
+
```
peterdavidhamilton, line 120: It would be good to include your PR comment here as a note for those unfamiliar with ASDF

peterdavidhamilton (approved):

PR #429, opened 2021-07-09
Next unanswered section

AndrewCowle:
The current implementation of _**next unanswered task**_ needs to be reworked and tested with a larger fixture. It needs to take the user to the next unfinished section, and also return the user to a previous section if they have reached the end.


peterdavidhamilton:
Can this PR use the revised template please

app/models/journey.rb
```diff
@@ -41,4 +41,31 @@ def start!
...
+    # NB: simplify/railsify this?
+    order_by = %(
+      CASE
+        WHEN tasks.section_id = '#{current_task.section_id}' THEN
+          CASE WHEN tasks.order > #{current_task.order} THEN
+            1
+          ELSE
+            2
+          END
+        ELSE
```
AndrewCowle, line 59: I would be inclined to have separate methods with the full query written - the switch statement defining which method to call - would be easier to read the full query and maybe easier / clear testing - what do you think @peterdavidhamilton ?

PR #435, opened 2021-07-15
#216: Multi-category support Contentful Entry Map page

EdwinKruglov:
## Changes in this PR

- Replace `journey_maps#new` which used the hard-coded default category, with `#index` listing all categories and `#show` for the chosen category.

### /journey_maps

image


peterdavidhamilton:
Would you mind rewriting the pull request text:

_Add an index page `/journey_maps` that lists all the categories and links to the right map at `/journey_maps/:id` where it shows the tasks for that category._

to something like 

_Replace `journey_maps#new` which used the hard-coded default category, with `#index` listing all categories and `#show` for the chosen category._

peterdavidhamilton (approved):

AndrewCowle (approved):

PR #436, opened 2021-07-15
Refactor CreateStep service object

EdwinKruglov:
## Changes in this PR

- rename `CreateStep.contentful_entry` to `contentful_step` to reflect the service object and decrease ambiguity


peterdavidhamilton:
Please could you add the _**"why"**_ to PR and commit. We are trying to align the service objects, be more descriptive, less ambiguous etc. For the same reason I feel like you should rename `fake_entry` in the specs to `contentful_step` as it mirrors the class under test and better conveys what is happening.

app/services/create_step.rb
```diff
@@ -32,15 +32,15 @@ class UnexpectedContentfulStepType < StandardError; end
 
   ALLOWED_STEP_TYPES = ALLOWED_CONTENTFUL_QUESTION_TYPES + ALLOWED_CONTENTFUL_STATEMENT_TYPES
```
peterdavidhamilton, line 34: How would you document the `attr_accessor`s?

app/services/create_step.rb
```diff
@@ -110,69 +110,69 @@ def unexpected_step_type?
...
+    return nil unless contentful_step.respond_to?(:extended_options)
 
-    contentful_entry.extended_options
+    contentful_step.extended_options
   end
 
   # @return [String]
   def step_type
-    contentful_entry.type.tr(" ", "_")
+    contentful_step.type.tr(" ", "_")
```
peterdavidhamilton, line 139: Don't change it, but do we still need `#tr`?

spec/services/create_step_spec.rb
```diff
@@ -253,7 +253,7 @@
                 allowed_content_models: CreateStep::ALLOWED_CONTENTFUL_MODELS.join(", "),
```
peterdavidhamilton, line 253: We are testing against internals again. It would be better to see the actual string of models here.

spec/services/create_step_spec.rb
```diff
@@ -294,7 +294,7 @@
...
                 allowed_step_types: CreateStep::ALLOWED_STEP_TYPES.join(", "))
```
peterdavidhamilton, line 295: ditto

spec/services/create_step_spec.rb
```diff
@@ -148,7 +148,7 @@
         )
 
         step, _answer = described_class.new(
-          task: task, contentful_entry: fake_entry, order: 0,
+          task: task, contentful_step: fake_entry, order: 0,
         ).call
 
         expect(step.primary_call_to_action_text).to eq(I18n.t("generic.button.next"))
```
peterdavidhamilton, line 154: Pure string opportunity here

app/services/create_step.rb
```diff
@@ -110,69 +110,69 @@ def unexpected_step_type?
...
-    contentful_entry.skip_call_to_action
+    contentful_step.skip_call_to_action
   end
 
   # @return [Boolean]
   def hidden?
-    return false unless contentful_entry.respond_to?(:always_show_the_user)
-    return false if contentful_entry.always_show_the_user.nil?
+    return false unless contentful_step.respond_to?(:always_show_the_user)
+    return false if contentful_step.always_show_the_user.nil?
```
peterdavidhamilton, line 161: Would it read better as `unless`?

app/services/create_step.rb
```diff
@@ -110,69 +110,69 @@ def unexpected_step_type?
...
   # @return [Nil, String]
   def skip_call_to_action_text
-    return nil unless contentful_entry.respond_to?(:skip_call_to_action)
+    return nil unless contentful_step.respond_to?(:skip_call_to_action)
 
-    contentful_entry.skip_call_to_action
+    contentful_step.skip_call_to_action
   end
 
   # @return [Boolean]
```
peterdavidhamilton, line 158: Can you add anything in the doc to describe the logic/reasoning here?

app/services/create_step.rb
```diff
@@ -80,12 +80,12 @@ def call
 
   # @return [String]
```
peterdavidhamilton, line 81: What does an id look like, is it `snake_case_string` or `37274c18-1d75-48ed-a662-96801dcffeae`?

spec/services/create_step_spec.rb
```diff
@@ -129,7 +129,7 @@
         )
 
         step, _answer = described_class.new(
-          task: task, contentful_entry: fake_entry, order: 0,
+          task: task, contentful_step: fake_entry, order: 0,
         ).call
 
         expect(step.primary_call_to_action_text).to eq("Go onwards!")
```
peterdavidhamilton, line 135: Personally, I like to remove brackets if they serve no purpose in specs to make it more like a sentence.

peterdavidhamilton:
Does the spec still work if you start using `let` for a shared setup?

app/services/create_step.rb
```diff
@@ -110,69 +110,69 @@ def unexpected_step_type?
...
+    return nil unless contentful_step.respond_to?(:help_text)
 
-    contentful_entry.help_text
+    contentful_step.help_text
   end
 
   # @return [Nil, String]
   def body
-    return nil unless contentful_entry.respond_to?(:body)
+    return nil unless contentful_step.respond_to?(:body)
```
peterdavidhamilton, line 125: [Argument for not having a guard clause here because it is not guarding from anything expensive.](https://codeclimate.com/github/DFE-Digital/buy-for-your-school/pull/436#readup-60f05e76741bee000100016d)

peterdavidhamilton:
At 30 lines long lets pull `Step.create` into its own private method @EdwinKruglov

peterdavidhamilton:
Any comments @AndrewCowle ?

AndrewCowle (approved):
happy with if @peterdavidhamilton's changes are made :)

app/services/create_step.rb
EdwinKruglov: I can change it to `attr_reader`, but I can't figure out how to document the individual variables. I feel like the documentation on the constructor describes them pretty well?

app/services/create_step.rb
EdwinKruglov: It looks like we do. The types returned from Contentful seem to have spaces in them.

peterdavidhamilton: What are they?

app/services/create_step.rb
peterdavidhamilton: `ToggleAdditionalSteps` has an example - I confess I have slipped in some invalid syntax because for some reason I was adding colons where they shouldn't be, but the `attr_reader`s are good.

peterdavidhamilton: If nothing needs the setter method then `attr_reader` is sufficient, to be fair if no external access is required then no reader is required.

app/services/create_step.rb
EdwinKruglov: `["number", "radios", "short text", "markdown", "checkboxes", "long text", "single_date"]`
Interestingly, `single_date` does have an underscore.

app/services/create_step.rb
EdwinKruglov: Like this: `1QdzZOVfL8x1d9Q9FA0u66`.

app/services/create_step.rb
EdwinKruglov: Without it I get an exception when the `body` field doesn't exist, in one of the tests at least.

app/services/create_step.rb
peterdavidhamilton: Screenshot 2021-07-21 at 09 55 52
Chat to the designers to understand the consequences should we want to change the type in Contentful to always use an underscore.

spec/services/create_step_spec.rb
```diff
@@ -1,29 +1,32 @@
...
-        category = create(:category, :catering)
-        journey = create(:journey, category: category)
-        section = create(:section, journey: journey)
-        task = create(:task, section: section)
+        # category = create(:category, :catering)
+        # journey = create(:journey, category: category)
+        # section = create(:section, journey: journey)
+        # task = create(:task, section: section)
 
         fake_entry = fake_contentful_step(
```
peterdavidhamilton, line 15: fake_entry -> contentful_step

spec/services/create_step_spec.rb
```diff
@@ -1,29 +1,32 @@
-require "rails_helper"
-
 RSpec.describe CreateStep do
   describe "#call" do
+    let(:category) { create(:category, :catering) }
+    let(:journey) { create(:journey, category: category) }
+    let(:section) { create(:section, journey: journey) }
+    let(:task) { create(:task, section: section) }
```
peterdavidhamilton, line 6: Wer can use the same `let(:fixture)` approach, combined with another where the described class is called, to remove much duplication. We don't need the answer but should always show it is returned `_answer`.

app/services/create_step.rb
```diff
@@ -32,16 +32,21 @@ class UnexpectedContentfulStepType < StandardError; end
...
-    self.task = task
-    self.contentful_entry = contentful_entry
-    self.order = order
+  def initialize(task:, contentful_step:, order:)
+    @task = task
+    @contentful_step = contentful_step
+    @order = order
   end
 
   # @return [Step]
```
peterdavidhamilton, line 52: Unless I just can't see it we need `@raise` here

app/services/create_step.rb
peterdavidhamilton: We discussed accessing the purely `Hash` interface of fields like `contentful_entry.fields[:body]`. The double doesn't expect this though. All these methods can be condensed and the double in `fake_contentful_step` can drop the mock methods for the fields as a result. That's my thought - might be wrong though?

spec/services/create_step_spec.rb
peterdavidhamilton: don't forget to remove old comments

app/services/create_step.rb
peterdavidhamilton: Can you add a comment explaining that steps can have different fields and that the `body` is for statements? Also, please document which ones relate to specific question types.

app/services/create_step.rb
peterdavidhamilton: You now know how to document the `reader`/`accessor`, but is the only interface on this `#call`? If so we can remove the `attrs` here completely.

app/services/create_step.rb
```diff
@@ -120,72 +120,62 @@ def unexpected_step_type?
...
   end
 
   # Determines if this step should be hidden in the task list.
   # Applies to steps dependent on answers of other steps.
   #
   # @return [Boolean]
   def hidden?
-    return false unless contentful_step.respond_to?(:always_show_the_user)
-    return false if contentful_step.always_show_the_user.nil?
+    return false if @contentful_step.fields[:always_show_the_user].nil?
```
peterdavidhamilton, line 166: This can still be one line with double negation

spec/jobs/warm_entry_cache_job_spec.rb
```diff
@@ -32,7 +32,7 @@
...
+          "\"{\\\"sys\\\":{\\\"space\\\":{\\\"sys\\\":{\\\"type\\\":\\\"Link\\\",\\\"link_type\\\":\\\"Space\\\",\\\"id\\\":\\\"jspwts36h1os\\\"}},\\\"id\\\":\\\"radio-question\\\",\\\"type\\\":\\\"Entry\\\",\\\"created_at\\\":\\\"2020-09-07T10:56:40.585Z\\\",\\\"updated_at\\\":\\\"2020-09-14T22:16:54.633Z\\\",\\\"environment\\\":{\\\"sys\\\":{\\\"id\\\":\\\"master\\\",\\\"type\\\":\\\"Link\\\",\\\"link_type\\\":\\\"Environment\\\"}},\\\"revision\\\":7,\\\"content_type\\\":{\\\"sys\\\":{\\\"type\\\":\\\"Link\\\",\\\"link_type\\\":\\\"ContentType\\\",\\\"id\\\":\\\"question\\\"}},\\\"locale\\\":\\\"en-US\\\"},\\\"fields\\\":{\\\"slug\\\":\\\"/which-service\\\",\\\"title\\\":\\\"Which service do you need?\\\",\\\"help_text\\\":\\\"Tell us which service you need.\\\",\\\"type\\\":\\\"radios\\\",\\\"extended_options\\\":[{\\\"value\\\":\\\"Catering\\\"},{\\\"value\\\":\\\"Cleaning\\\"}],\\\"always_show_the_user\\\":true}}\"",
```
peterdavidhamilton, line 35: Is this what a genuine Contentful::Entry returns though?

spec/services/create_step_spec.rb
```diff
@@ -58,117 +55,99 @@
...
 
     context "when no 'skipCallToAction' is provided" do
-      it "default copy is used for the button" do
-        fake_entry = fake_contentful_step(
-          contentful_fixture_filename: "steps/skippable-checkboxes-question.json",
-        )
+      let(:fixture) { "skippable-checkboxes-question" }
 
+      it "default copy is used for the button" do
         step, _answer = described_class.new(
```
peterdavidhamilton, line 125: One of my previous comments was about reducing all this duplication using `let`

app/services/create_step.rb
```diff
@@ -34,10 +35,6 @@ class UnexpectedContentfulStepType < StandardError; end
 
   # @return [Task]
   attr_reader :task
```
peterdavidhamilton, line 37: Can we not just use `@task` in `Step.create` too?

spec/support/contentful_helpers.rb
```diff
@@ -182,22 +182,18 @@ def fake_contentful_task(contentful_fixture_filename:)
   def fake_contentful_step(contentful_fixture_filename:)
     raw_response = File.read(Rails.root.join("spec/fixtures/contentful/004-#{contentful_fixture_filename}"))
     hash_response = JSON.parse(raw_response)
+    hash_response.deep_transform_keys!(&:underscore)
+    hash_response.deep_symbolize_keys!
 
     double(
```
peterdavidhamilton, line 188: Remember our goal is to not force tests to pass but mimic what Contentful would give us.

EdwinKruglov: From what I've seen, genuine `Contenful::Entry` objects have their hash entries symbolised and snake-cased. So maybe a less drastic change would be to only make these changes to `hash_response["fields"]`, rather than the whole thing.

This all started with `contentful_caching_spec` failing the test `checks the Redis cache instead of making an external request`, because there it reaches the method `CacheableEntry#find_and_build_entry_from_cache`, where `Contentful::ResourceBuilder` takes the raw JSON and builds a proper `Contentful::Entry`, and there everything is a symbol and snake-cased.

So in `CreateStep`, if I call `fields["type"]` the majority of the tests will pass, but in the actual application that call will fail.

app/services/create_step.rb
EdwinKruglov: My brain is probably fried, but double negation would ensure we get `false` if the key is `nil`, but it wouldn't flip the boolean value of `:always_show_the_user` if it is there, which is what line 168 achieves? Unless I'm completely misunderstanding what you're saying?

peterdavidhamilton: You're right and it's because I keep thinking that it's the inverse (always_show_the_user=true), I think it works if that field had false in it. Thank you for the follow up.

spec/support/contentful_helpers.rb
peterdavidhamilton: What drew my attention was that the raw JSON in other specs need to change for it to work, indicating we'd gone too far. If fields are all snake_case symbols great, but I'd not expect our tweak to cause ripples into other specs. I've checked out the code now to see if I'm talking rubbish again. ;p

peterdavidhamilton:
We'd spotted that the spec had `service.call` sometimes appear to return an unused answer. That is never the case so it fails to document the object under test and would be confusing.

spec/support/contentful_helpers.rb
EdwinKruglov: Yeah, I agree that having to change the JSON in other specs is too far. I can fix this by having the double `Contentful::Entry#raw` return the JSON hash before we transform it, which would reflect what the actual `Contentful::Entry#raw` returns.

peterdavidhamilton:
If you revert the request and job specs (which shouldn't be effected) and try this instead, we should be good.

```ruby
  def fake_contentful_step(contentful_fixture_filename:)
    raw_response = File.read(Rails.root.join("spec/fixtures/contentful/004-#{contentful_fixture_filename}"))
    hash_response = JSON.parse(raw_response)

    double(
      Contentful::Entry,
      id: hash_response.dig("sys", "id"),
      content_type: double(id: hash_response.dig("sys", "contentType", "sys", "id")),
      raw: hash_response,
      fields: hash_response["fields"].dup.deep_transform_keys!(&:underscore).deep_symbolize_keys!,
    )
  end
```

Notice `dup` is in there too.

spec/support/contentful_helpers.rb
```diff
@@ -187,17 +187,10 @@ def fake_contentful_step(contentful_fixture_filename:)
       Contentful::Entry,
       id: hash_response.dig("sys", "id"),
       title: hash_response.dig("fields", "title"),
```
EdwinKruglov, line 189: I've left `title` and `type` in here because `journey_maps/show.html.erb` uses those methods and so `journey_map_spec` relies on these mocks.

app/services/create_step.rb
```diff
@@ -120,72 +120,62 @@ def unexpected_step_type?
...
 
   # @return [Nil, Array<Hash>]
   def additional_step_rules
-    return nil unless contentful_step.respond_to?(:show_additional_question)
-
-    contentful_step.show_additional_question
+    @contentful_step.fields[:show_additional_question]
   end
 
   # @return [String]
```
peterdavidhamilton, line 173: I think it would be good to add little documentation to highlight the bits of our fixtures that are important to us. If you use `@example` and make quite a healthy size comment with some parred down JSON

spec/support/contentful_helpers.rb
peterdavidhamilton: Cool, I have just spotted this so please confirm/ignore, `AnswersController#further_information_params` shows that JSON field values aren't messed with, only the field keys; so we should be using `deep`.

peterdavidhamilton: Also, you've probably noticed that I'm angling for `Contentful::Entry` to disappear and just have a `Hash` interface.

peterdavidhamilton: Something is telling me `AnswerHelper` shouldn't exist

app/services/create_step.rb
peterdavidhamilton: Please could you add that to the `@return` doc. Always nice to be reminded of what is going to come out of a method.

app/services/create_step.rb
peterdavidhamilton: The one-liner version of this, that should also satisfy the two Codeclimate warnings, is:
```ruby
!@contentful_step.fields.fetch(:always_show_the_user, true)
```

Also be aware that new functionality appeared in Contentful this week that would allow our designers to set a default value to fields. 
Screenshot 2021-07-21 at 09 54 33

spec/services/create_step_spec.rb
```diff
@@ -291,10 +197,10 @@
                 contentful_entry_id: "unexpected-contentful-question-type",
                 content_model: "question",
                 step_type: "telepathy",
-                allowed_content_models: CreateStep::ALLOWED_CONTENTFUL_MODELS.join(", "),
-                allowed_step_types: CreateStep::ALLOWED_STEP_TYPES.join(", "))
+                allowed_content_models: "question, statement",
+                allowed_step_types: "long_text, short_text, checkboxes, radios, currency, number, single_date, markdown")
           .and_call_original
-        expect { described_class.new(task: task, contentful_entry: fake_entry, order: 0).call }
+        expect { service.call }
```
peterdavidhamilton, line 203: Now it's so much shorter can we get things on one line?

AndrewCowle (approved):

peterdavidhamilton (approved):
image

PR #440, opened 2021-07-20
User-friendly URLs for journey maps

EdwinKruglov:
## Changes in this PR

- rename `journey_maps` to `design` for a shorter URL
- replace Contenful IDs with slugs in the `design` route for user-friendly URLs


peterdavidhamilton:
We'll need yours to rebase once my spec suite changes are merged @EdwinKruglov so we can revisit this once that is done.

spec/features/design/design_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+      expect(find("h1.govuk-heading-l")).to have_text "No categories found"
+    end
+  end
+
+  context "when there are categories" do
+    let(:fixture) { "journey-with-multiple-entries" }
+
+    context "when on index page" do
+      before do
+        stub_multiple_contentful_categories(category_fixtures: [
```
peterdavidhamilton, line 27: We need to stub `mfd` and `catering` to do this properly and the last few PRs got this prepared. Hopefully, we can tweak `stub_multiple_contentful_categories` to achieve that here.

spec/features/design/design_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+
+          within(list_items[0]) do
+            expect(find("a.govuk-link")).to have_text "Catering"
+            expect(find("a.govuk-link")[:href]).to eq "/design/catering"
+          end
+        end
+      end
+    end
+
+    context "when on show page" do
```
peterdavidhamilton, line 53: We should have the feature spec for `show` in its own file.

peterdavidhamilton: The context here is `when it is valid` because `context` flips state. My recent PR attempts to summarise the approach to specs.

spec/features/design/design_spec.rb
```diff
@@ -0,0 +1,128 @@
+RSpec.feature "Content Designers can see" do
+  before do
+    user_is_signed_in
+  end
+
+  context "when there are no categories" do
+    before do
+      contentful_connector = stub_contentful_connector
```
peterdavidhamilton, line 8: I think some of this setup is duplicating the `stub_multiple_contentful_categories` helper.

AndrewCowle (approved):

spec/features/design/design_spec.rb
```diff
@@ -20,107 +15,36 @@
...
-                                      href: "https://app.contentful.com/spaces/test/environments/master/entries/long-text-question"
+    it "lists expected categories" do
+      within("ul.govuk-list.govuk-list--bullet") do
+        list_items = find_all("li")
 
-            # design.preview_step_link_text
-            expect(page).to have_link "Preview step in service", href: "/preview/entries/long-text-question"
-          end
+        within(list_items[0]) do
+          expect(find("a.govuk-link")).to have_text "Catering"
```
peterdavidhamilton, line 41: I think I favour `have_link` with `class:` added because it affirms tag, class, text and link all in one. 
I feel this reads better but also I think it works if there are multiple links? What's your opinion?

spec/features/design/show_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+        list_items = find_all("li")
+
+        within(list_items[0]) do
+          expect(find("b")).to have_text "Which service do you need?"
+
+          # step_id internal tag
+          expect(find("code")).to have_text "{{ answer_radio-question }}"
+
+          # design.edit_step_link_text
+          expect(page).to have_link "Edit step in Contentful",
```
peterdavidhamilton, line 26: Please could we check the class for links here too

spec/features/design/design_spec.rb
EdwinKruglov: You're right, it looks much nicer that way. Wasn't aware I could do all that with `have_link`!

peterdavidhamilton (approved):

PR #451, opened 2021-07-27
Establish connection to Contentful preview content in each environment

peterdavidhamilton:
## Changes in this PR

- rewrite connection logic under the `content` namespace
- establish a connection to both delivery and preview endpoints
  + preview is exclusively used within `Preview::EntriesController`
- rely on our client object for determining connection config when reporting errors
  + remove ENV VARS where possible
- reuse a disposable journey for each designer when they preview steps
  + the journey will be cleaned up automatically and is hidden from the dashboard
- delete unused preview environment view template logic and replace with a flash notice
- update spec stubbing and service objects
- improve `/design/<category>` pages
  + remove redundant preview links, as only published steps appear here
  + make the step title a link to edit in Contentful

### Before

`/designs/mfd`
Screenshot 2021-07-28 at 16 20 38

### After

`/designs/mfd`
Screenshot 2021-07-28 at 16 15 55

A previewed step from Contentful
Screenshot 2021-07-28 at 16 16 37


app/controllers/preview/entries_controller.rb
```diff
@@ -1,14 +1,41 @@
...
+    end
+
+    section = Section.find_or_create_by!(title: "Designer Preview Section") do |s|
+      s.contentful_id = 0
+      s.order = 0
+      s.journey = journey
+    end
+
+    task = Task.find_or_create_by!(title: "Designer Preview Task") do |t|
+      t.contentful_id = 0
```
AndrewCowle, line 26: I've done the same with my category work - not keen on this logic in the controllers, guess we can wrap this into another factory style object later

app/controllers/design_controller.rb
```diff
@@ -21,17 +21,37 @@ class DesignController < ApplicationController
...
+  end
+
+private
+
+  def client
+    Content::Client.new
+  end
 
-    flash[:notice] = "#{contentful_category.environment.id.capitalize} Environment"
+  def env_banner
```
AndrewCowle, line 54: Maybe we could have a helper that includes all the banners?

app/services/create_step.rb
```diff
@@ -216,17 +216,17 @@ def additional_step_rules
   #     }
   #   }
   #
-  # @return [String] JSON
+  # @return [Hash] JSON
   def raw
     @contentful_step.raw
   end
 
   def send_rollbar_warning
```
AndrewCowle, line 224: I've see this method in alot of services - would be nice to refactor this into a service that accepts various opts to send the message.

peterdavidhamilton: It is the cause of many CodeClimate flags but certainly post-assessment.

app/controllers/design_controller.rb
peterdavidhamilton: I agree. There are a few comments about our `Contentful::Entry` objects needing decorators to encapsulate view logic floating around.

app/controllers/preview/entries_controller.rb
peterdavidhamilton: Certainly. The creation of these chained associations is repeated throughout and needs to be wrapped in a transaction to protect from orphans.

EdwinKruglov (approved):

AndrewCowle (approved):

peterdavidhamilton:
@AndrewCowle are you happy for this to be merged after having tested it locally?

AndrewCowle:
@peterdavidhamilton happy to merge

PR #454, opened 2021-07-29
Enable manual triggering of Github actions (30 day limit)

peterdavidhamilton:
## Changes in this PR

The option to manually retrigger deployment would be useful if variables in GH secrets are updated for example.
There is a 30 day maximum restriction to rerunning actions.

`gh` has also been added to the Brewfile for convenience.
https://docs.github.com/en/actions/managing-workflow-runs/re-running-a-workflow


peterdavidhamilton:
Is this all that is required? @EdwinKruglov @AndrewCowle 
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#example

AndrewCowle (approved):

EdwinKruglov (approved):

.github/workflows/deploy.yml
```diff
@@ -1,6 +1,7 @@
 name: deploy
 
 on:
+  workflow_dispatch:
```
peterdavidhamilton, line 4: Is this line here all that is required to get a manual trigger function?

PR #473, opened 2021-08-09
Refactor activity logging

peterdavidhamilton:
## Changes in this PR

- add `ActivityLogItem` model validation and spec coverage
- start to group activity logging specs and tidy changed files
- add failed validation context to spec covering question answering


spec/features/steps/answer_question_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  context "when the only question is answered" do
+    let(:answer) { create(:short_text_answer, response: "answer") }
+
+    scenario "the user is returned to the same position on the journey page" do
+      fill_in "answer[response]", with: "email@example.com"
+      click_update
+
+      expect(find("strong.app-task-list__tag")).to have_text "Completed"
+      expect(page.current_url).to eql "http://www.example.com:3000/journeys/#{journey.id}##{answer.step.id}"
```
AndrewCowle, line 25: I think we use current_url & current_path across feature specs, I would prefer we use current path so we can omit the domain name from specs

spec/services/record_action_spec.rb
```diff
@@ -9,7 +7,7 @@
         user_id: "23456789",
       ).call
 
-      expect(action.action).to eq("begin_journey")
+      expect(action.action).to eq "begin_journey"
```
AndrewCowle, line 10: Could we rename the var `action = described_class....` so that it doesn't read `action.action` - it just reads weird in my head... picky I know :/

peterdavidhamilton: That's the exact level of pedantry I'm aiming at. 👍🏼 

My pattern for this is, as seen in other service object tests, is to establish the `subject(:service)` and then `call` it. I'd be inclined to use `activity` as the variable name at that point, does that sound ok? 

There will be a few more iterations over these specs before we are done. I'm not going to make this change in this PR because I also think we need to reduce the params for `RecordAction` by having a hash grouping the UUIDs so this is just the first pass.

AndrewCowle: Yes CodeClimate was having a wobbler about the number of params - putting it all into a hash would solve this :)

EdwinKruglov (approved):

PR #476, opened 2021-08-11
Added new fields to user

AndrewCowle:
## Changes in this PR

Added full_name, email_address, phone_number, contact_preferences to User.

## Next steps

Create `SupportRequest` model


peterdavidhamilton:
@AndrewCowle we've discussed how data models will be influx for a bit, but as we go forward please add a tests that cover AR fields and data types.

PR #478, opened 2021-08-12
Support Request Model

AndrewCowle:
## Changes in this PR

- Added `SupportRequest` model, factory and spec
- User has_many `SupportRequests`

## Next steps

- Build `Schools` and support request UI flow out

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_08_05_130832) do
+ActiveRecord::Schema.define(version: 2021_08_11_120013) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "pgcrypto"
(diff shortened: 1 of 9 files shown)


db/schema.rb
```diff
@@ -151,6 +151,16 @@
     t.index ["task_id"], name: "index_steps_on_task_id"
   end
 
+  create_table "support_requests", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.jsonb "specification_ids", default: {}
+    t.jsonb "category_ids", default: {}
```
peterdavidhamilton, line 156: What are `category_id`s for?

spec/models/support_request_spec.rb
```diff
@@ -0,0 +1,10 @@
+RSpec.describe SupportRequest, type: :model do
```
peterdavidhamilton, line 1: Can we get an assertion that creates a Support request and asserts the fields please.

db/schema.rb
```diff
@@ -151,6 +151,16 @@
...
+    t.string "message"
+    t.bigint "user_id"
```
peterdavidhamilton, line 158: Should this be `UUID`? I'd also like us to start getting persistence level checks like `null: false` in our migrations.

PR #484, opened 2021-08-17
Allow steps to be skipped

EdwinKruglov:
## Changes in this PR

- add `skipped_ids` field to `Task`, skip interface to `Step`, and `Skip for now` button to the question page to allow users to skip questions

### Before

image

### After

image

## Next steps

- the interface provided by this ticket allows for a future ticket to skip statment steps

### DIFF
--- a/app/controllers/answers_controller.rb
+++ b/app/controllers/answers_controller.rb
@@ -41,7 +41,7 @@ class AnswersController < ApplicationController
     ).call
 
     if result.success?
-      parent_task.save! if parent_task.skipped_ids.delete(@step.id)
+      @step.unskip! if @step.skipped?
 
       if parent_task.has_single_visible_step?
         redirect_to journey_path(@journey, anchor: @step.id)
--- a/app/controllers/answers_controller.rb
+++ b/app/controllers/answers_controller.rb
@@ -41,6 +41,8 @@ class AnswersController < ApplicationController
     ).call
 
     if result.success?
+      parent_task.save! if parent_task.skipped_ids.delete(@step.id)
+
       if parent_task.has_single_visible_step?
         redirect_to journey_path(@journey, anchor: @step.id)
       elsif parent_task.all_steps_completed?
@@ -175,6 +177,8 @@ private
     { response: format_date(date_hash) }
   end
 
+  # Specific to checkboxes
+  #
   # @return [Boolean]
   def skip_answer?
     params.fetch("skip", false)
--- a/app/controllers/steps_controller.rb
+++ b/app/controllers/steps_controller.rb
@@ -71,23 +71,25 @@ class StepsController < ApplicationController
   end
 
   def update
-    @journey = current_journey
+    journey = current_journey
     @step = Step.find(params[:id])
 
-    parent_task.skipped_ids << @step.id unless parent_task.skipped_ids.include?(@step.id)
-    parent_task.save!
+    @step.skip!
 
+    # allow the user to skip a step and come back to it later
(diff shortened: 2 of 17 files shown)


app/models/task.rb
```diff
@@ -123,6 +139,7 @@ def tally_steps
       acknowledged: statement_ids.count,                    # visible completed statement steps
       questions: visible_steps.that_are_questions.count,    # visible question steps
       answered: visible_questions_with_answers.count,       # visible completed question steps
+      skipped: skipped_ids.count,                           # visible skipped question steps
```
peterdavidhamilton, line 142: I haven't pulled this yet but I guess you need to check the presence of `skipped_ids` first which will be nil in those migrations.

db/schema.rb
```diff
@@ -160,8 +160,10 @@
     t.integer "order"
     t.jsonb "step_tally", default: "{}"
     t.text "statement_ids", default: [], null: false, array: true
+    t.text "skipped_ids", default: [], null: false, array: true
```
AndrewCowle, line 163: Probably also want to add_index using: 'gin' on array fields :)

app/controllers/steps_controller.rb
```diff
@@ -70,6 +70,27 @@ def edit
     render "steps/#{@step.contentful_type}", locals: { layout: "steps/edit_form_wrapper" }
   end
 
+  def update
+    @journey = current_journey
```
peterdavidhamilton, line 74: Are we rendering so do we need instance variables?

app/controllers/steps_controller.rb
```diff
@@ -70,6 +70,27 @@ def edit
...
+    @step = Step.find(params[:id])
+
+    parent_task.skipped_ids << @step.id unless parent_task.skipped_ids.include?(@step.id)
+    parent_task.save!
```
peterdavidhamilton, line 78: - Would it make sense to move this logic to the task so it can be responsible?
- Does this give us a greater opportunity to unit test?

app/controllers/steps_controller.rb
```diff
@@ -70,6 +70,27 @@ def edit
...
+
+    if parent_task.has_single_visible_step?
```
peterdavidhamilton, line 80: It'd be good to have a descriptive note here to explain the business logic not the code.

spec/features/steps/skip_question_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+
+        it "returns to the task page" do
+          expect(page).to have_a_task_path
+        end
+      end
+    end
+
+    context "when all steps are skipped" do
+      before do
+        task.skipped_ids << number_step.id << checkbox_step.id << radio_step.id
```
peterdavidhamilton, line 97: bitmoji

app/controllers/steps_controller.rb
peterdavidhamilton: I see you were referring to the `statement_ids` further down the file. It was me but I haven't left it tidy for you. We know have `acknowledge!` and `acknowledged?` on the step. I think that is the better pattern so if we apply that here the `Step` is responsible. Just try and get it into the model and add note for us to tidy up that statement logic later.

app/controllers/answers_controller.rb
```diff
@@ -41,6 +41,8 @@ def create
     ).call
 
     if result.success?
+      @step.unskip! if @step.skipped?
```
peterdavidhamilton, line 44: This sort of interface makes me happy

peterdavidhamilton (approved):
Smashing

PR #498, opened 2021-08-26
Add Cases controller, views and presenters

ClaudioCarmeli:
## Changes in this PR

- Added `show` and `index` controller actions to serve the two templates showing support cases.
- Added model stubs and presenters to mimic what will be available to the front-end after the backend work will be merged
- Add locales across the Rails templates for better language support
- **Require Dsi library so tests can be run in isolation**

## Screenshots of UI changes

`/support/cases`

Screenshot 2021-09-01 at 17 06 13

`/support/cases/xxx`

Screenshot 2021-09-01 at 17 05 15

## Next steps

- Merge the backend changes
- Demo to stakeholders


ClaudioCarmeli:
Left to do:
- [ ] build specs
- [ ] fix code climate
- [ ] address comments RE stubs

app/controllers/support/cases_controller.rb
```diff
@@ -1,15 +1,15 @@
+# TODO: remove :nocov: and start testing
+# :nocov:
 module Support
   class CasesController < ApplicationController
-  #class CasesController < ActionController::Base
     def index
-      @cases = Support::Cases::Index.new.call
+      @cases = Support::Case.all.map { |c| CasePresenter.new(c) }
```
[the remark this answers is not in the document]
ClaudioCarmeli: I am not sure if we'll go with `Support::Cases::Index.new.call`. Maybe worth merging @AndrewCowle 's work first. I suspect signature may end up different. What do you think?

app/models/support/case.rb
```diff
@@ -0,0 +1,43 @@
...
+        status: nil,
+        enquiry_text: nil,
+        support_level: nil,
+        enquiry_source: nil,
+        participation_agreement_exists: false,
+        created_at: 7.hours.ago,
+        updated_at: 6.hours.ago,
+
+        # Associations
+        interactions: Interaction.all.map { |i| InteractionPresenter.new(i) },
```
peterdavidhamilton, line 32: The goal was to have nested presenters and not tackle the models. As this class is going to be replaced in its entirety it causes some issues.
- merge conflict with the genuine AR model, not that we can't deal with it, but it could be avoided by just stubbing in the controller
- the nested presenters are initialised here and not in the parent, this is an issue because it means we can't just pass a `Case` with all its associations and have them be decorated. See the `TaskPresenter` for an example of what we were aiming for.

spec/features/supported/cases_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  it "shows My cases tab" do
+    expect(find(".govuk-tabs__list-item--selected")).to have_text "My cases"
+  end
+
+  it "shows cases" do
+    expect(find("#my-cases .govuk-table")).to be_visible
+  end
+
+  it "shows correct columns" do
```
peterdavidhamilton, line 25: Can we get HTML elements in the lookup because it documents the structure and also get specs that document what the expectation is, so something like `it renders a table with columns for org id, category name, case status and updated timestamp`. individual checks for column headers that assert their position would be even better.

spec/presenters/supported/base_presenter_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      expect(presenter.created_at).to eq(' 1 September 2021')
+    end
+  end
+
+  describe "#updated_at" do
+    it "returns a better formatted timestamp" do
+      kase = Support::Case.find_by
+      # kase = create(:case, updated_at: '2021-12-01 00:00 BST')
+
+      presenter = Support::CasePresenter.new(kase)
```
peterdavidhamilton, line 19: We should try and leverage `let` and `subject` blocks

spec/presenters/supported/base_presenter_spec.rb
```diff
@@ -0,0 +1,23 @@
+require "rails_helper"
+
+RSpec.describe Support::BasePresenter do
+  describe "#created_at" do
+    it "returns a better formatted timestamp" do
+      kase = Support::Case.find_by
+      # kase = create(:case, created_at: '2021-12-01 00:00 BST')
+
+      presenter = Support::CasePresenter.new(kase)
```
peterdavidhamilton, line 9: This spec has just decided to test something else here

spec/presenters/supported/base_presenter_spec.rb
```diff
@@ -0,0 +1,23 @@
+require "rails_helper"
+
+RSpec.describe Support::BasePresenter do
+  describe "#created_at" do
+    it "returns a better formatted timestamp" do
+      kase = Support::Case.find_by
```
peterdavidhamilton, line 6: A `double` or `struct` with the single field could also have worked and needed no change to the spec after

spec/spec_helper.rb
```diff
@@ -23,6 +23,8 @@
 require "webmock/rspec"
 WebMock.disable_net_connect!(allow_localhost: true)
 
+require_relative "../lib/dsi/client.rb"
```
peterdavidhamilton, line 26: Can you add a comment why we felt the additional include was needed please

peterdavidhamilton:
Got some linting errors to address for consistency before it'll pass.

peterdavidhamilton:
Can we please have screenshots that capture the whole application because it helps give context and also a line to say which is which, having `/support/cases/xxx` is handy when we bring in new pages.

spec/features/supported/case_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    expect(all(".govuk-tabs__list-item").count).to eq(3)
+  end
+
+  it "shows visible tabs" do
+    all(".govuk-tabs__list-item").each do |tab|
+      expect(tab).to be_visible
+    end
+  end
+
+  it "shows School details tab" do
```
peterdavidhamilton, line 17: Is the assertion that the school details are what we see by default?

spec/features/supported/case_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    expect(find(".govuk-tabs__list-item--selected")).to have_text "School details"
+  end
+
+  it "shows School details" do
+    expect(find("#school-details .govuk-summary-list")).to be_visible
+  end
+
+  it "shows correct content" do
```
peterdavidhamilton, line 25: Is the assertion that a school contact's details are in this tab?

spec/presenters/supported/base_presenter_spec.rb
ClaudioCarmeli: Good point - I was testing as it would be run in the app (through an inheriting class) but it does make the test less clear - will address

spec/features/supported/case_spec.rb
ClaudioCarmeli: Great point - That would make it more clear actually

app/models/support/case.rb
ClaudioCarmeli: I struggled with this and now I know why. Thanks

peterdavidhamilton:
Regarding the screenshots, I appreciate you are trying to highlight what the PR has contributed but can you adopt the existing approach detailed in the PR template, omitting the browser, and not adding any extras.

peterdavidhamilton:
No need to mention the feature/unit specs in the PR because that is our definition of done but do we have sufficient coverage? Are we testing interactions with links on the page and changing content?

peterdavidhamilton:
@EdwinKruglov do you have any suggestions?

app/controllers/support/cases_controller.rb
```diff
@@ -0,0 +1,89 @@
+# TODO: remove and use ActiveRecord resources instead
+# :nocov:
+module Support
```
peterdavidhamilton, line 3: Is this even necessary, why not just put it in the controller without the psuedo lookup logic? Could've actually made the real model by now surely.

app/presenters/support/case_presenter.rb
```diff
@@ -0,0 +1,23 @@
+module Support
+  class CasePresenter < BasePresenter
+    # @return [String]
+    def state
+      super.upcase
+    end
+
+    # @return [Array<InteractionPresenter>]
+    def interactions
+      @interactions ||= super.map { |i| Support::InteractionPresenter.new(i) }
```
peterdavidhamilton, line 10: Shouldn't need the namespacing.

app/presenters/support/contact_presenter.rb
```diff
@@ -0,0 +1,8 @@
+module Support
+  class ContactPresenter < BasePresenter
+    # @return [String]
+    def full_name
+      [format_name(first_name), format_name(last_name)].join(" ")
```
peterdavidhamilton, line 5: I think string interpolated with `capitalize` would be more readable and simple, that way we can do without the `format_name` method.

app/controllers/support/cases_controller.rb
ClaudioCarmeli: Do you mean this:
```ruby
module Support
    def index
      @cases = [OpenStruct.new(...)].map { |c| CasePresenter.new(c) }
    end
    ...
```

instead of
```ruby
module Support
  class Case
    def all
      [OpenStruct.new(...)]
    end
  end
end

module Support
    def index
      @cases = Support::Case.all.map { |c| CasePresenter.new(c) }
    end
    ...
```
?

app/views/support/cases/_row.erb
```diff
@@ -0,0 +1,25 @@
...
+  <td class="govuk-table__cell"><%= kase.category.name %></td>
+
+  <td class="govuk-table__cell">
+    <strong class="govuk-tag govuk-tag--green">
+      <%= kase.state %>
+    </strong>
+  </td>
+
+  <% if extra_fields.include?(:assigned_to) %>
+    <td class="govuk-table__cell"><%= kase.case_worker_account.name %></td>
```
peterdavidhamilton, line 15: Having presenters to delegate to gives us a cleaner interface. For example `CasePresenter#agent_name` could do this?

spec/features/support/cases_spec.rb
```diff
@@ -0,0 +1,43 @@
+RSpec.feature "Supported Cases View" do
+  before { user_is_signed_in }
+
+  context "when on the index page" do
+    before { visit "/support/cases" }
+
+    it "tabs are visible" do
```
peterdavidhamilton, line 7: Would `it "has 3 tabs"` read better?

spec/features/support/cases_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      expect(all(".govuk-tabs__list-item", visible: true).count).to eq(3)
+    end
+
+    it "shows #my-cases tab as default" do
```
peterdavidhamilton, line 11: `it "defaults to the 'My Cases' tab"` maybe?

app/presenters/support/contact_presenter.rb
ClaudioCarmeli: I think you are right. Are both `first_name` and `last_name` guaranteed not to be empty or `nil`?

app/views/support/cases/_row.erb
ClaudioCarmeli: Good one, I will add it

spec/features/support/cases_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      expect(find(".govuk-tabs__list-item--selected")).to have_text "My cases"
+    end
+
+    it "shows cases" do
```
peterdavidhamilton, line 15: Does this assert that any cases are shown?

spec/features/support/cases_spec.rb
ClaudioCarmeli: Yes, it would 👍

spec/presenters/support/case_presenter_spec.rb
```diff
@@ -0,0 +1,38 @@
+require "rails_helper"
```
peterdavidhamilton, line 1: We can drop this because we have `.rspec` loading it.

spec/presenters/support/case_presenter_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  describe "#contact" do
+    it "returns a decorated contact" do
+      expect(presenter.contact.class).to eq(Support::ContactPresenter)
+    end
+  end
+
+  describe "#category" do
+    it "returns a decorated category" do
+      expect(presenter.category.class).to eq(Support::CategoryPresenter)
```
peterdavidhamilton, line 35: Better to use the `be_a` matcher

app/controllers/support/cases_controller.rb
ClaudioCarmeli: > why not just put it in the controller without the psuedo lookup logic?
I thought in this way it'd be cleaner to "cut out" when the rest will get merged. Also pseudo-lookup is handy as in controller actions there'll be nothing else to change (show or index). I believe that this cost me around 7 extra lines of code, so I thought it may be worth it (assuming I understand correctly the what your alternative would have been...see previous comment)

spec/features/support/cases_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      expect(find("#my-cases .govuk-table")).to be_visible
+    end
+
+    it "shows correct columns" do
+      expect(find("#my-cases .govuk-table__head")).to have_text "Organisation Category Status Last updated"
+    end
+  end
+
+  context "when on the show page" do
+    before { visit "/support/cases/1" }
```
peterdavidhamilton, line 25: Shouldn't we also need to be authenticated?

spec/presenters/support/contact_presenter_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "rails_helper"
+
+RSpec.describe Support::ContactPresenter do
+  let(:presenter) { described_class.new(OpenStruct(first_name: "foo", last_name: "bar")) }
```
peterdavidhamilton, line 4: If we set up the `subject` here but have a `let` for the object it receives, it means as the spec grows we can swap it out for something with different variables.

app/views/support/cases/_row.erb
```diff
@@ -0,0 +1,25 @@
...
+  <% end %>
+
+  <% if extra_fields.include?(:last_updated) %>
```
peterdavidhamilton, line 18: Can this logic be improved?

ClaudioCarmeli: it's not immediately obvious to me how - I'll think about it 💭

app/presenters/support/case_presenter.rb
ClaudioCarmeli: It gives me an error if I don't put it in - maybe something screwed up with my setup..

app/views/support/cases/_row.erb
peterdavidhamilton: Try passing in the tab, the context the row is being used.

app/presenters/support/contact_presenter.rb
peterdavidhamilton: What is the `Contact`? Is it someone that has authenticated via DSI?

app/views/support/cases/_row.erb
ClaudioCarmeli: makes sense

peterdavidhamilton:
@ClaudioCarmeli the automated test suite hasn't run because of the linting failures. `script/test` is used in the pipeline and will tell you if what you push will fail.

app/views/support/cases/_row.erb
```diff
@@ -11,15 +11,15 @@
     </strong>
   </td>
 
-  <% if extra_fields.include?(:assigned_to) %>
+  <% if %i[all_cases].include?(tab) %>
```
peterdavidhamilton, line 14: Now with the tab passsed in as the singular context it might read better if we used a case statement. Good use of presenters. Can you document those methods with a plain english description please.

peterdavidhamilton:
Is there coverage of the tabs being clicked yet?

spec/features/supported/cases_spec.rb
peterdavidhamilton: Can you do the position check?

app/presenters/support/case_presenter.rb
peterdavidhamilton: @EdwinKruglov can you give @ClaudioCarmeli a hand so we don't have to keep referring to `Support` within itself going forward. Likely needs an explicit require so it is aware the namespace exists (specs only I hope).

spec/features/support/cases_spec.rb
```diff
@@ -0,0 +1,46 @@
+RSpec.feature "Supported Cases View" do
```
peterdavidhamilton, line 1: Can we change this to __*"Case Management Dashboard"*__

spec/features/support/cases_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      # TODO: Change ".all.count" with ".count"
+      expect(all("#my-cases .govuk-table__row").count).to eq(Support::Case.all.count + 1)
+    end
+
+    it "renders a table with columns for org id, category name, case status and updated timestamp" do
+      expect(find("#my-cases .govuk-table__head")).to have_text "Organisation Category Status Last updated"
+    end
+  end
+
+  context "when on the show page" do
```
peterdavidhamilton, line 27: The thing under test here is a different page which isn't a change of context. Can you pull it into its own file please.

spec/features/support/cases_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    it "has 3 visible tabs" do
+      expect(all(".govuk-tabs__list-item", visible: true).count).to eq(3)
+    end
+
+    it "defaults to the 'School details' tab" do
+      expect(find(".govuk-tabs__list-item--selected")).to have_text "School details"
+    end
+
+    it "shows School details section" do
```
peterdavidhamilton, line 38: Can we use a `within` block if we are drilling down.

app/presenters/support/case_presenter.rb
```diff
@@ -0,0 +1,38 @@
+module Support
+  class CasePresenter < BasePresenter
+    # @return [String]
+    def state
+      super.upcase
+    end
+
+    # @return [String]
+    def agent_name
```
peterdavidhamilton, line 9: Should the `case worker agent` be being decorated by the `agent presenter`? It was in there as an example but if it isn't relevant to the entities we don't need it.

spec/features/support/cases_spec.rb
```diff
@@ -0,0 +1,46 @@
+RSpec.feature "Supported Cases View" do
+  before { user_is_signed_in }
+
+  context "when on the index page" do
+    before { visit "/support/cases" }
+
+    it "has 3 visibile tabs" do
```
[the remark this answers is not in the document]
ClaudioCarmeli: Thank you 👍 I will action

app/presenters/support/interaction_presenter.rb
```diff
@@ -0,0 +1,11 @@
+# TODO: remove :nocov: and start testing
+# :nocov:
```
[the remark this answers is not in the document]
ClaudioCarmeli: Totally forgot - thank you

app/views/support/cases/_interaction.html.erb
```diff
@@ -0,0 +1,30 @@
...
+    </dt>
+    <dd class="govuk-summary-list__value">
+      <%# interaction.case.category.name %>
+      <%= "Catering" %>
+    </dd>
+  </div>
+
+  <div class="govuk-summary-list__row">
+    <dt class="govuk-summary-list__key">
+      <%= t "support.cases.problem_description" %>
```
peterdavidhamilton, line 14: Please use `I18n.t("")` for consistency with other templates for the time being.

ClaudioCarmeli: Screenshot 2021-09-07 at 12 20 50

That did it

peterdavidhamilton:
@AndrewCowle can you have a first pass at reviewing this seeing as it is being folded into your code and I've likely made enough noise here already. Cheers

AndrewCowle (approved):

peterdavidhamilton (approved):

EdwinKruglov (approved):

PR #504, opened 2021-09-01
Use category slug as an HTML identifier (bugfix)

EdwinKruglov:
## Changes in this PR

- fix a bug where database migration would fail if the `Category.contentful_id` field had an invalid ID while running the `AddSlugsToExistingCategories` migration


EdwinKruglov:
I assumed all the `Categories` would have a valid `contentful_id`, but I forgot to consider the "design preview" category that gets created when we preview Contentful steps. Such a category has a `contentful_id` of `0`, which I found out by looking at the staging database after staging deployment failed.

AndrewCowle (approved):

ClaudioCarmeli (approved):
💯

PR #505, opened 2021-09-02
Disable validation for breaking migrations

EdwinKruglov:
## Changes in this PR

- fix a bug where migration `AddSlugsToExistingCategories` fails due to outdated data in staging by disabling validation


AndrewCowle (approved):

db/migrate/20210825142733_add_slugs_to_existing_categories.rb
```diff
@@ -5,7 +5,10 @@ def up
     categories = Category.where(slug: nil)
```
peterdavidhamilton, line 5: We could move this above the guard clause to fire it once.

db/migrate/20210825142733_add_slugs_to_existing_categories.rb
```diff
@@ -5,7 +5,10 @@ def up
...
     categories.each do |category|
       contentful_category = Content::Client.new.by_id(category.contentful_id)
-      category.update!(slug: contentful_category.slug) unless contentful_category.nil?
+      unless contentful_category.nil?
+        category.slug = contentful_category.slug
+        category.save!(validate: false)
```
peterdavidhamilton, line 10: If I've read this correctly we are skipping validations to change the `slug` because there may still be other failed validations like a missing `description`, is that correct?

spec/migrations/self-serve/add_slugs_to_existing_categories_spec.rb
```diff
@@ -37,4 +38,22 @@
...
+  context "when there is invalid existing data" do
+    before do
+      # simulate an existing invalid record
+      build(:category, description: nil, contentful_id: "contentful-category-entry", slug: nil).save!(validate: false)
+      stub_contentful_category(fixture_filename: "mfd-radio-question.json")
+    end
+
+    it "bypasses validation and populates the slug" do
+      expect(Category.count).to eq 1
+      expect(Category.first.slug).to eq nil
```
ClaudioCarmeli, line 52: This looks great. Just a question (I have no idea what the answer is) - is it possible to reference `let`'s from within `before` blocks? If so, it'd be cool to build the category in a `let` outside the `before` block, then reference it in the block and `save!` it. This would make it possible to also reference it in the test here eg
```ruby
expect(category.slug).to be_nil
```

ClaudioCarmeli (approved):
💯

db/migrate/20210825142733_add_slugs_to_existing_categories.rb
EdwinKruglov: That's right. That seems to be the error behind the failing deployment--there's a `Category` record in the staging database without a description.

PR #507, opened 2021-09-02
Updating Category via Webhook

EdwinKruglov:
## Changes in this PR

- update Contentful Webhook handling to allow existing `Categories` to be updated
- add null constraints to `description` and `slug` columns on `categories`


app/controllers/api/contentful/categories_controller.rb
```diff
@@ -4,16 +4,19 @@ class Api::Contentful::CategoriesController < Api::Contentful::BaseController
...
+      slug: contentful_category.slug,
+    },
+    unique_by: :contentful_id,
+    returning: "title")
 
-    if category
+    unless category.empty?
       render json: { status: "OK" }, status: :ok
-      Rollbar.info("Processed published webhook event for Contentful Category", category: category.title)
+      Rollbar.info("Processed published webhook event for Contentful Category", category: category.to_a[0]["title"])
```
peterdavidhamilton, line 19: An opportunity to splat all the category's fields to see what the new values are

app/controllers/api/contentful/categories_controller.rb
```diff
@@ -4,16 +4,19 @@ class Api::Contentful::CategoriesController < Api::Contentful::BaseController
...
+      description: contentful_category.description,
+      liquid_template: contentful_category.combined_specification_template,
+      contentful_id: contentful_category.id,
+      slug: contentful_category.slug,
+    },
+    unique_by: :contentful_id,
+    returning: "title")
 
-    if category
+    unless category.empty?
```
peterdavidhamilton, line 17: If we append `first` the the `upsert` this can revert, does that work?

spec/models/self-serve/category_spec.rb
```diff
@@ -4,9 +4,12 @@
   it { is_expected.to have_many(:journeys) }
 
   describe "validations" do
```
peterdavidhamilton, line 6: Not sure if this is for a later ticket, but there is no `slug` here yet.

AndrewCowle (approved):

app/controllers/api/contentful/categories_controller.rb
```diff
@@ -4,16 +4,26 @@ class Api::Contentful::CategoriesController < Api::Contentful::BaseController
...
-    end
+    category = Category.upsert({
+      title: contentful_category.title,
+      description: contentful_category.description,
+      liquid_template: contentful_category.combined_specification_template,
+      contentful_id: contentful_category.id,
+      slug: contentful_category.slug,
+    },
+    unique_by: :contentful_id,
+    returning: %w[title description contentful_id slug liquid_template])
```
peterdavidhamilton, line 15: Might be easier to do another lookup and splat that. We are by passing the validations though aren't we, so are the necessary checks present on the CMS side?

peterdavidhamilton: We could take the oppportunity to put database level checks in for `slug` and `description` etc to ensure it. We can also warn designers if the upsert fails, currently every single error from staging is pinging slack. Can you document the approach and how we are ensuring valid data and useful feedback please.

ClaudioCarmeli (approved):
Code looks great, I especially like the tests 💯

app/controllers/api/contentful/categories_controller.rb
peterdavidhamilton: Ping me if this is unclear @EdwinKruglov 👌🏼

peterdavidhamilton:
Hey @EdwinKruglov, I keep forgetting the CHANGELOG. What changes are being added here? I can see from @montychristo's PR screenshots that markdown templates are already published in `staging` so it would be good to have a plan documented in this PR explaining how updates are made and when. For the benefit of both developers and content designers. This is functionality that needs to be recorded in Confluence so maybe @quaypho can contribute? Just mindful that until we can process MD we don't want to break staging.

app/controllers/api/contentful/categories_controller.rb
```diff
@@ -4,16 +4,19 @@ class Api::Contentful::CategoriesController < Api::Contentful::BaseController
...
-    category = Category.find_or_create_by!(contentful_id: contentful_category.id) do |cat|
-      cat.title = contentful_category.title
-      cat.description = contentful_category.description
-      cat.liquid_template = contentful_category.combined_specification_template
-      cat.slug = contentful_category.slug
-    end
+    category = Category.upsert({
+      title: contentful_category.title,
+      description: contentful_category.description,
+      liquid_template: contentful_category.combined_specification_template,
```
peterdavidhamilton, line 10: My point is that right now it will change the templates in staging to markdown which we currently haven't implemented. cc @quaypho [this ticket](https://trello.com/c/4mXsUaai/343-updating-category-via-webhook) mentions the implications, so we should have a plan on the order we implement this.

peterdavidhamilton (approved):

PR #511, opened 2021-09-08
Seed Case Management Categories using Rake

peterdavidhamilton:
## Changes in this PR

- source data from YAML config
- use the format `title`, `description`, `slug` for both entities
- add default value for timestamps
- load data with a functional service call
- make data available for development using a rake task
- update descriptive class comments to use business logic

## Next steps

- complete missing `slug` and `description` values
- enable validations for both fields


montychristo (approved):

EdwinKruglov (approved):

db/schema.rb
```diff
@@ -202,18 +206,15 @@
     t.index ["support_request_id"], name: "index_support_enquiries_on_support_request_id"
   end
 
-  create_table "support_requests", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
```
peterdavidhamilton, line 205: @AndrewCowle your other PR has made its way into the schema even though it's not merged yet 👀

AndrewCowle (approved):

db/schema.rb
AndrewCowle: Weird!!!

ClaudioCarmeli (approved):

PR #513, opened 2021-09-09
Decorate Models

EdwinKruglov:
## Changes in this PR

- add missing presenters and move code to existing ones


app/controllers/steps_controller.rb
```diff
@@ -105,7 +97,11 @@ def update
 
 private
 
+  def step
+    @step ||= StepPresenter.new(Step.find(params[:id]))
+  end
+
   def parent_task
```
peterdavidhamilton, line 104: Is it possible to remove this?

app/presenters/journey_presenter.rb
```diff
@@ -0,0 +1,18 @@
+class JourneyPresenter < SimpleDelegator
+  # @return [Array<SectionPresenter>]
+  def sections
+    @sections ||= super.includes(:tasks).map do |section|
```
peterdavidhamilton, line 4: Is it possible to not call `#includes` here because it means our Presenter is aware of AR methods

app/controllers/answers_controller.rb
```diff
@@ -17,13 +17,11 @@ class AnswersController < ApplicationController
   # @see SaveAnswer
   def create
     @journey = current_journey
-    @step = Step.find(step_id)
-    @step_presenter = StepPresenter.new(@step)
 
-    @answer = AnswerFactory.new(step: @step).call
-    @answer.step = @step
+    @answer = AnswerFactory.new(step: step).call
+    @answer.step = step.__getobj__
```
peterdavidhamilton, line 22: What are we doing here?

EdwinKruglov: `step` returns a `StepPresenter` and `Answer.step` needs the actual object we're delegating to, which is what `__getobj__` returns. But now that I've looked at it closer, it doesn't look like we really need to wrap it in a `StepPresenter` in this controller as we only have two actions--`create` and `update`. It looks like the presenter was only used here to help determine the step type in a cleaner way on lines 63 and 66.

peterdavidhamilton (approved):

AndrewCowle (approved):

montychristo (approved):

PR #516, opened 2021-09-10
Add views and actions for assigning case to worker

ClaudioCarmeli:
## Changes in this PR
- Add `support_agents` table (to hydrate already existing `Support::Agent` model)
- Add functionality to the "Assign to caseworker" button
- Add template view and controller action for choosing and assigning caseworkers to cases

## Screenshots of UI changes

Screenshot 2021-09-15 at 07 35 57

## Next steps

- Add functionality for assigning a case to a worker
- Wrap logic up in a service


app/controllers/support/cases_controller.rb
```diff
@@ -5,9 +5,25 @@ def index
     end
 
     def show
-      c = Support::Case.find_by(id: params[:id])
+      @case = CasePresenter.new(case_by_id)
+    end
+
+    def edit
+      case_by_id
```
peterdavidhamilton, line 12: may be call this with a before hook?

app/views/support/cases/show.html.erb
```diff
@@ -62,7 +62,7 @@
 
 <ul class="govuk-list">
   <li>
-    <a class="govuk-link govuk-link--no-visited-state" href="#">
+    <a class="govuk-link govuk-link--no-visited-state" href="/support/cases/<%= @case.id %>/edit">
```
peterdavidhamilton, line 65: is the path helper not available?

spec/features/support/case_edit_spec.rb
```diff
@@ -0,0 +1,19 @@
+RSpec.feature "Case Management Dashboard - edit" do
+  before do
+    user_is_signed_in
+    visit "/support/cases/#{create(:support_case).id}/edit"
```
peterdavidhamilton, line 4: Can you pull the creation call out so it is separate please

spec/features/support/case_edit_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  it "renders a title" do
+    expect(find(".govuk-fieldset__heading", visible: true)).to have_text "Assign to caseworker"
```
peterdavidhamilton, line 8: I want to see the HTML element in there for all of them please

spec/features/support/case_edit_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  it "renders list of agents" do
```
peterdavidhamilton, line 11: `renders` should be used sparingly because nearly every feature spec assertion is that something has been rendered. Better to write `it "lists agents by name"` maybe?

app/views/support/cases/show.html.erb
ClaudioCarmeli: That's right, forgot to refactor that

spec/features/support/case_edit_spec.rb
ClaudioCarmeli: I thought this could look hacky, Will update 👍

spec/features/support/case_edit_spec.rb
ClaudioCarmeli: thank you

peterdavidhamilton:
Please can you add a screenshot that includes header and footer, cropping off any white space on the sides, 👍🏼

app/controllers/support/cases_controller.rb
```diff
@@ -1,13 +1,29 @@
...
     def index
       @cases = Support::Case.all.map { |c| CasePresenter.new(c) }
     end
 
     def show
-      c = Support::Case.find_by(id: params[:id])
+    end
+
+    def edit
+      @agents = Support::Agent.all.map { |a| AgentPresenter.new(a) }
```
peterdavidhamilton, line 13: If the presenter doesn't need the explicit namespace being mentioned then neither should the model. @AndrewCowle and @ClaudioCarmeli it'd be good to work this out today.

spec/features/support/case_index_spec.rb
```diff
@@ -6,22 +6,22 @@
...
+  it "displays 3 tabs" do
     expect(all("li.govuk-tabs__list-item", visible: true).count).to eq(3)
   end
 
   it "defaults to the 'My Cases' tab" do
     expect(find("#my-cases .govuk-heading-l", visible: true)).to have_text "My cases"
   end
 
-  it "renders cases" do
+  it "lists cases" do
```
peterdavidhamilton, line 17: As mentioned before you can see what rack test see's in the browser by disabling javascript. The testing of clickable tabs will require chromedriver or similar.

spec/features/support/case_index_spec.rb
```diff
@@ -6,22 +6,22 @@
...
     expect(find("#my-cases .govuk-table")).to be_visible
 
     # TODO: Change ".all.count" with ".count"
     expect(all("#my-cases .govuk-table__row").count).to eq(Support::Case.all.count + 1)
```
peterdavidhamilton, line 21: As commented here before, we shouldn't see internals referenced in feature specs. Remove `Support::Case.all.count` please.

app/controllers/support/cases_controller.rb
AndrewCowle: Does this work without specifying `Support`?

app/controllers/support/cases_controller.rb
```diff
@@ -1,13 +1,31 @@
 module Support
   class CasesController < ApplicationController
+    before_action :find_case_by_id, only: %i[show edit update]
+
     def index
-      @cases = Support::Case.all.map { |c| CasePresenter.new(c) }
+      @cases = Case.all.map { |c| CasePresenter.new(c) }
```
peterdavidhamilton, line 6: So what was caused omitting the explicit namespace to bork for you?

app/controllers/support/cases_controller.rb
```diff
@@ -1,13 +1,31 @@
 module Support
   class CasesController < ApplicationController
+    before_action :find_case_by_id, only: %i[show edit update]
```
peterdavidhamilton, line 3: Is `current_case` a better method name and pattern? It means the instance variable is clearly different in the show action, from the index, plus it is always advisable to have a method return an instance variable of the same name.

app/models/support/agent.rb
```diff
@@ -1,8 +1,7 @@
-# TODO: remove :nocov: and start testing
-# :nocov:
 module Support
   class Agent < ApplicationRecord
-    has_one :profile
+    # has_one :profile
```
peterdavidhamilton, line 3: We have a change of plan with `profiles` so just give me a shout if anything you are working on involves them.

app/presenters/support/case_presenter.rb
```diff
@@ -7,7 +7,7 @@ def state
 
     # @return [String]
     def agent_name
-      case_worker_account.full_name
+      agent&.full_name
```
peterdavidhamilton, line 10: I do love terseness but feel if you had written `if agent` it would've flagged that a possible return value is `nil`

spec/presenters/support/agent_presenter_spec.rb
```diff
@@ -1,9 +1,11 @@
 RSpec.describe Support::AgentPresenter do
-  subject(:presenter) { described_class.new(double) }
+  subject(:presenter) { described_class.new(agent) }
+
+  let(:agent) { OpenStruct.new(first_name: "foo", last_name: "bar") }
 
   describe "#full_name" do
-    it "returns correctly formatted full name" do
-      expect(presenter.full_name).to eq("Example ProcOps Agent")
+    it "returns the correctly formatted full name" do
```
peterdavidhamilton, line 7: 99% of presenter methods return strings, the `describe` also tells us the thing we are asserting. 
`it "is capitalised"` could be better.
However, a problem with the extra decorator methods I mentioned before could be with, for example with a name  like `McDonald`. I would like the additional name formatting methods dropped and `full_name` to use string interpolation.

app/models/support/case.rb
```diff
@@ -46,11 +47,6 @@ def interactions
       ]
     end
 
-    # TODO: Replace with ActiveRecord association
-    def case_worker_account
-      OpenStruct.new(full_name: "Example Agent")
-    end
-
     # TODO: Replace with ActiveRecord association
     def contact
```
peterdavidhamilton, line 51: Before you build this who/what is the contact? It'd be really useful to have comments with business logic at the same time.

spec/models/support/agent_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  it "has a first name" do
+    expect(agent.first_name).to eq("John")
+  end
+
+  it "has a last name" do
+    expect(agent.last_name).to eq("Lennon")
+  end
+
+  context "when cases are assigned to it" do
```
peterdavidhamilton, line 12: Why not  use `it { is_expected.to have_many :cases }`? Refer to the assertions in existing model specs.

ClaudioCarmeli: That's much better - thank you!

app/controllers/support/cases_controller.rb
ClaudioCarmeli: Good question. Could it be restarting spring? I changed this after a while and no errors or warnings were fired, that's good!

AndrewCowle (approved):

EdwinKruglov (approved):

PR #521, opened 2021-09-13
Create Support Enquiry Service

AndrewCowle:
## Changes in this PR

- add a `SubmitSupportRequest` service which is at the API boundary 
- the service creates a `Support::Enquiry`
- this function is triggered via an `after_create` callback 

## Next steps

- create `Case` from `Enquiry`
- when refactoring, the `SubmitSupportRequest` will make an outbound API call to the Supported application
- the Supported API inbound endpoint will take ownership of the code in the logic in the function


app/services/create_support_enquiry.rb
```diff
@@ -0,0 +1,47 @@
+# Service to create a support enquiry after_commit of a Support Request
+# This should be refactored to an API call if/when specify and support become independent applications.
+
+class CreateSupportEnquiry
```
peterdavidhamilton, line 4: Shouldn't this be `Support::CreateEnquiry`?

AndrewCowle: I felt it should live in specify as the its specify that would push it to supported (like an api call) or an after create hook on the support request... happy to move if you feel otherwise?

app/services/create_support_enquiry.rb
```diff
@@ -0,0 +1,47 @@
...
+  # @return [Support::Document]
+  def build_document(support_request)
+    return if support_request.journey.blank?
+
+    markup = get_specification_markup(support_request)
+    Support::Document.new(file_type: "HTML attachment", document_body: markup)
+  end
+
+  # @return [String]
+  def get_specification_markup(support_request)
```
peterdavidhamilton, line 39: We'd said that it'd be the HTML that got passed across in the request. This has supported querying the journeys.

app/services/create_support_enquiry.rb
peterdavidhamilton: The way we had discussed it previously involved a worker that could run in the background, but could be triggered from the view for the purposes of a swift demo. I guess what we need to be mindful of is the cross-contamination. An `Enquiry` is a Supported concern, a `SupportRequest` is a Specify concern. It feels like this should be called `SubmitSupportRequest`.

AndrewCowle: Good shout!

app/services/create_support_enquiry.rb
```diff
@@ -0,0 +1,51 @@
...
+# created that triggers the creation of a Support::Enquiry
+
+class CreateSupportEnquiry
+  # @param [support_request][SupportRequest] SupportRequest Object
+
+  def initialize(support_request)
+    @support_request = support_request
+  end
+
+  # @return [Support::Enquiry]
```
peterdavidhamilton, line 13: This becomes an outbound call to the proposed API for case management, it would be good to have yard docs show what that is expected to look like.

app/services/create_support_enquiry.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    enquiry.save!
+    enquiry
+  end
+
+private
+
+  # Creates document attachment to attach to a Support::Enquiry.
+  # @return [Support::Document]
+  def build_document(support_request)
```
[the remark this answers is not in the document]
ClaudioCarmeli: @AndrewCowle , from what I can see, we are gonna end up with some "Supported" code in "Specify" or viceversa no matter what. What are your thoughts on this?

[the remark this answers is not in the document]
AndrewCowle: @ClaudioCarmeli you're correct.

I feel that the example API document explains this well, eventually this service should just be creating a JSON payload (creating the HTML string for the specifications etc) - I can add a button into the appropriate view in `Support` to trigger a worker off to call this service. This service does need to get the HTML string from the Specification and as building the `Enquiry` here it needs to create the `Document` without an API this logic needs to happen at least on one side of the fence.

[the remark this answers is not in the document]
peterdavidhamilton: Provided we thoroughly document the shape this is meant to take at the same time. The approach I had suggested had zero code crossover other than looking into a table, just make sure your plan is clear and unambiguous.

spec/services/self-serve/submit_support_enquiry_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      it "submits a new support enquiry record" do
+        expect(result).to eq Support::Enquiry.find_by(telephone: "0151 000 0000")
+      end
+
+      it "attaches a support document" do
+        expect(result.documents.count).to eq 1
+      end
+    end
+
+    context "when no support request is given " do
```
ClaudioCarmeli, line 19: trailing space `given "`

ClaudioCarmeli (approved):
LGTM

app/services/submit_support_enquiry.rb
```diff
@@ -0,0 +1,77 @@
...
+#       "name":"Joe Bloggs",
+#       "email":"example@exmaple.com",
+#       "telephone": "0151 000 0000"
+#       "message": "example message",
+#       "documents" : [ { "file_type": "html_markup", "document_body": "<h1>example html markup</h1>"}]
+#     }]
+#   }
+#   ```
+
+class SubmitSupportEnquiry
```
peterdavidhamilton, line 32: `SubmitSupportRequest` 
- "Enquiry" is a Supported concern
- "SupportRequest" is a Specify concern

peterdavidhamilton:
Updated your PR description. Are you also adding the callback?

peterdavidhamilton (approved):
Updated PR description to:
> this function is triggered via an after_create callback

PR #522, opened 2021-09-14
Markdown for liquid templates

EdwinKruglov:
## Changes in this PR

- use Markdown to render specification templates
- use `pandoc` to convert Markdown into `docx`

#### HTML

image

#### DOCX

image

#### HTML

image

#### DOCX

image

#### PDF

image


peterdavidhamilton:
So the way I pictured it is with two classes:

1. liquid template logic, replacing variables and filling answers, receiving MD and returning MD
2. document formatting, with a format param, that receives MD and can return HTML, or PDF, or DOCX, or ODT...

I use pandoc in my own projects. Can you have a play with a functional style of Ruby leveraging the dry-rb libs? :)

app/services/specification_renderer.rb
```diff
@@ -1,23 +1,42 @@
...
+  end
+
+private
+
+  # Remove HTML comments (<!-- -->)
+  #
+  # These will occassionally not be ignored by the parser and must be removed here
+  #
+  # @param [String]
+  def remove_html_comments(content)
```
peterdavidhamilton, line 34: pandoc has its own comments that can be stripped automatically I hope. @quaypho I shared some with you in the example files, essentially the are HTML comments with an extra hyphen. @EdwinKruglov if the templates from Contentful use those then they will be in our markdown but not in our HTML or docs. 🤞🏼

app/services/document_formatter.rb
```diff
@@ -0,0 +1,25 @@
...
+  extend Dry::Initializer
+
+  option :markdown, Types::String
+  option :format, Types::Symbol, default: proc { :docx }
+
+  # Return the formatted document
+  #
+  # @param [Boolean]
+  #
+  def call(journey_complete:)
```
peterdavidhamilton, line 17: Could do with the param name in the doc but I'd suggest `draft`, perhaps with a default of `true`.

app/services/document_formatter.rb
```diff
@@ -0,0 +1,25 @@
...
+require "types"
+require "pandoc-ruby"
+
+# Convert Markdown into a document format
+#
+class DocumentFormatter
+  extend Dry::Initializer
+
+  option :markdown, Types::String
+  option :format, Types::Symbol, default: proc { :docx }
```
peterdavidhamilton, line 11: Check out custom enumerable types, where we can validate the symbols passed in 👍🏼

app/services/document_formatter.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  option :markdown, Types::String
+  option :format, Formats, default: proc { :docx }
+
+  # Return the formatted document
+  #
+  # @param draft [Boolean]
+  #
+  def call(draft: true)
+    markdown.prepend(I18n.t("journey.specification.download.warning.incomplete")) if draft
```
peterdavidhamilton, line 20: Experience tells me it is wise to pass in a string for the prepended message rather than have the class be aware of I18n values.

app/services/document_formatter.rb
```diff
@@ -0,0 +1,24 @@
...
+  extend Dry::Initializer
+
+  Formats = Types::Symbol.enum(:docx)
+
+  option :markdown, Types::String
+  option :format, Formats, default: proc { :docx }
+
+  # Return the formatted document
+  #
+  # @param draft [Boolean]
```
peterdavidhamilton, line 17: Is this where we document how we are adding an `WIP` notice now?

app/services/specification_renderer.rb
```diff
@@ -1,23 +1,26 @@
-# Parse Liquid templates and render HTML
+require "dry-initializer"
+require "types"
+require "pandoc-ruby"
+
+# Parse Liquid templates and render Markdown
 #
 class SpecificationRenderer
```
peterdavidhamilton, line 7: I know this is a WIP, so forgive me if I'm jumping the gun, but do we have two classes stepping on each others toes currently? Ideally we'd have:
- class 1 parses liquid
- class 2 turns MD into anything want
We also need to watch out if we are calling `html_safe` and who fires that method. The goal is to let pandoc be solely responsible for turning MD into DOCX. 
It'd be great to be using `Pandoc` and `Liquid::Template` in one place and not have it spill into presenters, controllers etc.

spec/features/self-serve/journeys/download_spec.rb
```diff
@@ -42,12 +36,7 @@
       click_view
 
       warning_html = I18n.t("journey.specification.download.warning.incomplete")
```
peterdavidhamilton, line 38: Does it get messy if we have pure strings in this spec?

spec/services/self-serve/document_formatter_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        .and_call_original
+
+        formatter.call(draft: false)
+      end
+    end
+
+    context "when the journey is incomplete" do
+      it "converts Markdown into DOCX with added warning" do
+        md = "# Title"
+        modified_md = "<article id='warning'><p></b>You have not completed all the tasks in Create a specification. There may be information missing from your specification.</b></p></article># Title"
```
peterdavidhamilton, line 20: Is it possible to remove all HTML from this workflow? Does it make sense why?

peterdavidhamilton (approved):
- use your class in `StepsPresenter`
- add `@example` doc for inputs and outputs
- include screenshots of HTML and document appearance as before/after
- `Types::Strict::String` and any other validation hardening 
- check what happens with an explicit call `to_s`

app/presenters/step_presenter.rb
```diff
@@ -55,6 +57,6 @@ def response
   #
   # @return [String]
   def render_markdown(text)
```
peterdavidhamilton, line 59: Just an afterthought, could you change the method name to `convert_markdown` please. 👍🏼

app/services/specification_renderer.rb
```diff
@@ -1,23 +1,27 @@
...
+  option :draft_msg, Types::String, default: proc { I18n.t("journey.specification.warning") }
+
+  # @param draft [Boolean] - if true, prepends `draft_msg` to the spec
   #
-  def initialize(template:, answers:)
-    @template = Liquid::Template.parse(template, error_mode: :strict)
-    @answers = answers
-  end
+  # return [String]
+  def call(draft: true)
```
peterdavidhamilton, line 17: Do you think it makes more sense to move the completeness check in here because this object now has the whole `journey` to talk to so it can determine if it is a draft? 

Which cleans that logic from the controller, and the draft param would be an optional override.

config/locales/en.yml
```diff
@@ -108,12 +108,10 @@ en:
   journey:
     specification:
       header: Your specification
-      warning: You have not completed all the tasks. There may be information missing from your specification.
+      warning: You have not completed all the tasks in Create a specification. There may be information missing from your specification.
```
peterdavidhamilton, line 111: @quaypho might want to make a change here, perhaps:

`"You have not completed all the tasks. Your specification is incomplete"` 

@EdwinKruglov can we also change the key to `draft` to reflect its purpose.

peterdavidhamilton:
Looks fab. There is a bit that says "Markdown: Edits to comments" in the screenshots, is that a pandoc comment slipping through?

EdwinKruglov:
> Looks fab. There is a bit that says "Markdown: Edits to comments" in the screenshots, is that a pandoc comment slipping through?

That's text in the actual template in `develop`. I'm sure it will be removed in `staging`.

app/services/liquid_parser.rb
```diff
@@ -6,11 +6,21 @@
...
+  #     "answer_1UI0NhH8gu4FKVOKBWh4pI" => { "response" => "Yes", "further_information" => nil } }
+  option :answers, Types::Strict::Hash
 
   # Fill in answers and render specification
   #
+  # @example
+  #   - The cost of the service: 50
+  #
   # @return [String]
   def render
```
peterdavidhamilton, line 25: Another thought would be if you change this to `#call` it'll behave like a `proc/lambda` so is consistent with our other services

app/services/specification_renderer.rb
```diff
@@ -1,23 +1,29 @@
...
-  end
+  # return [String]
+  def call(draft: nil)
+    is_draft = draft.nil? ? !journey.all_tasks_completed? : draft
 
-  def to_html
-    @template.render(@answers).html_safe
-  end
+    template = LiquidParser.new(
+      template: journey&.category&.liquid_template,
```
peterdavidhamilton, line 21: Sorry @EdwinKruglov forgot this one, because we now have type checking on params, we can be confident that we have a template and steps, so we should't need `&` anymore. 🤞🏼

EdwinKruglov:
@peterdavidhamilton here's what `pandoc` comments end up looking like in HTML:
image

PR #523, opened 2021-09-14
Create Support Case Service

AndrewCowle:
## Changes in this PR

- Added service to create `Support::Case` from `Support::Enquiry` - the service changes the polymorphic association of any documents on the `Support::Enquiry` to `Support::Case` 

## Next steps

Liase with @ClaudioCarmeli  to work out when to put a temporary button for the process for demos.


app/services/support/create_support_case.rb
```diff
@@ -0,0 +1,30 @@
+module Support
+  #
+  # Service to create SupportCases from SupportEnquiry
```
peterdavidhamilton, line 3: Only because it could get confusing if we don't - can this say "Open a `Case` from an `Enquiry`" as this omits the namespace and uses phrasing more akin to the business. We don't want the same namespace duplication in the docs as well as the code.

app/services/support/create_support_case.rb
```diff
@@ -0,0 +1,30 @@
...
+  #
+  class CreateSupportCase
```
peterdavidhamilton, line 5: Is `CreateCase` acceptable?

app/services/support/create_support_case.rb
```diff
@@ -0,0 +1,30 @@
...
+    # @param [support_enquiry][SupportEnquiry] SupportEnquiry Object
+
+    def initialize(support_enquiry)
+      @support_enquiry = support_enquiry
```
peterdavidhamilton, line 9: Can we avoid the `support` prefix unless it really helps please, like with that horrid `kase` issue

app/models/support/enquiry.rb
```diff
@@ -10,5 +10,6 @@ class Enquiry < ApplicationRecord
     include Support::Documentable
```
peterdavidhamilton, line 10: I'm curious to know if the explicit namespace is required to keep AR happy?

app/services/support/create_case.rb
```diff
@@ -0,0 +1,30 @@
+module Support
+  #
+  # Service to open Case from Enquiry
+  #
+  class CreateCase
+    # @param [enquiry][SupportEnquiry] SupportEnquiry Object
```
peterdavidhamilton, line 6: Yard might error here. Build the docs every so often and see how they read in the HTML to be sure
Try:
`@param enquiry [Enquiry] incoming request for support`

app/services/support/create_case.rb
```diff
@@ -0,0 +1,30 @@
...
+      kase = @enquiry.build_case
+      attach_documents
+      kase.save!
+      kase
+    end
+
+  private
+
+    # Change the association of the documents from SupportEnquiry to SupportCase (polymorphic)
+    # @return [Array][Support::Documents]
```
peterdavidhamilton, line 23: `@return [Array<Documents>]`

db/schema.rb
```diff
@@ -172,8 +172,8 @@
 
   create_table "support_categories", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
     t.string "title", null: false
-    t.datetime "created_at", precision: 6, default: -> { "CURRENT_TIMESTAMP" }, null: false
-    t.datetime "updated_at", precision: 6, default: -> { "CURRENT_TIMESTAMP" }, null: false
+    t.datetime "created_at", precision: 6, null: false
```
peterdavidhamilton, line 175: Intentional?

peterdavidhamilton:
Your mention of the demo button in the PR description is not where I had expected. 
If you adopt a background service to periodically move data from an outboard request to an inbound enquiry, the browser button fires this for a smooth demo experience. It is an approach that keeps code separate. Are you referring to a different button that opens a case from an enquiry?

AndrewCowle:
> Your mention of the demo button in the PR description is not where I had expected.
> If you adopt a background service to periodically move data from an outboard request to an inbound enquiry, the browser button fires this for a smooth demo experience. It is an approach that keeps code separate. Are you referring to a different button that opens a case from an enquiry?

Yes I was to the opening cases from enquiries - added to the backlog (with a mockup) https://trello.com/c/7GDRIhAb/189-support-enquiries-list-view

ClaudioCarmeli (approved):
LGTM

db/schema.rb
AndrewCowle: No I think my migrations might have gone a bit wonky

PR #526, opened 2021-09-15
Feature/571 add buyer agent profiles

montychristo:


peterdavidhamilton:
@montychristo we need to catch up ASAP because the image I shared and its relevance to this ticket have not been understood.

peterdavidhamilton:
@montychristo I have added the following to the test DSI config for you:

- `https://localhost:3000/support/auth/dfe/callback`
- `https://localhost:3000/support/auth/dfe/signout`

PR #530, opened 2021-09-16
Abstract Update Case logic into separate service

ClaudioCarmeli:
## Changes in this PR

- Case Update logic was removed from the controller action and placed in [a self-contained testable unit](https://github.com/DFE-Digital/buy-for-your-school/pull/530/files#diff-d7bdf92c88afa3f7c447a24240ca7aac6c7b56132e0e8d92fd28b80b047adefdR5) (`Support::UpdateCase`)
- Redirect user to case page, after successful agent assignment
- [Added spec helper](https://github.com/DFE-Digital/buy-for-your-school/pull/530/files#diff-d35fdbf1d42c668f8f4c92df27b8d3f9151e36132faade568aff709c45044109R77) (following pattern already present elsewhere in the test suite)

### DIFF
--- a/app/controllers/support/cases_controller.rb
+++ b/app/controllers/support/cases_controller.rb
@@ -1,6 +1,7 @@
 module Support
   class CasesController < ApplicationController
     before_action :current_case, only: %i[show edit update]
+    before_action :agent, only: %i[update]
 
     def index
       @cases = Case.all.map { |c| CasePresenter.new(c) }
@@ -13,12 +14,7 @@ module Support
     end
 
     def update
-      # TODO: remove :nocov: and start testing
-      # :nocov:
-      @current_case.agent = Agent.find_by(id: params.dig(:support_case, :agent))
-      @current_case.state = "open"
-      @current_case.save!
-      # :nocov:
+      UpdateCase.new(current_case, agent).call
 
       redirect_to support_cases_path
     end
@@ -27,6 +23,10 @@ module Support
 
     def current_case
       @current_case ||= CasePresenter.new(Case.find_by(id: params[:id]))
+    end
+
+    def agent
+      @agent ||= Agent.find_by(id: params.dig(:support_case, :agent))
     end
   end
 end
--- a/app/controllers/support/cases_controller.rb
+++ b/app/controllers/support/cases_controller.rb
@@ -16,7 +16,7 @@ module Support
     def update
       UpdateCase.new(current_case, agent).call
 
-      redirect_to support_cases_path
+      redirect_to support_case_path(anchor: "case-history")
     end
 
   private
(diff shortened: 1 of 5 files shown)


peterdavidhamilton:
The definition of done would include updating the browser coverage in `/spec/features/support/case_edit_spec.rb`.
It might be desirable therefore to also redirect to the updated case.

app/controllers/support/cases_controller.rb
```diff
@@ -13,20 +14,19 @@ def edit
...
-      # :nocov:
-      @current_case.agent = Agent.find_by(id: params.dig(:support_case, :agent))
-      @current_case.state = "open"
-      @current_case.save!
-      # :nocov:
-
-      redirect_to support_cases_path
+      UpdateCase.new(current_case, agent).call
+
+      redirect_to support_case_path(anchor: "case-history")
```
AndrewCowle, line 19: Are you not displaying any confirmation message advising user something was updated?

AndrewCowle (dismissed):

montychristo (approved):
LGTM

app/controllers/support/cases_controller.rb
ClaudioCarmeli: No, but that's a good idea. It's not in the A/C's so I thought we May want to add it in a separate PR / Jira ticket. Happy to add if it others are happy with that 👍

EdwinKruglov (approved):

spec/features/support/case_edit_spec.rb
```diff
@@ -22,4 +22,12 @@
   it "displays the submit button" do
     expect(find_button("Assign")).to be_present
   end
+
+  context "when assigning agent to case" do
+    before { find_button("Assign").click }
+
+    it "is redirected to the case in question" do
```
peterdavidhamilton, line 29: Is this something that is seen in the browser? If so, can we get a minimal smoke test to assert that the user sees the change?

ClaudioCarmeli: Done 🎉  https://github.com/DFE-Digital/buy-for-your-school/pull/530/commits/ac3a23e0dffe10491bd94f0c44efe8a93f482539

AndrewCowle (approved):

spec/features/support/case_edit_spec.rb
```diff
@@ -24,10 +24,25 @@
   end
 
   context "when assigning agent to case" do
-    before { find_button("Assign").click }
+    before do
+      find("label.govuk-radios__label", text: "John Lennon").click
+      find_button("Assign").click
+    end
 
     it "is redirected to the case in question" do
```
peterdavidhamilton, line 32: Could do with using the same kind of language in our specs like `redirects to the case`

spec/features/support/case_edit_spec.rb
```diff
@@ -24,10 +24,25 @@
...
       expect(page).to have_a_support_case_path
     end
+
+    context "when returning to the main page" do
```
peterdavidhamilton, line 36: `when on the case management dashboard`

peterdavidhamilton (approved):

PR #531, opened 2021-09-16
Reflect DSI profile changes

peterdavidhamilton:
## Changes in this PR

- users who have previously used the service will have their names added
- changes to names within DSI will be propagated to the `users` table
- entries in the table are reported to via Rollbar when created and updated
- NB: profile changes in the test DSI environment may be slow to take effect

### DIFF
--- a/spec/features/self-serve/auth/sign_in_spec.rb
+++ b/spec/features/self-serve/auth/sign_in_spec.rb
@@ -17,6 +17,38 @@ RSpec.feature "DfE Sign-in" do
       click_continue
     end
 
+    context "but their DSI details have changed" do
+      let!(:user) { create(:user) }
+
+      let(:updated_user) do
+        build(:user,
+              dfe_sign_in_uid: user.dfe_sign_in_uid,
+              email: user.email,
+              first_name: "first_name changed in DSI",
+              last_name: "last_name changed in DSI")
+      end
+
+      before do
+        # force signout
+        visit "/auth/failure"
+        # change DSI details
+        user_exists_in_dfe_sign_in(user: updated_user)
+        # DfE sign in
+        click_start
+        # check new details
+        # FIXME: assert changes are visible on /profile
+        # visit "/users"
+        user.reload
+      end
+
+      it "updates the user's names" do
+        expect(user.first_name).to eq "first_name changed in DSI"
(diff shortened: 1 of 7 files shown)


spec/features/self-serve/auth/sign_in_spec.rb
```diff
@@ -17,6 +17,38 @@
...
+        # FIXME: assert changes are visible on /profile
+        # visit "/users"
+        user.reload
+      end
+
+      it "updates the user's names" do
+        expect(user.first_name).to eq "first_name changed in DSI"
+        expect(user.last_name).to eq "last_name changed in DSI"
+
+        # expect(find("dd.govuk-summary-list__value")[0]).to have_text "first_name changed in DSI last_name changed in DSI"
```
ClaudioCarmeli, line 48: is this part of
```ruby
# FIXME: assert changes are visible on /profile
```
?

spec/services/self-serve/create_user_spec.rb
```diff
@@ -8,30 +8,75 @@
...
+      end
     end
 
     context "when a user with that DSI UUID does not exist in the database" do
       let(:dfe_sign_in_uid) { "an-unknown-uuid" }
+      let(:email) { "unknown@example.com" }
 
       it "creates a new user record" do
         expect(result).to eq User.find_by(dfe_sign_in_uid: "an-unknown-uuid")
+        # expect(result).to eq User.last
```
ClaudioCarmeli, line 48: Is it intentional?

ClaudioCarmeli (approved):
A couple of questions

spec/services/self-serve/create_user_spec.rb
peterdavidhamilton: Yes, was undecided which to use because the test needs a context for multiple users. I'm coming back to this in another ticket I have on the go.

spec/features/self-serve/auth/sign_in_spec.rb
peterdavidhamilton: It is, the page exists in another branch, this is a note for me to assert the change in the browser not just the persistence layer.

montychristo (approved):

EdwinKruglov (approved):

PR #532, opened 2021-09-16
Add informative comment delineating class purpose

ClaudioCarmeli:
## Changes in this PR

Added an informative comment delineating the purpose of the `Agent` class and what it means in Business terms to minimise confusion.

### DIFF
--- a/app/models/support/agent.rb
+++ b/app/models/support/agent.rb
@@ -1,4 +1,9 @@
+# frozen_string_literal: true
+
 module Support
+  #
+  # A public servant who picks up and handles "support cases"
+  #
   class Agent < ApplicationRecord
     # has_one :profile
 
--- a/app/models/support/agent.rb
+++ b/app/models/support/agent.rb
@@ -2,7 +2,8 @@
 
 module Support
   #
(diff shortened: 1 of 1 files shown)


app/models/support/agent.rb
```diff
@@ -1,4 +1,9 @@
+# frozen_string_literal: true
+
 module Support
+  #
+  # A public servant who picks up and handles "support cases"
```
peterdavidhamilton, line 5: Was thinking it would include all the other names people might use in meetings.

peterdavidhamilton:
You don't have to worry about adding this so I have removed it from your PR description. The template includes reminders of our definition of done btw.

```
## Screenshots of UI changes

`N/A` Nothing visual

## Next steps

`N/A` Merge PR
```

app/models/support/agent.rb
ClaudioCarmeli: Thank you @peterdavidhamilton - Updated now. What do you think?

AndrewCowle (approved):

peterdavidhamilton (approved):

PR #538, opened 2021-09-22
Create restricted entry point for case management agents

peterdavidhamilton:
## Changes in this PR

- access to Supported first requires authenticating via Specify
- add a login page for caseworkers
- restrict access to case management via uid/email
- populate agents table using data from users table
- strip unnecessary code
- rephrase some specs
- add missing frozen string literal
- encapsulate caseworker authentication in an Rspec context
- make caseworker role filtering work using affiliation to an organisation 

## Screenshots of UI changes

`/support`
Screenshot 2021-09-22 at 11 46 42

## Next steps

- add SBP filtering at authentication using establishment type/category
- register uses with the proc ops team


app/controllers/support/agents_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+      #
+      if ::PagesController.dsi?
+        specify_user.update!(
+          email: session[:dfe_sign_in_uid],
+          first_name: session[:dfe_sign_in_uid],
+          last_name: session[:dfe_sign_in_uid],
+        )
+      end
+
+      if AGENT_UUIDS.include?(specify_user.dfe_sign_in_uid)
```
montychristo, line 20: Do we also need a check to ensure we can login via DSI using the `AGENT_EMAILS` whitelist too?

config/initializers/case_workers.rb
```diff
@@ -0,0 +1,11 @@
+# AGENT_EMAILS = %w[
+#   agent1@example.com
+#   agent2@example.com
+# ]
+
+# use in development when bypassing DSI
+#
+AGENT_UUIDS = %w[
```
montychristo, line 8: Should we make these environment variables?

config/routes.rb
```diff
@@ -58,9 +58,10 @@
   #
   # Supported ------------------------------------------------------------------
   #
-  namespace :support do
-    get "admin", to: "admin#show"
+  get "support", to: "support/pages#start_page", as: :support_root
```
montychristo, line 61: We could use `root_path` inside the support namespace. Only a personal preference though

montychristo (dismissed):
Couple of comments but mainly LGTM - nicely done! Also - a few code climate violations

app/controllers/support/agents_controller.rb
peterdavidhamilton: At the moment this code is bypassing DSI in development so we can change UIDs to filter on.
If we deployed that approach, you're right we'd want to instead be filtering emails, but it's not the only option and it wouldn't be sustainable.

The way we are going to do it now is to assign agents to a particular organisation; that way access control is in the hands of ProcOps and we don't need developer input going forward.

This PR needs the org filtering to be added before it's ready to merge.

config/initializers/case_workers.rb
peterdavidhamilton: This file is going away once an organisation exists.

config/routes.rb
peterdavidhamilton: I did do that originally and now I can't remember what prompted me to change it? Think there was a little gotcha and doing it this way made it more workable.

spec/features/support/case_agent_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  it "displays the submit button" do
+    expect(find_button("Assign")).to be_present
+  end
+
+  context "when assigning an agent to a case" do
+    before do
+      find("label.govuk-radios__label", text: "Procurement Specialist").click
+      find_button("Assign").click
```
peterdavidhamilton, line 26: @ClaudioCarmeli this is me renaming your `case_edit_spec` to focus on one specific editing feature, i.e. the assignment of a caseworker. I'm going to change this line to `click_button`, just drawing your attention to that method and this upcoming change.

EdwinKruglov (approved):

app/presenters/support/base_presenter.rb
```diff
@@ -1,20 +1,22 @@
+# frozen_string_literal: true
+
 module Support
   class BasePresenter < SimpleDelegator
     # @return [String]
     def created_at
-      format_date(super)
+      super.strftime(date_format)
```
peterdavidhamilton, line 7: @EdwinKruglov for your PR this is better

app/presenters/support/base_presenter.rb
```diff
@@ -1,20 +1,22 @@
...
     end
 
   private
 
     # @return [String]
-    def format_date(date)
-      date.strftime("%e %B %Y")
+    def date_format
+      "%e %B %Y"
```
AndrewCowle, line 19: date.strftime("%e %B %Y at %H:%M")

I changed this in my PR so we show times alongside (spoke to Adam and Robin, was some variance in the PT and we felt having time makes more sense) the change will go in with my PR, just wanted to raise it.

config/routes.rb
AndrewCowle: happy with support_root - need to look at the namescape vs scope in route so it's better unique name

AndrewCowle (approved):

app/presenters/support/agent_presenter.rb
```diff
@@ -1,8 +1,21 @@
...
     def full_name
       "#{first_name} #{last_name}"
     end
+
+    # @return [false]
+    def guest?
+      false
+    end
+
+    # TODO: use this in auth checks
```
Claudio-And, line 15: Question: why will `.agent` be needed in those check?

Claudio-And (approved):

EdwinKruglov (approved):

app/presenters/support/base_presenter.rb
peterdavidhamilton: good stuff if we have a few types we can put them in a hash and call them by name or something

spec/features/support/case_agent_spec.rb
peterdavidhamilton: I did not make this change as promised @Claudio-And could you do that please?

PR #547, opened 2021-09-28
Fix specification appearance

EdwinKruglov:
## Changes in this PR

- add `govuk` styling to unordered lists
- have Pandoc strip comments when converting specifications
- other changes made in Contentful to fix Markdown

### Before

image

### After

image

### DIFF
--- a/app/assets/stylesheets/components/_specification.scss
+++ b/app/assets/stylesheets/components/_specification.scss
@@ -16,6 +16,11 @@
     @extend .govuk-list--number;
   }
 
+  ul {
+    @extend .govuk-list;
+    @extend .govuk-list--bullet;
+  }
+
   table {
     @extend .govuk-table !optional;
   }
--- a/app/assets/stylesheets/components/_specification.scss
+++ b/app/assets/stylesheets/components/_specification.scss
@@ -1,3 +1,8 @@
+/**
+* Specification templates have no classes of their own so we're overriding the default styling to
+* keep the look consistent.
+**/
+
 #specification, .specification {
   h2 {
     @extend .govuk-heading-m;
--- a/app/presenters/long_text_answer_presenter.rb
+++ b/app/presenters/long_text_answer_presenter.rb
@@ -1,11 +1,4 @@
 class LongTextAnswerPresenter < SimpleDelegator
-  include ActionView::Helpers::TextHelper
-
-  # @return [String]
-  def response
-    simple_format(super)
-  end
-
   # @return [Hash]
   def to_param
     {
--- a/app/services/document_formatter.rb
+++ b/app/services/document_formatter.rb
@@ -1,6 +1,7 @@
 require "dry-initializer"
 require "types"
 require "pandoc-ruby"
+require "pry"
 
 # Document converter
 #
@@ -18,6 +19,7 @@ class DocumentFormatter
   #
   # @return [String]
   def call
+    # binding.pry
     PandocRuby.convert(content, "--strip-comments", from: from, to: to).to_s
   end
 end
--- a/app/services/document_formatter.rb
+++ b/app/services/document_formatter.rb
@@ -1,7 +1,6 @@
 require "dry-initializer"
 require "types"
 require "pandoc-ruby"
-require "pry"
 
 # Document converter
 #
@@ -19,7 +18,6 @@ class DocumentFormatter
   #
   # @return [String]
   def call
-    # binding.pry
     PandocRuby.convert(content, "--strip-comments", from: from, to: to).to_s
   end
 end
--- a/app/services/document_formatter.rb
+++ b/app/services/document_formatter.rb
@@ -18,6 +18,6 @@ class DocumentFormatter
   #
   # @return [String]
   def call
-    PandocRuby.convert(content, from: from, to: to).to_s
+    PandocRuby.convert(content, "--strip-comments", from: from, to: to).to_s
   end
 end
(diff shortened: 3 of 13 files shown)


peterdavidhamilton:
bitmoji

spec/features/self-serve/planning/start_page_spec.rb
```diff
@@ -1,9 +1,6 @@
 RSpec.feature "Users can see a start page for planning their purchase" do
-  # TODO: reinstate spec once new content is finalised
-  xscenario "Start page content is shown on the root path" do
-    visit root_path
-
-    click_on("procuring a new catering service for a school")
+  scenario "Start page content is shown on the root path" do
+    visit "/planning"
 
     expect(page).to have_content("Catering services")
```
peterdavidhamilton, line 5: @EdwinKruglov can we assert the tag and class unless @PiperS52 is already doing that elsewhere

spec/features/self-serve/steps/presenter_spec.rb
```diff
@@ -47,7 +47,7 @@
     fill_in "answer[response]", with: "\r\n\r\nfoo\r\n\r\n"
     click_on "Continue"
     click_on "Back"
-    expect(page.html).to include "<p></p>\n\n<p>foo</p>"
+    expect(page).to have_text "\r\r \r\r foo\r\r \r\r"
```
peterdavidhamilton, line 50: Same here, where exactly does this appear on the page. Could use a `within` block or `find`.

spec/presenters/self-serve/long_text_answer_presenter_spec.rb
```diff
@@ -5,14 +5,14 @@
     it "returns the response in HTML using the Rails simple_format method" do
       step = build(:long_text_answer, response: "Lots of text")
       presenter = described_class.new(step)
-      expect(presenter.response).to eq("<p>Lots of text</p>")
+      expect(presenter.response).to eq "Lots of text"
```
peterdavidhamilton, line 8: Whilst we are in this spec can we pull the common setup to the top and delete the `require` at the top please

spec/presenters/self-serve/long_text_answer_presenter_spec.rb
```diff
@@ -5,14 +5,14 @@
     it "returns the response in HTML using the Rails simple_format method" do
```
peterdavidhamilton, line 5: This is no longer what it does or how it does it

spec/presenters/self-serve/long_text_answer_presenter_spec.rb
```diff
@@ -5,14 +5,14 @@
...
       step = build(:long_text_answer, response: "Lots of text")
       presenter = described_class.new(step)
-      expect(presenter.response).to eq("<p>Lots of text</p>")
+      expect(presenter.response).to eq "Lots of text"
     end
 
     context "when the text includes line breaks" do
       it "returns 2 HTML paragraphs" do
```
peterdavidhamilton, line 12: This spec info also needs changing

app/presenters/long_text_answer_presenter.rb
```diff
@@ -1,11 +1,4 @@
 class LongTextAnswerPresenter < SimpleDelegator
-  include ActionView::Helpers::TextHelper
```
peterdavidhamilton, line 2: This makes me happy

app/services/document_formatter.rb
```diff
@@ -18,6 +18,6 @@ class DocumentFormatter
   #
   # @return [String]
   def call
-    PandocRuby.convert(content, from: from, to: to).to_s
+    PandocRuby.convert(content, "--strip-comments", from: from, to: to).to_s
```
peterdavidhamilton, line 21: I little addition to the notes that shows the type of comment that is being stripped and why would be useful

peterdavidhamilton: We might have touched on this before but what does it return without the explicit `to_s`?

app/assets/stylesheets/components/_specification.scss
```diff
@@ -16,6 +16,11 @@
     @extend .govuk-list--number;
   }
 
+  ul {
```
peterdavidhamilton, line 19: Please add a comment at the top explaining that our template has no classes so we are overriding the defaults for elements to make the spec doc look consistent

peterdavidhamilton (approved):

app/services/document_formatter.rb
EdwinKruglov: It seems to return a String normally.

app/services/document_formatter.rb
```diff
@@ -14,7 +14,8 @@ class DocumentFormatter
   option :from, ReaderFormats
   option :to, WriterFormats
 
-  # Return the converted document
+  # Return the converted document.
+  # HTML comments are stripped out.
```
peterdavidhamilton, line 18: do these need to be `<!---` (triple dash)? what format comment will be stripped and what is let through if any?

EdwinKruglov: Regular HTML comments `<!--` and triple-dash comments `<!---` are stripped out. The reference style link ones from here: https://www.jamestharpe.com/markdown-comments/ are ignored as expected, too.

app/services/document_formatter.rb
peterdavidhamilton: If that's the case we can remove the redundant `to_s`

PR #550, opened 2021-09-30
removed I18n .t instances in spec and added relevant tags

PiperS52:
## Changes in this PR

- removed I18n .t instances in spec and replaced with literal strings
- added css class tags

### DIFF
--- a/spec/features/self-serve/journeys/view_spec.rb
+++ b/spec/features/self-serve/journeys/view_spec.rb
@@ -11,8 +11,9 @@ RSpec.feature "Users can see their specification" do
 
     expect(page).to have_breadcrumbs ["Dashboard", "Create specification", "View specification"]
 
-    # TODO: remove I18n.t from specs
-    expect(page).to have_content(I18n.t("journey.specification.header"))
+    # journey.specification.header
+    expect(page).to have_text "Your specification"
+    expect(find("h1.govuk-heading-xl")).to have_text "Your specification"
 
     within("article#specification") do
       expect(page).to have_content("Menus and ordering")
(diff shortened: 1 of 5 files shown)


spec/helpers/self-serve/application_helper_spec.rb
```diff
@@ -1,21 +1,23 @@
...
+      # banner.beta.tag
+      expect(helper.banner_tag).to eq "beta"
     end
   end
 
   describe "#banner_message" do
     it "returns the beta message by default" do
-      expect(helper.banner_message).to eq I18n.t("banner.beta.message", support_email: ENV.fetch("SUPPORT_EMAIL"))
+      # banner.beta.message
+      expect(helper.banner_message).to eq("This is a new service – your <a href="mailto:%{support_email}" class="govuk-link">feedback</a> will help us to improve it.", support_email: ENV.fetch("SUPPORT_EMAIL"))
```
peterdavidhamilton, line 13: For this one, you'll need to do the string interpolation so that the ENV is mixed in.

spec/features/self-serve/journeys/view_spec.rb
```diff
@@ -11,8 +11,9 @@
 
     expect(page).to have_breadcrumbs ["Dashboard", "Create specification", "View specification"]
 
-    # TODO: remove I18n.t from specs
-    expect(page).to have_content(I18n.t("journey.specification.header"))
+    # journey.specification.header
+    expect(page).to have_text "Your specification"
+    expect(find("h1.govuk-heading-xl")).to have_text "Your specification"
```
peterdavidhamilton, line 16: This second assertion is a more specific version of the one before. So it can replace it entirely, however, if this string appears elsewhere, and we aren't checking, then we can target that HTML element instead.

spec/features/self-serve/journeys/view_spec.rb
```diff
@@ -60,7 +63,9 @@
     click_continue
     click_view
 
-    expect(page).to have_content(I18n.t("journey.specification.header"))
+    # journey.specification.header
+    expect(page).to have_text "Your specification"
```
peterdavidhamilton, line 67: Maybe this is also used as the page title?

peterdavidhamilton:
There are maybe 10 more instances of `expect(page).to have_text` appearing in the test suite, can we make them more specific?

peterdavidhamilton:
@PiperS52 are there any more locales or constants hiding in the test suite?

PR #551, opened 2021-10-01
199 resolve case

Claudio-And:


app/controllers/support/cases_controller.rb
```diff
@@ -1,6 +1,6 @@
 module Support
   class CasesController < ApplicationController
-    before_action :current_case, only: %i[show edit update]
+    before_action :current_case, only: %i[show edit update resolve]
```
peterdavidhamilton, line 3: This is not RESTful

app/services/support/resolve_case.rb
```diff
@@ -0,0 +1,22 @@
+module Support
```
peterdavidhamilton, line 1: Again heavy on the service objects which requires unit testing, we don't necessarily need them out the gate on every controller action. For example, is there a service that can handle any and all updates to case? You're essentially defining the interface without knowing enough about it at this point in my opinion.

spec/services/support/resolve_case_spec.rb
```diff
@@ -0,0 +1,54 @@
+RSpec.describe Support::ResolveCase do
```
peterdavidhamilton, line 1: We've mentioned before that the feature coverage is more powerful and is the priority for our testing.

app/controllers/support/cases_controller.rb
AndrewCowle: I agree - did discuss this this afternoon with @Claudio-And - said it would get picked up at code review - think this might be an interim thing. I suggested a 'change state controller' as this could be called for various changes.

app/services/support/resolve_case.rb
AndrewCowle: Yes, I think service objects are great if functionality is called in multiple places - like maybe adding an interaction - just keep it simple at this point, do this on the CRUD controllers for now - can always refactor later.

app/views/support/cases/show.html.erb
```diff
@@ -30,7 +30,7 @@
 
   <div class="govuk-tabs__panel govuk-tabs__panel--hidden" id="request-details" role="tabpanel" aria-labelledby="tab_request-details">
     <p class="govuk-body">
-      requested at
+      <%= I18n.t("support.cases.show.requested_at", time: @current_case.interactions.first.created_at) %>
```
AndrewCowle, line 33: this has nothing to do with resolving a case - keep the PRs light and concise - it's something i'm going to do in another PR....

app/views/support/cases/show.html.erb
```diff
@@ -44,7 +44,23 @@
...
+            </h2>
+            <div class="govuk-accordion__section-summary govuk-body" id="accordion-with-summary-sections-summary-1">
+              <%= interaction.created_at %>
+            </div>
+          </div>
+          <div id="accordion-default-content-1" class="govuk-accordion__section-content" aria-labelledby="accordion-default-heading-1">
+            <%= render "interaction", interaction: interaction %>
+          </div>
+        </div>
+      <% end %>
```
AndrewCowle, line 63: have you pulled from my list interactions? again not needed in this PR

spec/features/support/case_show_spec.rb
```diff
@@ -101,6 +101,49 @@
...
   end
 
+  context "when the case is resolved" do
+    let(:state) { "resolved" }
+
+    it "shows the reopen case link" do
+      expect(find("a.reopen")).to have_text "Reopen case"
+    end
+
+    it "shows accordion" do
```
AndrewCowle, line 111: accordion - is this unique might be better to have this in a within block to make sure you're targeting the specific interaction.

spec/services/support/resolve_case_spec.rb
AndrewCowle: I think this isn't the right use-case of service objects, just do it in a controller that's sole responsibility is for changing state.

AndrewCowle (changes requested):
Appreciate this is a work in progress and you felt was blocked with not having interactions in your code to use, but I can't see anything adding an interaction - this is just changing state, if you shift it all to a controller who's just in charge of changing state we could add callbacks into the model to do things like adding a note when state changes (similar to how Acts As State Machine works). DB transactions are a good suggestion but what we are doing here doesn't warrant the work - as it's only touching a single record, int change and a note isn't going to be resource heavy. 

Apologies if the above sounds negative, I'm just a little anxious about the S&T on Monday and it would have been awesome to have all those 'action links' working.

Pick this up Monday.

app/controllers/support/cases_controller.rb
Claudio-And: Changed - it's now just a `get` action for showing the template

peterdavidhamilton: This afternoon can you schedule a little meet and you can try to sell us on this custom action approach

app/controllers/support/cases_controller.rb
```diff
@@ -10,13 +9,43 @@ def index
...
+            case_id: current_case.id,
+            agent_id: current_user.id,
+          }],
+        }
+      end
+
+      current_case.update!(attrs)
 
-      redirect_to support_case_path(anchor: "case-history")
+      redirect_to support_case_path(anchor: "case-history", notice: "Update successful")
```
AndrewCowle, line 48: flash in locale better

AndrewCowle (approved):
Cool :)

ryantk:
This PR should be closed in favour of https://github.com/DFE-Digital/buy-for-your-school/pull/589

PR #562, opened 2021-10-08
697 - Display signed in user's name in header bar

ryantk:
## Changes in this PR

The name of the signed in user is displayed in the header bar when signed in.

### Before

Screenshot 2021-10-08 at 09 56 18

### After

Screenshot 2021-10-08 at 09 55 15


peterdavidhamilton:
Great stuff this work will be warmly received 👍🏼

PR #593, opened 2021-10-26
Replace fixtures with factory

PiperS52:
## Changes in this PR

- refactoring relevant test files across code base, replacing fixtures with use of factories


spec/features/self-serve/journeys/task_list/view_hidden_spec.rb
```diff
@@ -1,35 +1,53 @@
...
-      let(:fixture) { "show-one-additional-question-in-order.json" }
+      before do
+        task_1 = create(:task, title: "One additional question task", section: section)
+        create(
+          :step,
+          :radio,
+          title: "What support do you have available?",
+          task: task_1,
+          options:
+          [{ "value" => "School expert",
```
peterdavidhamilton, line 35: Was expecting to see `options` and other `json` fields pulled out in to `let` for the goal of readibility, formatting, but likely also handy for testing other contexts

spec/features/self-serve/journeys/task_list/view_steps_spec.rb
```diff
@@ -38,11 +46,9 @@
   end
 
   context "when a task has one step" do
-    let(:fixture) { "section-with-single-task.json" }
-
     it "user can navigate back to the task list from a step" do
       within(".app-task-list") do
-        click_on "Task with a single step" # > checkboxes_task.json
+        click_on "Long text task" # > checkboxes_task.json
```
peterdavidhamilton, line 51: The comments that help guide someone to the old `json` fixture are no longer relevant but at worse could be missleading if left in.

spec/features/self-serve/metrics/log_view_task_spec.rb
```diff
@@ -117,7 +126,7 @@
         expect(last_logged_event.journey_id).to eq(journey.id)
         expect(last_logged_event.user_id).to eq(user.id)
         expect(last_logged_event.contentful_category_id).to eq "contentful-category-entry"
-        expect(last_logged_event.contentful_section_id).to eq "multiple-tasks-section"
+        expect(last_logged_event.contentful_section_id).to eq "contentful-section-entry"
```
peterdavidhamilton, line 129: I think what would be wisest when refactoring specs is to never change the tests until after. So here you're not only replacing fixtures with factories but also effecting the assertions. Rather, build a `section` whose `id` exactly matches that coming from the old fixture. It might be fine, but changing both at the same time could be problematic.

spec/features/self-serve/steps/presenter_spec.rb
```diff
@@ -1,121 +1,139 @@
...
-    click_link "long_text"
-    expect(page).to have_a_step_path
-    fill_in "answer[response]", with: "\r\n\r\nfoo\r\n\r\n"
-    click_on "Continue"
-    expect(page).to have_a_step_path
-    click_on "Back"
-    expect(page).to have_a_task_path
-    within("div.govuk-summary-list__row", text: "Describe what you need") do
-      expect(find("dd.govuk-summary-list__value")).to have_text "foo"
+    specify do
```
peterdavidhamilton, line 35: What does the output look like in the `documentation` format? Can we add a description to these blocks to make it any better in your opinion?

peterdavidhamilton (changes requested):
There are more files in the test suite that didn't have `NOTES` identifying them but they are also on the cards. If you can identify them these are going to be good tasks to fall back on whilst waiting for a code review.

app/models/journey.rb
```diff
@@ -60,7 +60,7 @@ def start!
   #
   # @return [Section, Nil]
   def next_incomplete_section(current_section = nil)
-    next_in_order = sections.detect do |section|
+    next_in_order = sections.includes([:tasks]).detect do |section|
```
peterdavidhamilton, line 63: So this was where you silenced that Bullet n+1 error?

spec/features/self-serve/journeys/task_list/view_answered_spec.rb
```diff
@@ -1,17 +1,27 @@
 RSpec.feature "Users can view the task list" do
   let(:user) { create(:user) }
-  let(:fixture) { "section-with-multiple-tasks.json" }
+  let(:category) { create(:category, :catering, contentful_id: "contentful-category-entry") }
+  let(:journey) { create(:journey, user: user, category: category) }
+  let(:section_one) { create(:section, title: "Catering", journey: journey, contentful_id: "contentful-section-entry") }
```
peterdavidhamilton, line 5: `section_one` and below `section_two` don't have their `order` set unless I'm mistaken. Does this mean that sometimes they are reversed?

app/models/journey.rb
PiperS52: Yeah

spec/features/self-serve/journeys/task_list/view_steps_spec.rb
```diff
@@ -126,25 +130,25 @@
       click_continue
 
       check "Breakfast"
-      click_continue
+      click_continue # > error clicking 'Continue'
```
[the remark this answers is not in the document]
PiperS52: bug flagging - I've removed

spec/features/self-serve/journeys/task_list/view_answered_spec.rb
PiperS52: Order doesn't seem to be necessary as section_two is only used on it's own for a single context

spec/features/self-serve/steps/presenter_spec.rb
PiperS52: Description added to make tests clearer

EdwinKruglov (approved):

d-a-v-e (approved):
much clearer with the factories, nice work.

spec/features/self-serve/journeys/task_list/view_steps_spec.rb
peterdavidhamilton: Can we get rid of the mention of `checkboxes_task.json` please and anything else that is now unrelated

peterdavidhamilton (approved):

peterdavidhamilton:
Time well spent 

bitmoji

PR #635, opened 2021-11-03
212 - Case categorisation

ryantk:
## Changes in this PR

- Added ability for a case to be recategorised by the user
- Refactored Support::Category model to be a tree structure to support further nesting of categories in the future.
- Removal of Support::SubCategory

## Screenshots of UI changes

Screen Shot 2021-11-03 at 13 24 52 pm

Screen Shot 2021-11-03 at 13 25 02 pm


ryantk:
[Display of Category / SubCategory](https://trello.com/c/ChelZRbr/212-display-of-category-subcategory)

AndrewCowle (approved):

peterdavidhamilton (changes requested):
We need to add a missing slug to the appropriate YAML category entry for `mfd`, otherwise `CaseCategorisationForm` will error.

peterdavidhamilton:
Is it ok for `govuk-radios__divider` to force `Office supplies` onto two lines?

ryantk:
> We need to add a missing slug to the appropriate YAML category entry for `mfd`, otherwise `CaseCategorisationForm` will error.

Hi Peter can you explain this a bit more for me please, I don't understand.

ryantk:
> @ryantk Is it ok for `.govuk-radios__divider` to force `Office supplies` onto two lines?

good spot

peterdavidhamilton:
We need to support two categories when this is merged and we use the slug to marry up, but the seed data in Supported does not currently have an `mfd` slug to achieve this. Cases for `catering` will work but there other won't and will cause a 500 error because `category_id` will be nil.

peterdavidhamilton:
I'll approve the requested change so it doesn't block you if I get called away.

peterdavidhamilton (approved):

ryantk:
> We need to support two categories when this is merged and we use the slug to marry up, but the seed data in Supported does not currently have an `mfd` slug to achieve this. Cases for `catering` will work but the other won't and will cause a 500 error because `category_id` will be nil.

I get it now. I actually spotted this myself after you put it and my fix above will fix it. I have added mfd slug into the categories seeds too :)

PR #637, opened 2021-11-03
Add user feedback link (Qualtrics)

d-a-v-e:
## Changes in this PR

- Button linking to Qualtrics added to view specifications page
- qualtrics env's `<ENV>_QUALTRICS_SURVEY_URL`'s added to github secrets

### Before

Screenshot 2021-11-03 at 18 29 38

### After

Screenshot 2021-11-03 at 18 29 47

### Next steps

- [x] check page is rendering correctly (supplied screenshots are from the prototype site)


d-a-v-e:
@peterdavidhamilton little unsure where to add this in the changelog, can you sanity check where I've put it?

d-a-v-e:
[Qualtrics survey link](https://trello.com/c/vugnvaW0/766-qualtrics-survey-link)

.env.example
```diff
@@ -16,6 +16,7 @@ REDIS_URL=redis://localhost:6379
 
 # Miscellaneous
 DAYS_A_JOURNEY_CAN_BE_INACTIVE_FOR=30
+QUALTRICS_SURVEY_URL=https://dferesearch.fra1.qualtrics.com/jfe/form/SV_cH0KmiM6qDX3Bye
```
peterdavidhamilton, line 19: Best not include the actual env value here

.env.test
```diff
@@ -33,4 +33,7 @@ DFE_SIGN_IN_API_ENDPOINT=https://test-api.signin.education.gov.uk
 DFE_SIGN_IN_API_SECRET=secret
 
 # Case Management
-PROC_OPS_TEAM=
\ No newline at end of file
+PROC_OPS_TEAM=
+
+# Miscellaneous
+QUALTRICS_SURVEY_URL=https://dferesearch.fra1.qualtrics.com/jfe/form/SV_cH0KmiM6qDX3Bye
```
peterdavidhamilton, line 39: Or here

config/locales/en.yml
```diff
@@ -131,6 +131,10 @@ en:
       next_steps:
         link_text: see what the next steps are.
         text: "Once you have a completed specification, "
+      feedback:
+        title: Giving us feedback
```
peterdavidhamilton, line 135: Can we use keys of `heading`, `body`, `button` please

spec/features/self-serve/journeys/download_spec.rb
```diff
@@ -17,12 +17,18 @@
 
       # journey.specification.header
       expect(find("h1.govuk-heading-xl")).to have_text "Your specification"
+      # journey.specification.feedback.title
+      expect(find("h2.govuk-heading-m")).to have_text "Giving us feedback"
+      # journey.specification.feedback.text
```
peterdavidhamilton, line 22: Would using a `within` block around some assertions improve this?

app/views/specifications/show.html.erb
```diff
@@ -17,5 +17,13 @@
     <%= I18n.t("journey.specification.next_steps.text") %>
     <%= link_to I18n.t("journey.specification.next_steps.link_text"), "/next-steps-#{@journey.category.slug}" %>
   </p>
+  <h2 class="govuk-heading-m">
+    <%= t("journey.specification.feedback.title") %>
+  </h2>
+  <p>
+    <%= t("journey.specification.feedback.text") %>
+  </p>
+  <%= button_to t("journey.specification.feedback.button_text"), ENV.fetch('QUALTRICS_SURVEY_URL'), class: "govuk-button govuk-button--secondary", "data-module" => "govuk-button" %>
```
peterdavidhamilton, line 26: `link_to` should get styled the same, we don't have a test that asserts the value of `QUALTRICS_SURVEY_URL` is being used in the link though. I think that's important.

app/views/specifications/show.html.erb
```diff
@@ -17,5 +17,13 @@
     <%= I18n.t("journey.specification.next_steps.text") %>
     <%= link_to I18n.t("journey.specification.next_steps.link_text"), "/next-steps-#{@journey.category.slug}" %>
   </p>
+  <h2 class="govuk-heading-m">
+    <%= t("journey.specification.feedback.title") %>
```
peterdavidhamilton, line 21: Can we stick with `I8n.t` for consistency for now

peterdavidhamilton (changes requested):
Some pedantry but the feature test coverage of the env is essential

spec/features/self-serve/journeys/download_spec.rb
```diff
@@ -17,16 +17,24 @@
...
       expect(find("p.govuk-body")).to have_text "Your answers have been used to create a specification, which also includes standard rules and regulations. You can go back and edit your answers if needed."
 
-      # journey.specification.next_steps
-      expect(find("article.specification p")).to have_text("Once you have a completed specification,")
-      expect(find("article.specification p")).to have_link("see what the next steps are.", href: "/next-steps-catering")
-
       # journey.specification.download.button
-      expect(find("a.govuk-button")).to have_text "Download (.docx)"
-      expect(find("a.govuk-button")[:role]).to eq "button"
+      expect(find("a.govuk-button:contains('Download (.docx)')")[:role]).to eq "button"
```
d-a-v-e, line 25: @peterdavidhamilton 

Finding the project expectation of these feature specs a tad confusing. I've ended up with this as a result...which looks convoluted to me.

I'd have just relied on the `click_on` further down to confirm this is on the page, but thought If I removed this it would end up not getting a pass in review again.

If I leave it as it is, It clashes with the other `a.govuk-button` .. I suppose I could add another class to distinguish between this lin and the `Give feedback (opens in a new tab)` link but that feels wrong too.

peterdavidhamilton:
@d-a-v-e `CHANGELOG` is fine, it's just because we have been forced to go without a tagged release for a while and have essentially two logs to split now. There is a ticket to tidy and divide the log before a proper release version.

spec/features/self-serve/journeys/download_spec.rb
peterdavidhamilton: You will not find consistency on all the specs yet, which is something we are working towards, it is evident there is a mix of developer approaches in play. My opinion is that the specs should not depend on any code internals and would therefore check against raw strings for example. When asserting page content, start with `expect(page).to` but as complexity increases the goal would be to articulate the key takeaways that the page is trying to achieve. So if there is a particular tag or class being used, consider if there is improved clarity by looking that up. There is a balance and I guess what I'm saying is, be as specific as you need to add value to the test, which is going to depend on the content and structure.

spec/features/self-serve/journeys/download_spec.rb
```diff
@@ -17,16 +17,24 @@
...
+      within "article.specification" do
+        # journey.specification.next_steps
+        expect(page).to have_text("Once you have a completed specification,")
+        expect(page).to have_link("see what the next steps are.", href: "/next-steps-catering")
+        # journey.specification.feedback.heading
+        expect(page).to have_text "Giving us feedback"
+        # journey.specification.feedback.body
+        expect(page).to have_text "Tell us how Get Help Buying for Schools is working for you."
+        # journey.specification.feedback.button
+        expect(page).to have_link("Give feedback (opens in a new tab)", href: "https://dferesearch.fra1.qualtrics.com")
```
peterdavidhamilton, line 36: For example, asserting that the class `govuk-button--secondary` exists on this link would tell me it is different to the other link, and that it is not green. I'd like to see headings get found because then the spec reads like the page and you can tell the structure, I can infer that from the locales comment but only because of the `heading` key.

spec/features/self-serve/journeys/download_spec.rb
peterdavidhamilton: The more appropriate finder for this tag is probably `have_link`

CHANGELOG.md
```diff
@@ -85,6 +85,7 @@ The format is based on [Keep a Changelog 1.0.0].
 - add `skipped_ids` to `Task` to allow users to skip questions
 - add error summary component
 - customise presence validation message
+- provide qualtrics feedback link
```
peterdavidhamilton, line 88: Upon reflection, mentioning `QUALTRICS_SURVEY_URL` would be useful.

peterdavidhamilton (dismissed):
Could you add the `h2`, button class, and remove the byebug history please, then this is good.

peterdavidhamilton (approved):

EdwinKruglov (approved):

PiperS52 (approved):

PR #643, opened 2021-11-05
Enabled skipped specs

AndrewCowle:
## Changes in this PR

- enabled skipped specs


peterdavidhamilton (approved):

db/schema.rb
```diff
@@ -44,7 +45,7 @@
     t.datetime "created_at", precision: 6, default: -> { "CURRENT_TIMESTAMP" }, null: false
     t.datetime "updated_at", precision: 6, default: -> { "CURRENT_TIMESTAMP" }, null: false
     t.integer "journeys_count"
-    t.string "slug"
+    t.string "slug", null: false
```
kennyevil, line 48: I think you might want to watch out for this. It keeps sneaking in so we'll have to do something about it.

kennyevil (changes requested):
Looks good. Just need to remove the schema change on line 48 and this should be good to go.

ryantk (approved):

kennyevil (approved):

PR #653, opened 2021-11-10
Capability to use chrome with capybara using js: true

ryantk:
## Changes in this PR

Config required to allow chrome usage with capybara to allow more advanced javascript in tests.

Use `js: true` in tests to activate the chrome driver.

### DIFF
--- a/script/server
+++ b/script/server
@@ -6,4 +6,11 @@
 set -e
 
 docker-compose up --detach
-docker attach buy-for-your-school_web_1
+
+check_container_name_format_command='docker ps --filter "name=buy-for-your-school_web_1" --format "{{.ID}}"'
+
+if [[ $(eval $check_container_name_format_command) ]]; then
+  docker attach buy-for-your-school_web_1
+else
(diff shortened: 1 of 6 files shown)


script/server
```diff
@@ -6,4 +6,11 @@
 set -e
 
 docker-compose up --detach
-docker attach buy-for-your-school_web_1
+
+check_container_name_format_command='docker ps --filter "name=buy-for-your-school_web_1" --format "{{.ID}}"'
```
ryantk, line 10: I have upgraded my docker setup and it is now replacing underscores with hyphens, so I have made this check in the script to enable both versions to work.

kennyevil (approved):
All looks good. That change to Docker is something I'm amazed by. Glad you caught it.

script/server
peterdavidhamilton: Interesting, what version are you currently?

AndrewCowle (approved):

peterdavidhamilton (approved):

PR #654, opened 2021-11-10
Correct yard syntax for dry-initialiser classes

EdwinKruglov:
## Changes in this PR

- update yard documentation for dry-initializer `options` and `params`

### DIFF
--- a/app/forms/support/case_assignment_form.rb
+++ b/app/forms/support/case_assignment_form.rb
@@ -3,6 +3,8 @@ module Support
     extend Dry::Initializer
     include Concerns::ValidatableForm
 
+    # @!attribute [r] agent_id
+    # @return [String]
     option :agent_id, Types::Params::String, optional: true
 
     def new_agent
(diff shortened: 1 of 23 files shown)


EdwinKruglov:
[Correct yard syntax for dry-initialiser classes](https://trello.com/c/2hnEtIm9/781-correct-yard-syntax-for-dry-initialiser-classes)

ryantk (approved):

peterdavidhamilton (changes requested):
The indentation is inconsistent but I don't know if that matters. I'd like to see things indented under `@attribute` please, and if `default` is set, it'd be good to see that in the doc

EdwinKruglov:
> The indentation is inconsistent but I don't know if that matters. I'd like to see things indented under `@attribute` please, and if `default` is set, it'd be good to see that in the doc

@peterdavidhamilton I've fixed the indentation and added a note about default values where necessary. I was hoping YARD would have a `@default` tag or similar, but it doesn't, so I just mentioned the default values on the `@return` tag.

PiperS52 (approved):

peterdavidhamilton (approved):

PR #660, opened 2021-11-16
Sending case management emails

peterdavidhamilton:
## Changes in this PR

- enable the "send email to school" action in case management
- convert the class to use the template UUID now that a single service is used in all environments
- update the tests accordingly
- convert templates to follow the same convention for variables as used in support requests
  + an object that has methods `first_name`, `last_name` and `email` 

## Next steps

- Ensure template content is correct and links to the prototype have been changed.
- Identify why some emails arrived with duplicate greetings and signatures??


peterdavidhamilton:
[Sending emails to school](https://trello.com/c/0JIUWIZv/338-sending-emails-to-school)

app/services/support/emails/to_school.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require "notify/email"
+
+module Support
+  class Emails::ToSchool < Notify::Email
+    def call
+      Rollbar.info "Sending email to school"
```
d-a-v-e, line 8: Does this need any more context passed to Rollbar? is eventually for a slack notification?

peterdavidhamilton: Perhaps but we shouldn't identify the user. We will see this in the slack channel, so for now, it is just letting us know the attempt was made.

AndrewCowle (approved):

app/controllers/support/cases/emails_controller.rb
```diff
@@ -4,27 +4,39 @@ def create
       @case_email_content_form = CaseEmailContentForm.from_validation(validation)
 
       if validation.success?
-        create_email_interaction(@case_email_content_form.email_body)
+
```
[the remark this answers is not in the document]
peterdavidhamilton: Don't think you have looked at the latest code @AndrewCowle

app/controllers/support/cases/emails_controller.rb
```diff
@@ -4,27 +4,39 @@ def create
...
+      params.require(:case_email_content_form).permit(:email_body, :email_subject, :email_template)
+    end
+
+    def send_email_to_school
+      Support::Emails::ToSchool.new(
+        recipient: @current_case,
+        template: params.fetch(:email_template, "ac679471-8bb9-4364-a534-e87f585c46f3"),
+        reference: @current_case.ref,
+        variables: {
+          text: @case_email_content_form.email_body,
```
ryantk, line 36: We probably want to have the same variables here as in preview
https://github.com/DFE-Digital/buy-for-your-school/blob/766fc83a49ee43b080f0dc59352fc60a05265f51/app/controllers/support/cases/emails/content_controller.rb#L53

ryantk (approved):

PR #664, opened 2021-11-18
Page services

EdwinKruglov:
## Changes in this PR

- add `get` and `build` services for `Pages`

### DIFF
--- a/app/services/content/page/build.rb
+++ b/app/services/content/page/build.rb
@@ -1,13 +1,11 @@
-require "dry-initializer"
-
 # Persist Page entries from Contentful
 #
 class Content::Page::Build
   extend Dry::Initializer
 
   # @!attribute [r] contentful_page
-  # @return [Contentful::Entry] Contentful "Page" entity
-  # @api private
+  #   @return [Contentful::Entry] Contentful "Page" entity
+  #   @api private
   option :contentful_page, optional: false, reader: :private
 
   # @return [Page]
--- a/app/services/content/page/build.rb
+++ b/app/services/content/page/build.rb
@@ -1,13 +1,14 @@
 require "dry-initializer"
 
-# Persits Page entries from Contentful
+# Persist Page entries from Contentful
 #
 class Content::Page::Build
   extend Dry::Initializer
 
   # @!attribute [r] contentful_page
-  # @return [Contentful::Entry]
-  option :contentful_page
+  # @return [Contentful::Entry] Contentful "Page" entity
+  # @api private
+  option :contentful_page, optional: false, reader: :private
 
   # @return [Page]
   def call
@@ -20,7 +21,7 @@ class Content::Page::Build
         sidebar: contentful_page.sidebar,
       },
       unique_by: :contentful_id,
-      returning: %w[title contentful_id slug body sidebar],
+      returning: %w[title contentful_id slug],
     )
 
     Rollbar.info("Built Contentful page", **page.first) if page
--- a/app/services/content/page/build.rb
+++ b/app/services/content/page/build.rb
@@ -23,8 +23,8 @@ class Content::Page::Build
       returning: %w[title contentful_id slug body sidebar],
     )
 
-    Rollbar.info("Built Contentful page", **page.first) if page.first
+    Rollbar.info("Built Contentful page", **page.first) if page
 
-    page.first
+    Page.find_by(contentful_id: contentful_page.id)
   end
 end
--- a/app/services/content/page/build.rb
+++ b/app/services/content/page/build.rb
@@ -1,0 +1,30 @@
+require "dry-initializer"
+
+# Persits Page entries from Contentful
+#
+class Content::Page::Build
+  extend Dry::Initializer
+
+  # @!attribute [r] contentful_page
+  # @return [Contentful::Entry]
+  option :contentful_page
+
+  # @return [Page]
+  def call
+    page = Page.upsert(
+      {
+        title: contentful_page.title,
+        body: contentful_page.body,
+        contentful_id: contentful_page.id,
+        slug: contentful_page.slug,
+        sidebar: contentful_page.sidebar,
+      },
+      unique_by: :contentful_id,
+      returning: %w[title contentful_id slug body sidebar],
+    )
+
+    Rollbar.info("Built Contentful page", **page.first) if page.first
+
+    page.first
+  end
+end
--- a/app/services/content/page/build.rb
+++ b/app/services/content/page/build.rb
@@ -6,7 +6,7 @@ class Content::Page::Build
   # @!attribute [r] contentful_page
   #   @return [Contentful::Entry] Contentful "Page" entity
   #   @api private
-  option :contentful_page, optional: false, reader: :private
+  option :contentful_page, reader: :private
 
   # @return [Page]
   def call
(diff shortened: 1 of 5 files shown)


EdwinKruglov:
[Pages service object](https://trello.com/c/zTn8EZPz/31-pages-service-object)

app/services/content/page/get.rb
```diff
@@ -0,0 +1,37 @@
...
+  option :page_entry_id, Types::String
+  # @!attribute [r] client
+  # @return [Content::Client]
+  option :client, default: proc { Content::Client.new }
+
+  # @raise [GetEntry::EntryNotFound] if page entry is not found
+  # @see GetEntry
+  # @return [Contentful::Entry]
+  def call
+    GetEntry.new(entry_id: page_entry_id, client: client).call
```
ryantk, line 20: There is very little difference between this class and GetEntry I would argue they could be collapsed into one.

peterdavidhamilton: I'm imagining that this is because we are setting ourselves up to bypass the cache on the Page model, and caching happens in GetEntry, so we need something very similar so we can isolate this without impacting other things. @EdwinKruglov?

spec/services/self-serve/content/page/get_spec.rb
```diff
@@ -0,0 +1,31 @@
+RSpec.describe Content::Page::Get do
+  subject(:service) { described_class.new(page_entry_id: page_entry_id, client: stub_client) }
+
+  let(:page_entry_id) { "page_entry_id" }
+
+  # rubocop:disable RSpec/AnyInstance
```
peterdavidhamilton, line 6: Depending on the circumstances we should leverage the rubocop todo file, sometimes that will guide us to changing the rules.

app/services/content/page/get.rb
ryantk: I feel there may be better ways to isolate in that case, but it also sounds like YAGNI.

app/services/content/page/build.rb
```diff
@@ -0,0 +1,30 @@
...
+        slug: contentful_page.slug,
+        sidebar: contentful_page.sidebar,
+      },
+      unique_by: :contentful_id,
+      returning: %w[title contentful_id slug body sidebar],
+    )
+
+    Rollbar.info("Built Contentful page", **page.first) if page
+
+    Page.find_by(contentful_id: contentful_page.id)
```
d-a-v-e, line 28: does returning `page` not work here?

d-a-v-e, line 28: does returning `page` not work here?

app/services/content/page/build.rb
```diff
@@ -0,0 +1,30 @@
+require "dry-initializer"
+
+# Persits Page entries from Contentful
```
d-a-v-e, line 3: lil typo

app/services/content/page/build.rb
```diff
@@ -0,0 +1,30 @@
...
+#
+class Content::Page::Build
+  extend Dry::Initializer
+
+  # @!attribute [r] contentful_page
+  # @return [Contentful::Entry]
+  option :contentful_page
```
d-a-v-e, line 10: suggestion:

```ruby
# @!attribute env
    #   @return [Contentful::Entry] Contentful "Page" entity
    #   @api private
option :contenful_page, optional: false, reader: private
```

app/services/content/page/build.rb
```diff
@@ -0,0 +1,30 @@
...
+    page = Page.upsert(
+      {
+        title: contentful_page.title,
+        body: contentful_page.body,
+        contentful_id: contentful_page.id,
+        slug: contentful_page.slug,
+        sidebar: contentful_page.sidebar,
+      },
+      unique_by: :contentful_id,
+      returning: %w[title contentful_id slug body sidebar],
```
d-a-v-e, line 23: Maybe just return title, contentful_id, slug

config/static/terms_and_conditions.md
```diff
@@ -67,7 +67,7 @@ We are not liable for any loss or damage that may arise from using the Service o
 
 - the inability to use the Service and any websites linked to or from it
 - any direct, indirect or consequential losses
-- any loss or damage caused by civil wrongs ('tort', including negligence), breach of contract or otherwise
+- any loss or damage caused by civil wrongs ('tort', including negligence, duty of care or public law duty), breach of contract or otherwise
```
d-a-v-e, line 70: you can remove this (it's already in develop)

d-a-v-e (changes requested):

app/services/content/page/build.rb
EdwinKruglov: [`upsert`](https://edgeapi.rubyonrails.org/classes/ActiveRecord/Persistence/ClassMethods.html#method-i-upsert_all) does not return a model object, but an `ActiveRecord::Result`, so I need to make another query to get the actual `Page`.

d-a-v-e (approved):

app/services/content/page/build.rb
```diff
@@ -0,0 +1,31 @@
...
+
+# Persist Page entries from Contentful
+#
+class Content::Page::Build
+  extend Dry::Initializer
+
+  # @!attribute [r] contentful_page
+  # @return [Contentful::Entry] Contentful "Page" entity
+  # @api private
+  option :contentful_page, optional: false, reader: :private
```
peterdavidhamilton, line 11: @d-a-v-e might have mentioned consistent indentation already. I think that type checking the input is more powerful that `optional: false` because that is expected anyway.

app/services/content/page/get.rb
```diff
@@ -0,0 +1,36 @@
...
+  # @!attribute [r] page_entry_id
+  # @return [String]
+  option :page_entry_id, Types::String
+  # @!attribute [r] client
+  # @return [Content::Client]
+  option :client, default: proc { Content::Client.new }
+
+  # @return [Contentful::Entry, nil]
+  def call
+    page = client.by_id(page_entry_id)
```
peterdavidhamilton, line 18: Not having looked into how this connects, but just a reminder that returning a `:useful_symbol` is sometimes a better return value than nil. You'll have seen this in the org checking code.

app/services/content/page/get.rb
peterdavidhamilton: @ryantk what is needed is a quick reliable response from Contentful with sensible webhook endpoints, and caching that helps not hinders. I know @EdwinKruglov is going to reflect on how code can be reviewed more easily by setting out his stall. Could you signpost him to "the better ways" or grab a quick chat together maybe.

app/services/content/page/build.rb
EdwinKruglov: I wanted to add type checking to this field, but that makes it difficult to test it. It's not a fan of doubles.

d-a-v-e (approved):

ryantk (approved):

app/services/content/page/build.rb
peterdavidhamilton: I'd suggest removing the `optional` key because it doesn't do anything. 

Consider not using a double? Depending on the test you could instantiate with `Contentful::Entry.new({}, {})` but pass in the `Hash` of attributes you need.

I mean `Types::Nominal::Any` is a thing though it doesn't add much value. The `Hash` is parsed `JSON` from the client and one of the things we can consider is if pure `JSON` is better for us at some point.

peterdavidhamilton (approved):

PR #667, opened 2021-11-18
[28] Dynamic Pages

d-a-v-e:
https://trello.com/c/Q5LS2Gdk/36-update-page-activerecord-model

## Changes in this PR

* Pages now served from contentful:
    * T&C's
    * Next Steps
    * Accessibility Statement
    * Privacy Notice
    * Participation agreement
* Removes
    * static pages previously defined in `config/pages`
    * use of Highvoltage for serving static pages

## Next steps

- [x] update changelog
- [x] port pages into staging
- [x] update footer
- [x] revert slugs back
- [x] update (unused) slug definitions in staging
- [x] update (unused) slug definitions in production
- [x] add contenful_ids to migration
- [ ] add pattern constraint to contentful slug field

## Deployment
* Go to [contentful](https://app.contentful.com/spaces/rwl7tyzv9sys/environments/staging/entries?id=yJyxPIX88387g6Vj&contentTypeId=page&order.fieldId=updatedAt&order.direction=descending&displayedFieldIds=metadata.tags&displayedFieldIds=author&displayedFieldIds=updatedAt&displayedFieldIds=contentType&roles=4WiVKNv33JSiWZSHxQAqjF&folderId=default&page=1), 
* check all pages
* hit `Publish`


peterdavidhamilton:
Where has the coverage dropped?

app/services/content/page/get.rb
```diff
@@ -0,0 +1,36 @@
+# Retrieve Page entries from Contentful
+#
+class Content::Page::Get
+  extend Dry::Initializer
+
+  # @!attribute [r] page_entry_id
+  #   @return [String]
+  option :page_entry_id, Types::String
```
peterdavidhamilton, line 8: Can we set this up so it is more `type` agnostic? If the first param were `entry_id`

app/services/content/page/get.rb
```diff
@@ -0,0 +1,36 @@
...
+  # @!attribute [r] client
+  #   @return [Content::Client]
+  option :client, default: proc { Content::Client.new }
+
+  # @return [Contentful::Entry, Symbol]
+  def call
+    page = client.by_id(page_entry_id)
```
peterdavidhamilton, line 15: And the variable were `entry`

app/services/content/page/get.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    if page.nil?
+      send_rollbar_error(message: "A Contentful page entry was not found")
```
peterdavidhamilton, line 18: And the message could cater to entries of different types

.rubocop_todo.yml
```diff
@@ -51,3 +54,7 @@ Rails/ReversibleMigration:
     - 'db/migrate/20201112123028_remove_answer_table.rb'
     - 'db/migrate/20210118110545_change_options_to_jsonb.rb'
     - 'db/migrate/20210125171840_remove_next_entry_id.rb'
+
+Style/TrailingCommaInArguments:
+  Exclude:
+    - 'spec/requests/self-serve/api/contentful/page_spec.rb'
```
peterdavidhamilton, line 60: This is a simple style one that could be addressed

app/controllers/api/contentful/base_controller.rb
```diff
@@ -10,6 +10,10 @@ def auth
 
 private
 
+  def contentful_id
+    params[:id] || params[:sys][:id]
```
peterdavidhamilton, line 14: I'm guessing `sys.id` is the default payload, so is `id` from a custom one?

app/controllers/pages_controller.rb
```diff
@@ -1,58 +1,24 @@
...
-
-  def show_route
+  def specifying_start_page
     @start_now_button_route = current_user.guest? ? "/auth/dfe" : dashboard_path
+    @start_now_button_method = current_user.guest? ? :post : :get
   end
 
-  def show_method
-    @start_now_button_method = current_user.guest? ? :post : :get
+  # TODO: remove this once pages are dynamic
```
peterdavidhamilton, line 14: This comment might be misleading, the bypass caters for working offline but would require a dev to fill in names because those fields would be blank. Do we think it is needed anymore? Just calling out that perhaps this shouldn't hinder us from having front page content be powered by contentful.

d-a-v-e:
[Dynamic Pages](https://trello.com/c/cn2Qfs4m/28-dynamic-pages)

app/controllers/api/contentful/base_controller.rb
d-a-v-e: `params[:id]` is used in the case of unpublish/delete.

app/controllers/pages_controller.rb
d-a-v-e: we've not moved the start page content to contentful as of yet, so we'll leave this in for now (unless I'm missing your point 🤔)

app/presenters/page_presenter.rb
```diff
@@ -0,0 +1,31 @@
+class PagePresenter < BasePresenter
+  def updated_at
+    text = I18n.t("page.updated_at")
+    date_format = I18n.t("page.date_format")
+    date = __getobj__.updated_at.strftime(date_format)
+
+    "#{text} #{date}"
```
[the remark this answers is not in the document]
d-a-v-e: done

app/presenters/page_presenter.rb
```diff
@@ -28,4 +25,10 @@ def container_class
   def body_class
     return "govuk-grid-column-two-thirds" if sidebar
   end
+
+private
+
+  def date_format
```
peterdavidhamilton, line 31: I was meaning we already have `date_format` set in the base presenter so it needn't be in a locale, but if it is to be controlled in the locales it needs to be the single place it is defined.

peterdavidhamilton (approved):

ryantk (approved):

PR #669, opened 2021-11-19
[34] Dynamic Routes

d-a-v-e:
## Changes in this PR
* Adds a new generic root route (yep!) 
* Any `/any-page-slug` request will be routed to `PagesController` where the page will be retrieved (if it exists) from the pages table
* Route is kept at the bottom of our routing config to make sure it doesn't supersede anything else


d-a-v-e:
[Dynamic routes](https://trello.com/c/HQLTiFOh/34-dynamic-routes)

config/routes.rb
```diff
@@ -18,12 +18,16 @@
   get "/422", to: "errors#unacceptable"
   get "/500", to: "errors#internal_server_error"
 
+  # Dyamic Pages (Page refreshes routes on create commits)
```
peterdavidhamilton, line 21: typo

d-a-v-e:
@peterdavidhamilton @EdwinKruglov @PiperS52 

I've removed the dynamically generated routes block and replaced with a generic routing to `Pages`

I assume this will fix the issue we were seeing in docker!

EdwinKruglov (approved):

peterdavidhamilton (changes requested):
@d-a-v-e we should add an ADR to document the move away from `high_voltage` (which doesn't have an ADR) to using Contentful to power page content. Try using the cli tool

peterdavidhamilton (approved):

PR #672, opened 2021-11-23
Add support case activity logging

kennyevil:
## Changes in this PR

Adds ActivityLogItem for support cases.


app/services/support/record_support_case_action.rb
```diff
@@ -0,0 +1,67 @@
+module Support
+  # Track support case activity in controller actions
+  #
+  # @see Support::ActivityLogItem
+
+  class RecordSupportCaseAction
+    class UnexpectedActionType < StandardError; end
+
+    ACTION_TYPES = %w[
```
peterdavidhamilton, line 9: I think we should keep the imperative mood of the actions so `open_case` rather than `opening_case`.

app/services/support/record_support_case_action.rb
```diff
@@ -0,0 +1,67 @@
...
+      changing_state
+      resolving_case
+      closing_case
+    ].freeze
+
+    # @param action [String]
+    # @param support_case_id [UUID]
+    # @param data [Hash]
+    #
+
```
peterdavidhamilton, line 22: Might need to tidy the YARD syntax a little, remove line gap and change UUID to String. Could we also adopt `dry-intitializer` here instead so that we can benefit from type checking the action when we initialise.

peterdavidhamilton (changes requested):
Does it make sense to also use `Support::RecordAction` and follow the same naming convention in the namespace?

app/services/support/record_action.rb
```diff
@@ -0,0 +1,70 @@
...
+      resolve_case
+      close_case
+    ].freeze
+
+    # TODO: add RecordAction for change_service_level, change_state and close_case when
+    # functionality has been added to the case controllers
+
+    # @!attribute action
+    #   @return [String]
+    option :action, Types::String
```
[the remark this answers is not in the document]
peterdavidhamilton: We could still report to rollbar by rescuing in the controller methods though. What's your opinion?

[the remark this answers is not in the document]
kennyevil: There's very few places that this action is called and certainly none of them can be affected by a malicious user. It feels like a custom error isn't really necessary as it probably isn't ever going to be used. If there is an issue with the code, having the stack trace from a standard rollbar error will probably be enough to determine the cause. I'll take out the custom error type.

spec/services/support/record_action_spec.rb
```diff
@@ -31,25 +31,7 @@
             action: "invalid_action",
             support_case_id: support_case.id,
           ).call
```
peterdavidhamilton, line 33: This will error even before we get to call it so it would be good to have the `call` removed.

peterdavidhamilton (changes requested):
```ruby
context "when the action does has additional parameters specified" do
end
```

The spec could be improved with a tweak to the text

app/services/support/record_action.rb
```diff
@@ -0,0 +1,44 @@
...
+      open_case
+      add_interaction
+      change_category
+      change_service_level
+      change_state
+      resolve_case
+      close_case
+    ].freeze
+
+    # TODO: add RecordAction for change_service_level, change_state and close_case when
```
peterdavidhamilton, line 21: Could we move this up so the comment is on the Constant please

app/services/support/record_action.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    # @!attribute support_case_id
+    #   @return [String]
+    option :support_case_id, Types::String
+
+    # @!attribute data
+    #   @return [Hash]
+    option :data, Types::Hash, default: proc { {} }
+
+    def call
```
peterdavidhamilton, line 36: It would be good to have the `@return` value in our docs

spec/services/support/record_action_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  let(:support_case) { create(:support_case) }
+
+  describe "#call" do
+    it "records the action in the database" do
+      action = described_class.new(
+        action: "open_case",
+        support_case_id: support_case.id,
+      ).call
+
+      expect(action.action).to eq "open_case"
```
peterdavidhamilton, line 11: It would be good to follow the convention of naming the POROs in the service directory as `service` in the specs. Here it would avoid having `action.action`.

app/controllers/support/interactions_controller.rb
```diff
@@ -33,6 +35,16 @@ def safe_interaction
         redirect_to(support_case_path(current_case))
     end
 
+    def record_support_case_activity_log_item
```
peterdavidhamilton, line 38: From a design perspective would it make sense to have a common method name in our controllers for logging transactions. The controller is setting the context in my opinion, and the namespace is clarifying the database table. So `log_activity` or `record_action` could be used instead; 6 snake-cased words seems a bit unwieldy.

peterdavidhamilton: `record_case_categorisation_change`, `record_case_resolved`, `record_support_case_activity_log_item`, `record_case_opening` - These methods serve the same purpose. it would make any refactoring easier and could facilitate the Rollbar reporting. Setting it in a base controller method that we overwrite, like `super("action_name", {arbitary: "data"})`

peterdavidhamilton:
[Manual export of service activity data](https://trello.com/c/ZuoGbMQg/107-manual-export-of-service-activity-data)

spec/services/support/record_action_spec.rb
```diff
@@ -1,37 +1,31 @@
...
-    it "records the action in the database" do
-      action = described_class.new(
-        action: "open_case",
-        support_case_id: support_case.id,
-      ).call
+    let(:action) { "open_case" }
 
-      expect(action.action).to eq "open_case"
+    it "records the action in the database" do
+      expect(service.call.action).to eq "open_case"
```
peterdavidhamilton, line 11: Could we assert that the log entries increment by one

app/controllers/support/application_controller.rb
```diff
@@ -19,5 +19,13 @@ def authenticate_agent!
 
       redirect_to support_root_path, notice: "You are not a recognised case worker"
     end
+
+    def record_action(support_case_id:, action:, data: {})
```
peterdavidhamilton, line 23: Could we make the param `case_id`, we only have "support" floating around because of our namespacing but I don't want it to influence methods etc

peterdavidhamilton (approved):

d-a-v-e (approved):

PR #675, opened 2021-11-23
Interim solution to export KPI data

peterdavidhamilton:
## Changes in this PR

- add a rake task to export activity log items as CSV data for convenience
- once the task has been run manually and the file downloaded it should be removed from the server

## TODO

- potential for restricted access CSV download route so Data Analysts can interact without developer input

### DIFF
--- a/lib/tasks/export.rake
+++ b/lib/tasks/export.rake
@@ -1,0 +1,7 @@
+desc "Export ActivityLog items"
+task export: :environment do
+  file = Rails.root.join("public/activity_log.csv")
+  data = ActivityLogItem.to_csv
+
+  File.open(file, "w+") { |f| f.write(data) }
+end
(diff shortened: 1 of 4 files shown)


peterdavidhamilton:
[Activity Log data exporting](https://trello.com/c/5NrYdIdz/110-activity-log-data-exporting)

kennyevil (approved):

lib/tasks/export.rake
```diff
@@ -0,0 +1,7 @@
+desc "Export ActivityLog items"
+task export: :environment do
```
kennyevil, line 2: Just a thought, but we might be running a few export tasks from here. Probably worth namespacing this task to reflect what it's exporting.

peterdavidhamilton: So far this is the only one and this is a stopgap in the absence of something automated so my hope is it goes away, but I'd agree on namespacing and also keeping the `support` tasks separate

d-a-v-e (approved):

PR #676, opened 2021-11-23
[32] Show dynamic pages

d-a-v-e:
## Changes in this PR

* `PagePresenter` added for rendering in `view/pages/show`


app/presenters/page_presenter.rb
```diff
@@ -0,0 +1,48 @@
+class PagePresenter < BasePresenter
+  def time_stamp
+    updated_at.strftime("%e %B %Y")
+  end
+
+  def present_body
```
d-a-v-e, line 6: rename to body, and use super within method

d-a-v-e:
[Show dynamic pages](https://trello.com/c/ReBfqBVd/32-show-dynamic-pages)

app/controllers/pages_controller.rb
```diff
@@ -1,38 +1,26 @@
...
-      to: :html,
-    ).call
-  end
-
   def page
-    @page ||= Page.find_by(slug: params[:slug])
+    return @page if @page
+
+    @page = Page.find_by(slug: params[:slug])
+    @page = PagePresenter.new(@page)
```
peterdavidhamilton, line 24: I think it can be good to make sure that we only decorate just before it hits the view, so it might be good to have `@page = PagePresenter.new(page)` in `#show`

app/presenters/page_presenter.rb
```diff
@@ -0,0 +1,19 @@
+class PagePresenter < BasePresenter
+  def body
```
peterdavidhamilton, line 2: Please could you drop the return value (which is a string more than 90% of the time) on these methods. It encourages useful comments too.

EdwinKruglov (approved):

spec/features/self-serve/pages/show_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "rails_helper"
+
+RSpec.feature "Showing a Contentful page" do
+  let(:contentful_page) { create(:page, sidebar: sidebar, body: body) }
+
+  before do
```
peterdavidhamilton, line 6: Do we need this block?

spec/features/self-serve/pages/show_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "rails_helper"
+
+RSpec.feature "Showing a Contentful page" do
+  let(:contentful_page) { create(:page, sidebar: sidebar, body: body) }
```
peterdavidhamilton, line 4: I think experience shows that this might best be named `page` because there will be specs where we have webmock stub the `contentful_page` JSON and this avoids ambiguity

spec/features/self-serve/pages/show_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "rails_helper"
+
+RSpec.feature "Showing a Contentful page" do
```
peterdavidhamilton, line 3: This feature is more about rendering a persisted page that had once come from Contentful but really could have also come from the old rake task. No external service is being hit in this spec so we should not mention it in the text.

app/views/pages/show.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= content_for :title, @page.title %>
+
+<div class="<%= @page.container_class%> md-override">
+  <div class="<%= @page.body_class %>">
+    <%= @page.body %>
+    <p class="govuk-body-s">Last updated <%= @page.updated_at %></p>
```
peterdavidhamilton, line 6: Is this an opportunity to to move the text into a locale and into the presenter too?

peterdavidhamilton (approved):
Some minor text change suggestions and an opportunity to get text out of the template. Great stuff.

PiperS52 (approved):

spec/features/self-serve/pages/show_spec.rb
d-a-v-e: named it page before then bashed my head against my keyboard for a while wondering why my specs wouldn't run. `page` already refers to the page we're testing, so its not possible to `expect(page).to`.

spec/features/self-serve/pages/show_spec.rb
d-a-v-e: no - a vestige of my aforementioned head-bashing around `page` will remove!

PR #677, opened 2021-11-23
Add csv export task for Support::Cases

kennyevil:
## Changes in this PR

- add a `to_csv` method into `Support::Case`
- add case data as an option to the `export` rake task

### DIFF
--- a/lib/tasks/export.rake
+++ b/lib/tasks/export.rake
@@ -6,10 +6,10 @@ task export: :environment do
   File.open(file, "w+") { |f| f.write(data) }
 end
 
-namespace :support do
-  namespace :cases do
+namespace :export do
+  namespace :support do
     desc "Export Support Cases"
-    task export: :environment do
+    task cases: :environment do
       file = Rails.root.join("public/support_cases.csv")
       data = Support::Case.to_csv
 
--- a/lib/tasks/export.rake
+++ b/lib/tasks/export.rake
@@ -6,10 +6,10 @@ task export: :environment do
   File.open(file, "w+") { |f| f.write(data) }
 end
 
-namespace :export do
-  namespace :support do
+namespace :support do
+  namespace :case_management do
     desc "Export Support Cases"
-    task cases: :environment do
+    task export: :environment do
       file = Rails.root.join("public/support_cases.csv")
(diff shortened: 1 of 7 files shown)


peterdavidhamilton:
[Manual export of service activity data](https://trello.com/c/ZuoGbMQg/107-manual-export-of-service-activity-data)

AndrewCowle (commented):
What do you guys think about having this as a concern (likely that we might want to have CSV export on a few models) @peterdavidhamilton @kennyevil ?

kennyevil:
> What do you guys think about having this as a concern (likely that we might want to have CSV export on a few models) @peterdavidhamilton @kennyevil ?

This is a good idea. I've moved it into a concern called csvable.

peterdavidhamilton:
> > What do you guys think about having this as a concern (likely that we might want to have CSV export on a few models) @peterdavidhamilton @kennyevil ?
> 
> This is a good idea. I've moved it into a concern called csvable on #672

For the reasons discussed it might be easier not to. I'm not fussed though but we don't want concerns being shared across the namespace, which means we still double up, and the hope is we deleted this anyway. Are we ok to revert that?

peterdavidhamilton (approved):

lib/tasks/export.rake
```diff
@@ -1,3 +1,12 @@
...
+desc "Export ActivityLog items"
+task export: :environment do
+  file = Rails.root.join("public/activity_log.csv")
+  data = ActivityLogItem.to_csv
+
+  File.open(file, "w+") { |f| f.write(data) }
+end
+
 namespace :support do
   namespace :cases do
```
peterdavidhamilton, line 11: There are potentially two tasks in the support namespace, one for exporting cases, the other for log items. This one is named wrong.

lib/tasks/export.rake
```diff
@@ -5,3 +5,15 @@ task export: :environment do
 
   File.open(file, "w+") { |f| f.write(data) }
 end
+
+namespace :export do
```
peterdavidhamilton, line 9: There is a `case_management` namespace in use within our tasks

d-a-v-e (approved):

PR #680, opened 2021-11-25
Unsupported scope

PiperS52:
## Changes in this PR

- scope filters for both supported and unsupported orgs
- unsupported predicate method


PiperS52:
[`unsupported` scope](https://trello.com/c/e31Xypqw/17-unsupported-scope)

PiperS52:
[`unsupported?` predicate method](https://trello.com/c/BQwbA8uJ/16-unsupported-predicate-method)

app/models/user.rb
```diff
@@ -5,13 +5,18 @@ class User < ApplicationRecord
...
 
   # @return [false] distinguish from unauthenticated user
   #
   def guest?
     false
   end
+
+  # @return [Boolean]
+  def unsupported?
+    User.unsupported.include?(self)
```
peterdavidhamilton, line 20: Can we do without the class explicitly referencing itself

d-a-v-e (approved):

app/models/user.rb
EdwinKruglov: We tried that, but we couldn't find a way around it. The scopes live at the class level, so it doesn't look like we can simply call them from witihn an instance method.

app/models/user.rb
```diff
@@ -5,13 +5,23 @@ class User < ApplicationRecord
...
 
   # @return [false] distinguish from unauthenticated user
   #
   def guest?
     false
   end
+
+  # @return [Boolean]
+  def unsupported?
+    User.unsupported.include?(self) && User.proc_ops.exclude?(self)
```
peterdavidhamilton, line 25:
```ruby
# @return [Boolean] user is not an internal team member or in a supported organisation
self.class.supported.exclude?(self) && self.class.internal.exclude?(self)
```

app/models/user.rb
```diff
@@ -5,13 +5,23 @@ class User < ApplicationRecord
...
-  # from supported establishments
-  # scope :supported, ->(ids) { where("(data->'type'->>'id' ? :value)::int", value: ids) }
+  # users belonging to supported organisations
+  scope :supported, -> { where("orgs @> any(array[?]::jsonb[])", ORG_TYPE_IDS.map { |id| %([{"type": {"id": "#{id}"}}]) }) }
+
+  # users belonging to unsupported organisations
+  scope :unsupported, -> { where.not("orgs @> any(array[?]::jsonb[])", ORG_TYPE_IDS.map { |id| %([{"type": {"id": "#{id}"}}]) }) }
+
+  # users belonging to the proc-ops team
+  scope :proc_ops, -> { where("orgs @> ?", %([{"name": "#{ENV['PROC_OPS_TEAM']}"}])) }
```
peterdavidhamilton, line 15: I think it's wise to rename this, only because in this context, it's not so much that they are bound for the `/support` side of things, more that they are `internal` to the project.

peterdavidhamilton: Other scope name suggestions are welcomed

peterdavidhamilton (changes requested):

spec/models/self-serve/user/user_supported_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  describe "supported" do
+    before do
+      create(:user,
+             first_name: "test_supported",
+             orgs: [{
+               "id": "23F20E54-79EA-4146-8E39-18197576F023",
+               "name": "Supported School Name",
+               "type": {
+                 "id": "1",
```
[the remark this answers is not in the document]
EdwinKruglov: If the `id` is an int then no, it doesn't work. It looks like DSI always returns the organisation type ID as a string, though.

spec/models/self-serve/user/user_unsupported_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+               "name": "Supported School Name",
+               "type": {
+                 "id": "1",
+               },
+             }])
+    end
+
+    context "when single org" do
+      it "only returns unsupported users" do
+        expect(described_class.unsupported.count).to eq(1)
```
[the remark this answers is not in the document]
EdwinKruglov: We've updated the specs so they should be easier to follow now.

peterdavidhamilton (approved):

PR #689, opened 2021-11-29
Fix autocomplete default value

ryantk:
## Changes in this PR

Fix autocomplete so that we can use defaultValue (used when returning to a pre-populated form).
Previously the template was displaying incorrectly when focusing on the input field when the default value is populated, this fixes it by just showing the value itself.

I have also tweaked the template itself to look like the one in prototype.


peterdavidhamilton (approved):

ryantk:
https://github.com/DFE-Digital/buy-for-your-school/pull/689

kennyevil (approved):

AndrewCowle (approved):

PR #690, opened 2021-11-29
Change open status to opened on Support::Case

kennyevil:
## Changes in this PR

Change the status "open" to "opened" on Support::Case model to avoid method name collision


kennyevil:
[Fix method name colision](https://trello.com/c/fe5s7J9i/129-fix-method-name-colision)

ryantk (approved):

peterdavidhamilton (approved):

PR #691, opened 2021-12-01
Updated support request additional data (interaction) styling

AndrewCowle:
## Changes in this PR

- Updated 'additional data' styling for the support_request interaction,
- Updated case presenter to account for nil organisation
- Submit support request to use CreateCase & Create Interaction services

### Before

Screenshot 2021-12-01 at 08 48 52

### After

Screenshot 2021-12-01 at 08 48 35

### DIFF
--- a/app/services/submit_support_request.rb
+++ b/app/services/submit_support_request.rb
@@ -97,7 +97,8 @@ private
                                   email: user.email,
                                   phone_number: request.phone_number,
                                   organisation: org,
-                                  category: map_category)
+                                  category: map_category,
+                                  source: "digital")
 
     record_case_opening
 
--- a/app/services/submit_support_request.rb
+++ b/app/services/submit_support_request.rb
@@ -38,7 +38,7 @@ class SubmitSupportRequest
 
   # TODO: Replace with outbound API call
   #
-  # @return [nil, Notifications::Client::ResponseNotification]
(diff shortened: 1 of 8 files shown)


app/services/submit_support_request.rb
```diff
@@ -87,31 +87,39 @@ def map_category
...
+      additional_data:
+        { "support_request_id": request.id,
+          "first_name": user.first_name,
+          "last_name": user.last_name,
+          "email": user.email,
+          "phone_number": request.phone_number,
+          "category": category,
+          "message": request.message_body },
+    }
+    Support::CreateInteraction.new(@kase.id, "support_request", attrs: interaction_attrs).call
```
ryantk, line 119: This is up for discussion but I think it would be cleaner if `CreateInteraction` took in the objects it needs to create the interaction rather than the attributes themselves, that way if we change what is required to make an interaction we only need to update it in one place. This is not unique to `CreateInteraction` but something I've noticed about a few of our service classes.

ryantk (approved):

app/services/submit_support_request.rb
AndrewCowle: Yes, it was a little tricky since we aren't using API yet - different objects can create interactions. I agree this needs refactoring once the API is in place as this will all happen in the Supported side :)

kennyevil (approved):

PR #698, opened 2021-12-03
[154] Correct Page CSS

d-a-v-e:
## Changes in this PR

* Removes bullet styling of lists for page sidebar
* reinstates timestamp styling

### Before

Screenshot 2021-12-03 at 14 08 08

### After
Screenshot 2021-12-03 at 14 07 16

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -3,7 +3,7 @@ $govuk-image-url-function: "image-url";
 @import "base/frontend";
 @import "components/govuk-radios-small";
 @import "components/autocomplete";
-@import "components/override";
+@import "components/page";
 @import "components/task-list";
 @import "components/supported/email-preview";
 @import "components/supported/new-case-form";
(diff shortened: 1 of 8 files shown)


d-a-v-e:
[Correct Page CSS](https://trello.com/c/NjUQ5tEf/154-correct-page-css)

peterdavidhamilton (changes requested):
Could we go for an `#id` and `@extend` using the specific style we need.
https://sass-lang.com/documentation/at-rules/extend
We only have one sidebar per page and this would defer to the front-end lib.

d-a-v-e:
> Could we go for an `#id` and `@extend` using the specific style we need. https://sass-lang.com/documentation/at-rules/extend We only have one sidebar per page and this would defer to the front-end lib.

@peterdavidhamilton thanks, this looks better now. Have done the same for the timestamp as well.

EdwinKruglov (approved):

peterdavidhamilton (approved):
I think it would be good to eventually do away with the `specification` class and id because now we have markdown in multiple contexts. Good stuff. 👍🏼

PR #701, opened 2021-12-07
Ensure agent cant be nil

AndrewCowle:
## Changes in this PR

- Modified Support::ApplicationController to ensure agent is present when accessing the case management system
- Added specs
- Fixed interaction spec to have agent

### DIFF
--- a/app/controllers/support/application_controller.rb
+++ b/app/controllers/support/application_controller.rb
@@ -6,16 +6,14 @@ module Support
 
     helper_method :current_agent
 
-    # @return [Agent, nil]
+    # @return [AgentPresenter, nil]
     def current_agent
-      @current_agent ||= AgentPresenter.new(
-        Agent.find_by(dsi_uid: session[:dfe_sign_in_uid]),
-      )
+      AgentPresenter.new(Agent.find_by(dsi_uid: session[:dfe_sign_in_uid]))
     end
 
     # @return [nil]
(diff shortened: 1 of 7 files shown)


kennyevil (approved):

app/controllers/support/application_controller.rb
```diff
@@ -6,16 +6,14 @@ class ApplicationController < ::ApplicationController
 
     helper_method :current_agent
 
-    # @return [Agent, nil]
+    # @return [AgentPresenter, nil]
     def current_agent
-      @current_agent ||= AgentPresenter.new(
-        Agent.find_by(dsi_uid: session[:dfe_sign_in_uid]),
-      )
+      AgentPresenter.new(Agent.find_by(dsi_uid: session[:dfe_sign_in_uid]))
```
peterdavidhamilton, line 11: Nice one. I feel like the extra step here is to also only decorate the agent before we use them in the view, so I think that means in the `CaseController`; perhaps:

```ruby
def current_agent
  AgentPresenter.new(super)
end
```

We can also then be confident they exist in the show action where we have `current_agent&.id`. What do you think? The reason I say this is `current_agent.respond_to?(:id)` caught my eye.

AndrewCowle: @peterdavidhamilton sorted :)

PR #704, opened 2021-12-07
[156] - Procurement details

d-a-v-e:
## Changes in this PR

- add `support_procurements` and `support_contracts` models
- add savings fields to `support_cases`
- add "Procurement details" tab to the case view
- add procurement details, contract, and savings forms
- add generic date validation type `DateField`

## Screenshots of UI changes

image


app/models/application_record.rb
```diff
@@ -5,4 +5,8 @@
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
   self.implicit_order_column = "created_at"
+
+  def to_h
+    attributes.symbolize_keys
```
kennyevil, line 10: I don't know if it's the right thing to have this method in here or not. I can definitely see the use for it, and you have demonstrated it in this PR. If we do go with this though, you should probably use ```deep_symbolize_keys``` as ```symbolize_keys``` will only get the top layer.

peterdavidhamilton: I can see where you're coming from, I do think a comment explaining its use would help make clear that this is for form data and that should be unaltered.

peterdavidhamilton: @kennyevil could you spare 5 mins to check this out and try the form manually please.

peterdavidhamilton (approved):
I fully expect our QA will suggest more date validations as I have drawn his attention to this opportunity. Hoping we can build out composable granular messages 🤞🏼

ryantk (approved):

ryantk:
Test failures due to https://github.com/DFE-Digital/govuk-formbuilder/releases/tag/v3.0.0
Fix: remove "span" from your css selectors in the test (alerts are no longer spans)

app/models/application_record.rb
kennyevil: Seems to work fine either way, so it's definitely a judgement call. I think some comments and a tight constraint on where it gets used should be fine, unless we're expecting forms with nested params.

PR #706, opened 2021-12-09
Update contract form

PiperS52:
## Changes in this PR

- added controllers, routes, views and form for existing_contract and new_contract

## Screenshots of UI changes

screencapture-localhost-3000-support-cases-eb55c3a8-13de-466f-bb96-571422ee8e43-2021-12-13-10_15_01

Screenshot 2021-12-13 at 10 15 46

Screenshot 2021-12-13 at 10 16 12

# Next steps

- [ ] invalid date validation
- [x] fix currency decimal values
- [x] new contract start date formatting
- [x] remove all instances of “optional”


config/locales/en.yml
```diff
@@ -363,6 +363,34 @@ en:
       flash:
         updated: Successfully changed case category
 
+    case_contracts:
```
peterdavidhamilton, line 366: Minor change but can we make this `case_contract`

peterdavidhamilton:
Excited to see the screenshots

app/controllers/support/cases/existing_contracts_controller.rb
```diff
@@ -0,0 +1,48 @@
+# edit
+# update
+module Support
+  class Cases::ExistingContractsController < Cases::ApplicationController
+    before_action :set_back_url, only: %i[edit update]
+
+    include Concerns::HasDateParams
+
+    def edit
+      @case_contracts_form = CaseContractsForm.new(**current_case.existing_contract.attributes.symbolize_keys)
```
peterdavidhamilton, line 10: Can we introduce a private `contract_params` method in each existing/new controller? Passing a 4 deep chain of methods as a param is an opportunity to extract (and document).

app/controllers/support/cases/existing_contracts_controller.rb
```diff
@@ -0,0 +1,48 @@
+# edit
```
peterdavidhamilton, line 1: Removing these comments and adding one just above the `class` that describe the business concerns is better

app/controllers/support/cases/existing_contracts_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    def validation
+      @validation ||= CaseContractsFormSchema.new.call(**case_contracts_form_params)
+    end
+
+    def set_back_url
+      @back_url = support_case_path(@current_case, anchor: "procurement-details")
+    end
+
+    def case_contracts_form_params
+      params[:case_contracts_form][:started_at] = date_param(:case_contracts_form, :started_at).to_s
```
peterdavidhamilton, line 35: Can we extract the modification of the date param to its own method, this method is doing two things and we can make it a little clearer.

app/controllers/support/cases/new_contracts_controller.rb
```diff
@@ -0,0 +1,48 @@
+# edit
+# update
+module Support
+  class Cases::NewContractsController < Cases::ApplicationController
```
peterdavidhamilton, line 4: The new and existing controllers are very similar. After the refactoring mentioned above do you think it could be beneficial to DRY this up?

config/locales/en.yml
```diff
@@ -363,6 +363,34 @@ en:
       flash:
         updated: Successfully changed case category
 
+    case_contract:
+      existing:
+        edit:
```
peterdavidhamilton, line 368: Does intending this extra `edit` step help us?

peterdavidhamilton:
There is an inconsistency with the change links; a stray hyphen.

app/forms/support/case_contracts_form.rb
```diff
@@ -8,5 +8,20 @@ class CaseContractsForm
     option :ended_at, optional: true
     option :spend, optional: true
     option :duration, optional: true
```
peterdavidhamilton, line 10: This is a situation where we can coerce the value directly, a custom type is the best way to do this in my opinion. Let's get some time to discuss these libraries @d-a-v-e @EdwinKruglov @PiperS52

d-a-v-e (approved):

config/locales/en.yml
PiperS52: It shows that it's for the edit page, and is a pattern which seems to be used in the rest of the file

spec/features/support/case_summary_spec.rb
```diff
@@ -97,6 +97,8 @@
   end
 
   context "when the case is created" do
+    # let(:support_case) { create(:support_case, :opened, :with_documents, agent: agent) }
+
```
kennyevil, line 101: Think this could probably be cut out if it isn't being used

app/presenters/support/contract_presenter.rb
```diff
@@ -0,0 +1,37 @@
...
+    # @return [String]
+    def duration
+      super&.inspect || "-"
+    end
+
+    # @return [String]
+    def started_at
+      return "-" unless super
+
+      super.strftime("%e %B %Y")
```
peterdavidhamilton, line 27: `date.formats.presenter`

peterdavidhamilton: `#date_format`

app/controllers/support/cases/contracts_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+
+        redirect_to @back_url, notice: I18n.t("support.case_contract.flash.updated")
+      else
+        edit_view
+      end
+    end
+
+  private
+
+    def edit_view
```
peterdavidhamilton, line 26: instead of having `render` in a method could we refactor so we just have a method that returns the view name, then use that with string interpolation

peterdavidhamilton: Guess you could even, snakecase-ize and pluralize the class name....

PR #709, opened 2021-12-13
157 - Record case open times for KPIs

ryantk:
## Changes in this PR

Added more KPI logging in to match requirements in ticket https://trello.com/c/vuByBNy7/157-n8-procops-record-case-time-data-for-kpi-reporting

### DIFF
--- a/spec/services/support/record_action_spec.rb
+++ b/spec/services/support/record_action_spec.rb
@@ -8,7 +8,13 @@ RSpec.describe Support::RecordAction do
     let(:action) { "open_case" }
 
     it "records the action in the database" do
-      expect { service.call }.to change { Support::ActivityLogItem.count }.by(1)
+      service.call
+      expect(Support::ActivityLogItem.where(action: "open_case", support_case_id: support_case.id).count).to eq(1)
+    end
+
+    it "records a 'case_modified' action for every action called" do
+      service.call
+      expect(Support::ActivityLogItem.where(action: "case_modified", support_case_id: support_case.id).count).to eq(1)
     end
 
     context "when the action has additional parameters specified" do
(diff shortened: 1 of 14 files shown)


ryantk:
[N8: ProcOps - Record Case Time data for KPI reporting](https://trello.com/c/vuByBNy7/157-n8-procops-record-case-time-data-for-kpi-reporting)

spec/services/support/record_action_spec.rb
```diff
@@ -8,7 +8,13 @@
     let(:action) { "open_case" }
 
     it "records the action in the database" do
-      expect { service.call }.to change { Support::ActivityLogItem.count }.by(1)
+      service.call
+      expect(Support::ActivityLogItem.where(action: "open_case", support_case_id: support_case.id).count).to eq(1)
```
kennyevil, line 12: I prefer using ```to change``` on these kinds of specs as it  more confidently shows that the thing you've called is responsible for adding the new item(s), but on the other hand with isolated specs this works just as well.

kennyevil (approved):

spec/services/support/record_action_spec.rb
ryantk: I like to use them in a different scenario, for this I wanted to know not only that I had created an extra ActivityLogItem but that it had the correct case and action on it.

Perhaps a 3rd way could have been

```
expect { service.call }. to change { Support::ActivityLogItem.where(action: "open_case", support_case_id: support_case.id).count }.by(1)
```

kennyevil: Oh sorry, I should have been clearer here. I meant using it in the way you've put at the end there and you're right that it's more explicit in making sure that the test is not only passing, but passing in the right way.

peterdavidhamilton (approved):

PR #719, opened 2021-12-17
Log when statements are acknowledged

PiperS52:
## Changes in this PR

- amended RecordAction to capture when a step is answered and no user input required


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_11_29_120610) do
```
kennyevil, line 13: Watch out for schema changes leaking back into your commits where you don't mean for them to.

app/controllers/answers_controller.rb
```diff
@@ -32,18 +32,7 @@ def create
       end
 
     # TODO: refactor to a private #record_answer method that accepts the action string
```
kennyevil, line 34: You can remove this TODO now, because you've done the work

kennyevil (approved):
Looks good. Just a TODO to remove and this is good to go.

EdwinKruglov (approved):

PR #191, opened 2021-02-25
Update warning message for an incomplete specification

LornaHarwood:
## Changes in this PR

This warning message has been changed to help be more explicit in it's messaging.

## Screenshots of UI changes

Screen Shot 2021-02-25 at 10 59 31

PR #722, opened 2021-12-17
Feature/209 received emails appear in case history

kennyevil:
## Changes in this PR

Adds Support::Interaction when a Support::Email is assigned to a case

PR #724, opened 2021-12-17
#191 - Synchronise emails from shared inbox

ryantk:
## Changes in this PR

Service and job to synchronise emails from the defined shared inbox into our database.

### DIFF
--- a/db/migrate/20211217092216_add_extra_outlook_fields_to_support_email.rb
+++ b/db/migrate/20211217092216_add_extra_outlook_fields_to_support_email.rb
@@ -1,0 +1,10 @@
+class AddExtraOutlookFieldsToSupportEmail < ActiveRecord::Migration[6.1]
+  def change
+    add_column :support_emails, :outlook_id, :string
+    add_column :support_emails, :outlook_conversation_id, :string
+    add_column :support_emails, :is_read, :boolean, default: false
+    add_column :support_emails, :is_draft, :boolean, default: false
+    add_column :support_emails, :has_attachments, :boolean, default: false
+    add_column :support_emails, :body_preview, :text
+  end
+end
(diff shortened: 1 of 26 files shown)


ryantk:
[System synchronises emails and stores in database](https://trello.com/c/pjJu8D6w/191-system-synchronises-emails-and-stores-in-database)

db/migrate/20211217092216_add_extra_outlook_fields_to_support_email.rb
```diff
@@ -0,0 +1,10 @@
+class AddExtraOutlookFieldsToSupportEmail < ActiveRecord::Migration[6.1]
+  def change
+    add_column :support_emails, :outlook_id, :string
+    add_column :support_emails, :outlook_conversation_id, :string
```
kennyevil, line 4: There's already a ```conversation_id``` field. It might be  wroth removing that one as this is more explicitly named.

db/migrate/20211217092216_add_extra_outlook_fields_to_support_email.rb
```diff
@@ -0,0 +1,10 @@
...
+    add_column :support_emails, :is_read, :boolean, default: false
+    add_column :support_emails, :is_draft, :boolean, default: false
+    add_column :support_emails, :has_attachments, :boolean, default: false
```
kennyevil, line 7: For #722 we'll need to know how many attachments there are. It might be worth saving some of the attachments information (eg name) in a jsonb field.

kennyevil (changes requested):
Looks great. Just a couple of small changes to do with some requirements outside of this ticket. They could be done on the other branch instead, but thought it was worth flagging up.

PR #732, opened 2021-12-23
Hide internal team members from assignable case workers

EdwinKruglov:
## Changes in this PR

- add `internal` field to `agent`
- add `caseworkers` scope to `agent`
- filter out internal team members on the case assignment page


ryantk (approved):

EdwinKruglov:
[Hide internal team members from assignable case workers](https://trello.com/c/Z7M1hEfm/86-hide-internal-team-members-from-assignable-case-workers)

PR #746, opened 2022-01-07
Case source being set incorrectly on create a case

PiperS52:
## Changes in this PR

- add check for nw_hub and sw_hub according to identification number


PiperS52:
[Case Source being set incorrectly on Create a case](https://trello.com/c/lOPUhSTl/210-case-source-being-set-incorrectly-on-create-a-case)

ryantk (approved):

PR #760, opened 2022-01-12
FaF support request start page

PiperS52:
## Changes in this PR

- added controller with route and view

## Screenshots of UI changes
[screencapture-localhost-3000-procurement-support-new-2022-01-12-15_21_21.pdf](https://github.com/DFE-Digital/buy-for-your-school/files/7855655/screencapture-localhost-3000-procurement-support-new-2022-01-12-15_21_21.pdf)

PR #762, opened 2022-01-12
Feature/219 faf do you have dfe sign in

d-a-v-e:
## Changes in this PR

* step logic and faf form/schema
* dsi_or_search route, action and view

## Screenshots of UI changes
Screenshot 2022-01-12 at 18 09 50


d-a-v-e:
(FYI: this is branched off of #760 so currently has some of @PiperS52 's work)

peterdavidhamilton (changes requested):
Can we adopt inheritance here and copy the base classes we have for schema and form that we set up in the `support` namespace first.

app/views/fafs/steps/_step_2.html.erb
```diff
@@ -0,0 +1,8 @@
+<span class="govuk-caption-l">
+  step 2
+</span>
+<h1 class="govuk-heading-l">
+  Step 2
+</h1>
+
+<br>
```
d-a-v-e, line 8: (this is a placeholder)

EdwinKruglov:
@d-a-v-e I can see the `Create a new DfE Sign-in account` link in the screenshot, but I don't see it in the code (except for the string in the locale file)?

peterdavidhamilton (approved):

PR #767, opened 2022-01-13
Error on phone number validation

PiperS52:
## Changes in this PR

- amended rule for phone number validation in hub cases as follows; (i) permits empty string in event of no entry, (ii) 0 followed by 9 or 10 digits, OR +44 followed by 9 or 10 digits, (iii) no spaces at any point in the number
- max_digits? captures the + in +44, so value of 13, although message correctly displays 12, with number entered greater than 12 digits

### DIFF
--- a/app/forms/support/case_hub_migration_form_schema.rb
+++ b/app/forms/support/case_hub_migration_form_schema.rb
@@ -25,7 +25,7 @@ module Support
     end
 
     # TODO: custom macro for phone number validation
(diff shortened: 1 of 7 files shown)


peterdavidhamilton:
[Error on phone number validation](https://trello.com/c/n4PixrBt/258-error-on-phone-number-validation)

ryantk (approved):

PR #777, opened 2022-01-17
FaF support request - how can we help? page

PiperS52:
## Changes in this PR

- updated fafs controller, adding view for 'step 5' ('How can we help?') page
- message_body posting to support_requests db table, although this is to change to framework_request model, once ticket 239 is in place

## Screenshots of UI changes
Screenshot 2022-01-17 at 16 16 40


d-a-v-e (approved):

spec/features/self-serve/faf/user_query_page_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+      click_continue
+
+      expect(FrameworkRequest.count).to eq(1)
+    end
+  end
+
+  context "when the user is not signed in" do
+    before do
+      visit "/procurement-support/new"
+      find("label", text: I18n.t("faf.dsi_or_search.radios.no_dsi")).click
```
peterdavidhamilton, line 41: Absolutely tiny niggle but can we have a plain string in here not a locale lookup please.

spec/support/journey_helpers.rb
```diff
@@ -10,6 +10,8 @@ def click_start
   # generic.button.next
   def click_continue
```
peterdavidhamilton, line 11: Interesting because we were looking at which things could become helpers... do you feel that this is helpful still or not?

PR #778, opened 2022-01-17
Add "about your school" step for FaF

kennyevil:
## Changes in this PR

Add a page for selecting the school associated with the Find a Framework journey support request.

## Screenshots of UI changes

Screenshot 2022-01-19 at 11-45-49 Which school are you buying for
Screenshot 2022-01-19 at 11-54-03 Which school are you buying for


EdwinKruglov:
@kennyevil I think the only thing that's missing is the back link into the previous step?

d-a-v-e (approved):

EdwinKruglov (approved):

d-a-v-e (approved):
This works well, a couple thoughts:

* The edit views won't have backlinks (i.e. when someone has made it to the final "check your answers" page and then clicked "change" on school or description. It's not immediately obvious where the backlink should go in this context, but for me, it would back sense to go back to the "check your answers" page;

* Minor aesthetic critique: the backlink in the prototype sits on top of the main div, leaving a smaller gap between it and the header.

db/schema.rb
```diff
@@ -72,7 +72,7 @@
     t.datetime "created_at", precision: 6, default: -> { "CURRENT_TIMESTAMP" }, null: false
     t.datetime "updated_at", precision: 6, default: -> { "CURRENT_TIMESTAMP" }, null: false
     t.integer "journeys_count"
-    t.string "slug", null: false
+    t.string "slug"
```
PiperS52, line 75: Looks like you're also getting the slightly strange schema changes - not just me

PiperS52 (approved):

db/schema.rb
kennyevil: >     * The edit views won't have backlinks (i.e. when someone has made it to the final "check your answers" page and then clicked "change" on school or description. It's not immediately obvious where the backlink should go in this context, but for me, it would back sense to go back to the "check your answers" page;

I can probably just add in what we have for the ```support_requests``` form, ie just a generic back link. That should sort it out 

>     * Minor aesthetic critique: the backlink in the prototype sits on top of the main div, leaving a smaller gap between it and the header.

I know, but the only way to actually get it in there like that would be to put it in breadcrumbs partial and the logic for that could get messy. This is where the backlink sits in the ```support_requests``` form though, so I think placing it here for this form keeps it consistent.

kennyevil: > Looks like you're also getting the slightly strange schema changes - not just me

Going to do a quick double check. I thought I'd kept any db changes out of my PR.

PR #788, opened 2022-01-20
Keep agent assigned when resolving a case

ryantk:
## Changes in this PR

When resolving a case we need to keep the agent assigned as before.


ryantk:
[#26 Case remains assigned to me when I mark as resolved (funct change)](https://trello.com/c/LmLoBHd2/278-26-case-remains-assigned-to-me-when-i-mark-as-resolved-funct-change)

PR #792, opened 2022-01-20
phone number tests activated and copied into another spec

shafiqahmedkhan:
## Changes in this PR

- activated existing tests for phone number validation
- added above tests to another spec file


spec/features/self-serve/support_requests/create_support_request_spec.rb
```diff
@@ -60,7 +60,7 @@
     end
 
     # step 1
-    xit "asks for a phone number" do
+    it "asks for a phone number" do
```
peterdavidhamilton, line 63: We need these disabled because this form doesn't currently ask for a phone number (it will again very soon, hence why we opted to disable not remove the specs). The pipeline hasnt revealed that to you though because the linting error came first.

spec/features/self-serve/support_requests/create_support_request_spec.rb
```diff
@@ -95,6 +101,22 @@
...
 
+      it "international code" do
+        fill_in "support_form[phone_number]", with: "+3551234567"
+        click_continue
+        expect(find("h2.govuk-error-summary__title")).to have_text "There is a problem"
+        expect(page).to have_link "Your phone number must have no spaces and begin with a 0 or +44, with a minimum of 10 and maximum 12 digits", href: "#support-form-phone-number-field-error"
+        expect(find(".govuk-error-message")).to have_text "Your phone number must have no spaces and begin with a 0 or +44, with a minimum of 10 and maximum 12 digits"
+      end
+
+      it "white space in between the phone number" do
```
peterdavidhamilton, line 112: I think matching the syntax and going with `(white space)` as the string would be good.

spec/features/support/case_migration_spec.rb
```diff
@@ -95,6 +95,76 @@
...
+
+      it "validates a phone number (valid +44)" do
+        fill_in "support_form[phone_number]", with: "+441234567890"
+        click_continue
+        expect(find("legend.govuk-fieldset__legend--l")).to have_text "What are you buying?"
+      end
+    end
+
+    context "with invalid data it validates a phone number" do
+      # step 1
```
peterdavidhamilton, line 115: This is not a multi-step form so this comment could be misleading

peterdavidhamilton (approved):

PR #801, opened 2022-01-24
Organisations on a case are optional / Can be updated

ryantk:
## Changes in this PR

Ensuring that case organisations are optional so that auto creation of cases can succeed in further work.
When an organisation is missing then the contact email displays instead.

Ability to search, select and update an organisation on a case.

## Screenshots of UI changes

Screenshot 2022-01-24 at 11 30 39

Screenshot 2022-01-24 at 14 31 40

Screenshot 2022-01-24 at 14 31 51

Screenshot 2022-01-24 at 14 32 01


ryantk:
[#383 Organisation field to be optional](https://trello.com/c/y8ZVsfbe/270-383-organisation-field-to-be-optional)

kennyevil (approved):

PR #804, opened 2022-01-24
Feature/faf dsi entry

d-a-v-e:
## Changes in this PR

- move some logic into the form interface
- determine the initial state in the `new` action
- rename to follow the conventions used in the original support request form

## Next steps

- correct formatting on non-dsi step pages


app/controllers/framework_requests_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+  end
+
+private
+
+  # @return [FrameworkSupportForm] form object populated with validation messages
+  def form
+    FrameworkSupportForm.new(
+      step: form_params[:step],
+
+      dsi: !current_user.guest?, # ensures a boolean
```
d-a-v-e, line 87: Do we know if we can/should expect users to use the "no dsi" path even if they are signed in? perhaps to submit a request for a school not associated with them?

app/controllers/framework_requests_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+        school_urn: current_user.school_urn,  # (step 4)
+        # message (step 5)
+      }
+    end
+  end
+
+  # @return [FrameworkSupportForm] form object with updated step number if validation successful
+  def advance_form
+    if @framework_support_form.step == 2 && current_user.school_urn
+      @framework_support_form.advance!(2)
```
d-a-v-e, line 147: I find the addition a little confusing. Feels like when you pass in an argument to `advance!` it should go to that step rather than adding the value to the current step count.

app/controllers/framework_requests_controller.rb
```diff
@@ -0,0 +1,168 @@
...
+  end
+
+  # @return [FrameworkSupportForm] form object with reverted step number
+  def revert_form
+    # binding.pry
+
+    # last step, dsi and inferred school
+    if (@framework_support_form.step == 5) && !current_user.guest? && !!current_user.school_urn
+
+      @framework_support_form.back!(4) # go to step 1
```
d-a-v-e, line 160: same comment about the `.back!` method. perhaps it would be simpler to have separate methods `back_to()` and `advance_to()`

app/views/framework_request_submissions/show.html.erb
```diff
@@ -0,0 +1,49 @@
...
+      </p>  
+
+      <h1 class="govuk-heading-m">
+        <%= I18n.t("faf_submissions.header.what_you_can_do_next") %>
+      </h1>
+      <ul class="govuk-list govuk-list--bullet">
+        <li>
+          <%=
+            I18n.t("faf_submissions.what_you_can_do_next.read_buying_procedures",
+              read_buying_procedures_link: link_to(I18n.t("faf_submissions.what_you_can_do_next.read_buying_procedures_link"), "https://buy-for-your-school-staging.herokuapp.com/beta/phase-6/gov/buying-procedures-for-schools", class: "govuk-link")
```
d-a-v-e, line 35: E  N c  e  p  t  i  o  n

app/controllers/framework_requests_controller.rb
peterdavidhamilton: Yes, with the previous form the stepping logic only needed that interface. Now as we leap around I think we see the need for a `#go_to` method on the form

app/controllers/framework_requests_controller.rb
```diff
@@ -0,0 +1,170 @@
+# :nocov:
+class FrameworkRequestsController < ApplicationController
+  skip_before_action :authenticate_user!
+  before_action :current_user
+  before_action :framework_request, only: %i[edit show update]
+
+  def index
+    session[:faf_referer] = request.referer || "none"
```
peterdavidhamilton, line 8: @AndrewCowle out of necessity we made some changes but then spotted that when a support request is started directly the referer is missing. Use this knowledge accordingly.

app/controllers/framework_requests_controller.rb
peterdavidhamilton: @d-a-v-e DSI provides verified users so my understanding is no, we make them seek out an org approver instead

AndrewCowle (approved):

PR #805, opened 2022-01-24
Auto creation of cases from incoming emails

ryantk:
## Changes in this PR

Cases are automatically created by an incoming email when that email cannot be assigned to an existing case.


ryantk:
[#383 Auto create (no case ref) a case from an inbound email that does not contain a case reference](https://trello.com/c/406pnAOK/266-383-auto-create-no-case-ref-a-case-from-an-inbound-email-that-does-not-contain-a-case-reference)

kennyevil (approved):

PR #816, opened 2022-01-26
Feature/386 add establishment groups

kennyevil:
## Changes in this PR

Adds ```EstablishmentGroup``` and ```EstablishmentGroupType``` models, services to seed the database with group information pulled from GIAS and a scheduled job to refresh the EstablishmentGroup data  once a month.


kennyevil:
[#386 FaF Entry Point - None DSI - Seed groups from GIAS](https://trello.com/c/j3RbfXWR/268-386-faf-entry-point-none-dsi-seed-groups-from-gias)

peterdavidhamilton (approved):
Lot's of copy/paste so ripe for DRYing up I'm sure. The job has a comment that mentions the number of records which is copied from establishments, how many group records do we get? If you can tweak that and the other comments that are misnamed (`GroupType` for example) it's all good.

PR #820, opened 2022-01-26
Add case reference to personalisation params on non-templated emails

kennyevil:
## Changes in this PR

Sends the case reference as a personalised parameter for non-templated emails. 

## Next steps

Basic email template will have to be updated to include the reference in the subject and email body.

### DIFF
--- a/app/controllers/support/cases/emails/content_controller.rb
+++ b/app/controllers/support/cases/emails/content_controller.rb
@@ -63,12 +63,9 @@ module Support
     def personalisation
       contact = CaseContactPresenter.new(@current_case)
 
-      personalisation_params = { first_name: contact.first_name, last_name: contact.last_name, from_name: current_agent.full_name }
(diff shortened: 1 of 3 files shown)


app/controllers/support/cases/emails/content_controller.rb
```diff
@@ -65,7 +65,10 @@ def personalisation
 
       personalisation_params = { first_name: contact.first_name, last_name: contact.last_name, from_name: current_agent.full_name }
 
-      personalisation_params[:text] = basic_email_body if basic_template?
+      if basic_template?
+        personalisation_params[:text] = basic_email_body
+        personalisation_params[:reference] = @current_case.ref
```
ryantk, line 70: Can you make it so the reference is visible for all templates? We are going to eventually have it showing on every template.

ryantk (changes requested):

app/controllers/support/cases/emails/content_controller.rb
kennyevil: It's now been added to the personalisation params for all emails

ryantk (approved):

PR #828, opened 2022-01-27
List of All Cases in CM is showing the creation date in the 'Last updated' column

PiperS52:
## Changes in this PR

- added enum event types (procurement_updated, existing_contract_updated, new_contract_updated, savings_updated) to interaction model
- created new concern HasInteraction
- cases displaying according to updated_at (interaction) with oldest at top

## Screenshots of UI changes

Screenshot 2022-02-01 at 18 09 28

Screenshot 2022-02-01 at 18 09 46

screencapture-localhost-3000-support-cases-2022-02-01-18_08_46


peterdavidhamilton:
[List of All Cases in CM is showing the creation date in the 'Last updated' column](https://trello.com/c/FviFhp4F/290-list-of-all-cases-in-cm-is-showing-the-creation-date-in-the-last-updated-column)

app/presenters/support/case_presenter.rb
```diff
@@ -33,7 +33,7 @@ def received_at
 
     # @return [String] 30 January 2000 at 12:00
     def last_updated_at
```
peterdavidhamilton, line 35: @PiperS52 can't we just do away with this method now? @AndrewCowle can you sanity check this deletion because I think remember it going in.

AndrewCowle: If an agents adds a note to a case it would create an interaction, but would this touch the case (updated at) - a true reflection of updates on a case?

ryantk: @PiperS52 if you can confirm what @AndrewCowle says and remove "last_updated_at" and just use "updated_at" instead like @peterdavidhamilton  says I will approve :)

PiperS52: Good spot. It raises the question of what is to be captured in an 'update'? It seems a change to the 'procurement details' tab is not currently captured either

ryantk: @PiperS52 might be worth looking at https://apidock.com/rails/ActiveRecord/Persistence/touch

kennyevil (approved):

AndrewCowle (approved):

PR #845, opened 2022-02-03
List of All Cases in CM is showing the creation date in the 'Last updated' column

EdwinKruglov:
## Changes in this PR

- sort cases by `last_update_at` _date_, rather than its string representation


EdwinKruglov:
[List of All Cases in CM is showing the creation date in the 'Last updated' column](https://trello.com/c/FviFhp4F/290-list-of-all-cases-in-cm-is-showing-the-creation-date-in-the-last-updated-column)

EdwinKruglov:
[Last Updated date/time not updating when add Note or send email on a case](https://trello.com/c/CVkW0O2y/85-last-updated-date-time-not-updating-when-add-note-or-send-email-on-a-case)

kennyevil (approved):

AndrewCowle (approved):

PR #851, opened 2022-02-07
Groups + Organisations search on case

ryantk:
## Changes in this PR

- Database View to enable searching of organisations and establishment groups all at once
- Updates to autocomplete to enable hidden fields to send more data in forms
- Updated create a case to search for groups and organisations
- Refactored case organisation relationship to be polymorphic (Support::Organisation and Support::EstablishmentGroup)
- Refactored case search to use a view (pg_search does not work with polymorphic relations FYI @d-a-v-e) 

## Screenshots of UI changes

Screenshot 2022-02-14 at 16 12 38
Screenshot 2022-02-14 at 16 12 50


ryantk:
[Add groups to the orgs search functionality in Case Management](https://trello.com/c/6AaCpay3/343-add-groups-to-the-orgs-search-functionality-in-case-management)

peterdavidhamilton:
Really like the JS tidy up @ryantk. 

@d-a-v-e if we go with views and `scenic` we should have an ARD to record these choices.

@kennyevil you had thoughts on renaming some uses of `organisation` to `establishment` as we have a better understanding now of what we mean in different contexts; would this be an opportunity to incorporate that?
The `organisation` a `case` can be joined to and a person can be affiliated with is either an `establishment` or an `establishment_group`, I feel like we're not completely consistent yet.

db/views/support_case_searches_v01.sql
```diff
@@ -0,0 +1,16 @@
...
+  ses.ukprn AS organisation_ukprn,
+  sa.first_name || ' ' || sa.last_name AS agent_name,
+  sa.first_name AS agent_first_name,
+  sa.last_name AS agent_last_name
+FROM
+  support_cases sc
+LEFT JOIN support_agents sa
+  ON sa.id = sc.agent_id
+LEFT JOIN support_establishment_searches ses
+  ON sc.organisation_id = ses.id AND sc.organisation_type = ses.source;
```
d-a-v-e, line 16: Nice one, cheers @ryantk

doc/components/autocomplete.md
```diff
@@ -61,4 +66,37 @@ It is worth utilising the rails helpers to build a url:
...
+If you want to send more values to the server, you can do so with hidden fields.
+
+||Description|
+|-|-|
+|key|This is the name of the hidden field that will be generated in the form|
+|value|The hidden field will be populated with this field from the response json|
+
+```
+Given an API response of
+[{id: "1", name: "Joe Bloggs", age: 20}, {id: "2", "Jane Bloggs", age: 32}]
```
d-a-v-e, line 84: missing a key in the example
```suggestion
[{id: "1", name: "Joe Bloggs", age: 20}, {id: "2",  name: "Jane Bloggs", age: 32}]
```

db/migrate/20220211162056_create_support_case_searches.rb
```diff
@@ -0,0 +1,5 @@
+class CreateSupportCaseSearches < ActiveRecord::Migration[6.1]
+  def change
+    create_view :support_case_searches
```
d-a-v-e, line 3: might be worth making this materialized in the future for performance, but no need to do so prematurely. Is there anything in place (yet) for monitoring performance? @peterdavidhamilton

d-a-v-e (approved):
thanks, @ryantk. Some great improvements here. Introducing`scenic` makes good sense 👍

ryantk:
@peterdavidhamilton @d-a-v-e Thanks for delving into this PR, it was quite far reaching for just a simple search box.

db/migrate/20220211162056_create_support_case_searches.rb
peterdavidhamilton: This is something to have on the radar given @shafiqahmedkhan has recently raised JS response time and @ryantk, quite rightly, is keen to improve the granularity of our pipeline.

kennyevil (approved):

PR #859, opened 2022-02-09
Correct objects being used for pagination

kennyevil:
## Changes in this PR

Ensures correct cases are being used for pagination on case maangement


d-a-v-e (approved):
I've re-run the specs as the one failing is unrelated and intermittent.

PR #863, opened 2022-02-10
FaF Support Request - Modify eligibility check so MATs are eligible

EdwinKruglov:
## Changes in this PR

- allow users belonging to groups (MATs and SATs) to use the service

### DIFF
--- a/app/services/create_user.rb
+++ b/app/services/create_user.rb
@@ -41,20 +41,26 @@ class CreateUser
 
 private
 
+  # @param [String] field
+  #
   # @return [Array<Integer>] the current user's org types
-  def establishment_types
-    orgs.map { |org| org.dig("type", "id").to_i }
-  end
-
-  # @return [Array<Integer>] the current user's org categories
-  def establishment_categories
-    orgs.map { |org| org.dig("category", "id").to_i }
+  def establishment_types(field)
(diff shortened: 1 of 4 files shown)


EdwinKruglov:
[#234 FaF Support Request - Modify eligibility check so MATs are eligible](https://trello.com/c/FsL0CRnT/220-234-faf-support-request-modify-eligibility-check-so-mats-are-eligible)

app/services/create_user.rb
```diff
@@ -46,9 +46,15 @@ def establishment_types
     orgs.map { |org| org.dig("type", "id").to_i }
   end
 
+  # @return [Array<Integer>] the current user's org categories
+  def establishment_categories
```
peterdavidhamilton, line 50: I think you should refactor the existing `org` query method to cater for querying `type` or `category` depending on a param. Then have the `supported?` predicate call two new methods that use it i.e. `supported_establishment? || supported_establishment_group?` (however these names are CM aware so in this context perhaps `valid_school? || valid_trust?`)

peterdavidhamilton (changes requested):

peterdavidhamilton (approved):

PiperS52 (approved):

PR #865, opened 2022-02-11
Add page params for different case tabs

kennyevil:
## Changes in this PR

Tabs for the cases page will now use individual page params so that each tab remains independent of each other.


peterdavidhamilton (approved):
Can you drop a `FIXME` in there about why bullet is disabled to make it easier to find and flag this before it gets forgotten about.

peterdavidhamilton:
bitmoji

AndrewCowle (approved):

PR #866, opened 2022-02-11
Add group and school search to the non-dsi faf journey

kennyevil:
## Changes in this PR

Step 3 for the non-dsi faf journey is now choosing School or Group and step 4 is the check your group/school page.

PR #864, opened 2022-02-10
FaF MAT n Trusts - Add 'School type' to the None DSI CYA page

PiperS52:
## Changes in this PR

- School or group name, as well as school/group type displaying on CYA page

### After
Screenshot 2022-02-10 at 16 43 22

Screenshot 2022-02-10 at 16 34 14

PR #855, opened 2022-02-08
FaF support request - updates to DfE signed in journey for MATs

PiperS52:
## Changes in this PR

- update content for kick-out page 
- add trusts to the DSI journey school page
- add `(MAT)` string to trusts listed on DSI journey school page

### Before

image

### After

image
image

PR #850, opened 2022-02-07
FaF MAT n Feds - Group search page

EdwinKruglov:
## Changes in this PR

- add "Search for a group or trust" page to the FaF journey
- add `group_uid` to `framework_requests`
- update "Search for a school" help text

### Before

image

### After

image
image

PR #873, opened 2022-02-14
Addition of groups to the FaF journey

EdwinKruglov:
## Changes in this PR

- Make navigation changes (#866)
- Add "Is this the group or trust?" page (#847)
- Add trusts to DSI journey and update kick-out page (#855)
- Add school type to non-DSI CYA page (#864)
- Add group search page (#850)
- Add "Is it a school or group?" page (#849)
- Make trust members eligible to access the service (#863)


kennyevil (approved):

peterdavidhamilton:
@EdwinKruglov a reminder that work @kennyevil has done on forms give us the option to use `position?` and `go_to!`, if these methods improve readability and make it easier to understand the logic please make sure we use them in the controller.

app/controllers/framework_requests_controller.rb
```diff
@@ -153,27 +172,51 @@ def conditional_form_params
     else
       {
         dsi: true,                            # (step 1)
-        first_name: current_user.first_name,  # (step 2)
-        last_name: current_user.last_name,    # (step 2)
-        email: current_user.email,            # (step 3)
-        school_urn: current_user.school_urn,  # (step 4)
-        # message (step 6)
+        group: false,                         # (step 2)
```
peterdavidhamilton, line 175: Can we get these in order

app/forms/framework_support_form_schema.rb
```diff
@@ -33,11 +45,31 @@ class FrameworkSupportFormSchema < Schema
   end
 
   rule(:school_urn) do
-    key.failure(:missing) if key? && Support::Organisation.find_by(urn: value.split(" - ").first).nil?
+    # validate individual school_urn field only in the non-dsi journey
+    key.failure(:missing) if key? && !values[:dsi] && !values[:group] && !org_exists?(value.split(" - ").first, false)
```
peterdavidhamilton, line 49: What do you think about moving the `id` value coercion into `org_exists?`? 
`group` param is a boolean (which doesn't have a YARD btw) what other way could we do this?

app/presenters/framework_request_presenter.rb
```diff
@@ -22,4 +22,17 @@ def user
...
     user_id.present?
   end
+
+  def group_name
+    return I18n.t("support.case_categorisations.label.none") if group_uid.blank?
+
+    Support::EstablishmentGroup.find_by(uid: group_uid).name
+  end
+
+  def group_type
```
peterdavidhamilton, line 32: Here we have ended up crossing over the namespace. This is a value we could have returned from a single lookup using that proposed service object. Can you add a `TODO` in here to flag that this is something that needs extra thought please.

peterdavidhamilton (changes requested):
[Should](https://github.com/DFE-Digital/buy-for-your-school/blob/ce791aac84f6eff3b5d77cb898f48e7c449c229c/app/views/framework_requests/steps/_form.html.erb#L39) use `position?` instead. Made a few other comments.

app/controllers/framework_requests_controller.rb
```diff
@@ -153,27 +172,51 @@ def conditional_form_params
...
+  # Extract the group UID from the format "uid - group name"
+  # @example
+  #   "1000 - Group #1" -> "1000"
+  #
+  # @return [String, nil]
+  def group_uid
+    form_params[:group_uid]&.split(" - ")&.first
+  end
+
+  def forget_org
```
peterdavidhamilton, line 215: This is a prime candidate for a refactor. Notice how the `controller` is talking to the `form` object 6 times. The `form` is the one that needs to be told what to do and is capable of working it out on its own. If `forget_org` were moved into the `form` object the code gets simpler and the controller gets cleaner.

peterdavidhamilton (changes requested):

app/forms/framework_support_form.rb
```diff
@@ -86,5 +86,14 @@ def forget_group!
   def multiple_schools?
     instance_variable_get :@group
   end
+
+  # @return [nil]
```
peterdavidhamilton, line 90: Nice one @EdwinKruglov. Can these be a 2 liner, maybe with guard clause? Also, every method is currently public but I'm sure that's not the case anymore. @d-a-v-e is looking at getting the feature coverage of FaF up, but can you get unit test coverage for this form class in this branch, currently the whole file is being skipped.

peterdavidhamilton (changes requested):

peterdavidhamilton (approved):

PR #874, opened 2022-02-14
FaF school type change link fix

PiperS52:
## Changes in this PR

- No DSI journey - 'school type' change link now permits user to then select school/group after changing selection at step 2
- DSI journey - 'school type' change link no longer appearing.

### DIFF
--- a/app/controllers/framework_requests_controller.rb
+++ b/app/controllers/framework_requests_controller.rb
@@ -111,9 +111,15 @@ class FrameworkRequestsController < ApplicationController
       #   @support_form.advance!
       #   render :edit
       # else
-      framework_request.update!(**framework_request.attributes.symbolize_keys, **@framework_support_form.to_h.merge(school_urn: urn, group_uid: group_uid))
-      redirect_to framework_request_path(framework_request), notice: I18n.t("support_request.flash.updated")
-      # end
+      if @framework_support_form.step == 2
+        @framework_support_form.advance!
+        render :edit
+      else
+
+        framework_request.update!(**framework_request.attributes.symbolize_keys, **@framework_support_form.to_h.merge(school_urn: urn, group_uid: group_uid))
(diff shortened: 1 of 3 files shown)


kennyevil (approved):

app/controllers/framework_requests_controller.rb
```diff
@@ -111,9 +111,15 @@ def update
       #   @support_form.advance!
       #   render :edit
       # else
-      framework_request.update!(**framework_request.attributes.symbolize_keys, **@framework_support_form.to_h.merge(school_urn: urn, group_uid: group_uid))
-      redirect_to framework_request_path(framework_request), notice: I18n.t("support_request.flash.updated")
-      # end
+      if @framework_support_form.step == 2
```
peterdavidhamilton, line 114: Our form has evolved and we now have an improved interface to check the `position`.

app/controllers/framework_requests_controller.rb
```diff
@@ -111,9 +111,15 @@ def update
...
+        @framework_support_form.advance!
```
peterdavidhamilton, line 115: We can also `goto` a specific step now rather than advance, which helps be more explicit but should be accompanied by a helpful note because we are still using numbers and not named steps.

app/controllers/framework_requests_controller.rb
```diff
@@ -111,9 +111,15 @@ def update
...
+        render :edit
+      else
+
+        framework_request.update!(**framework_request.attributes.symbolize_keys, **@framework_support_form.to_h.merge(school_urn: urn, group_uid: group_uid))
+        redirect_to framework_request_path(framework_request), notice: I18n.t("support_request.flash.updated")
+        # end
```
peterdavidhamilton, line 121: Delete this too.

peterdavidhamilton (changes requested):
Choice of methods and opportunity to improve documentation

peterdavidhamilton (approved):

PR #877, opened 2022-02-14
Change order of cases in supported

kennyevil:
## Changes in this PR

Cases in supported are now ordered by ``` created_at: :desc```, ie newest cases first, rather than by the date of the latest interaction.


kennyevil:
[Change order of cases in case lists to be by Case number.](https://trello.com/c/u6b2YMVl/369-change-order-of-cases-in-case-lists-to-be-by-case-number)

ryantk (approved):

peterdavidhamilton (approved):

PR #892, opened 2022-02-22
FaF Group name not appearing on the Case created in CM

PiperS52:
## Changes in this PR

- when a FaF request is made with a group (MAT or Fed) this is now displayed in case management

### After
Screenshot 2022-02-22 at 12 15 37

Screenshot 2022-02-24 at 15 13 23

### DIFF
--- a/app/services/submit_framework_request.rb
+++ b/app/services/submit_framework_request.rb
@@ -45,7 +45,7 @@ private
     @user ||= FrameworkRequestPresenter.new(request).user
   end
 
-  # @return [Support::Organisation]
+  # @return [Support::Organisation] or [Support::EstablishmentGroup]
   def map_organisation
     Support::Organisation.find_by(urn: request.school_urn) || Support::EstablishmentGroup.find_by(uid: request.group_uid)
   end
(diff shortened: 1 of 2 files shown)


app/services/submit_framework_request.rb
```diff
@@ -47,7 +47,7 @@ def user
 
   # @return [Support::Organisation]
```
kennyevil, line 48: Might need to update the documentation here.

kennyevil (approved):

app/services/submit_framework_request.rb
```diff
@@ -45,9 +45,9 @@ def user
     @user ||= FrameworkRequestPresenter.new(request).user
   end
 
-  # @return [Support::Organisation]
+  # @return [Support::Organisation] or [Support::EstablishmentGroup]
```
peterdavidhamilton, line 48: `[Support::Organisation, Support::EstablishmentGroup]`

peterdavidhamilton: FYI this is another point where the proposed service function needs to be used to span the divide.

peterdavidhamilton (approved):
Follow up - it needs to be rolled in to the FaF refactor and could do with a school test for completeness

PiperS52:
[FaF Group name not appearing on the Case created in CM](https://trello.com/c/HGRJcSGw/403-faf-group-name-not-appearing-on-the-case-created-in-cm)

PR #887, opened 2022-02-17
Case worker can reopen a resolved case

PiperS52:
## Changes in this PR

- added aasm gem state machine with event transitions, refactoring assignments_controller and resolutions_controller, as well as added relevant openings_controller and closings_controlller
- 'reopen case' and 'close case'  functionality in case management
- incoming email received on a 'resolved' case reopens it
- incoming email received on a 'closed' case creates a new case
- 'pending' state renamed to 'on_hold'
- ```Support::Case``` state changes recorded to ```ActivityLogItem```

PR #895, opened 2022-02-22
Able to change Contact and Org on 'closed' case - should be disabled

PiperS52:
## Changes in this PR

- branched from https://github.com/DFE-Digital/buy-for-your-school/pull/887
- remove change links from school details tab in CM when case closed

### After

Screenshot 2022-02-22 at 16 12 04


app/services/support/incoming_emails/case_assignment.rb
```diff
@@ -17,7 +17,10 @@ def case_for_email
         case_reference ||= case_reference_from_conversation
 
         found_case = Support::Case.find_by(ref: case_reference) if case_reference.present?
-        found_case || (new_case_for_email unless email.sent_items?)
+
+        return found_case if found_case && !found_case.closed?
+
+        new_case_for_email unless email.sent_items?
```
ryantk, line 23: In the scenario "Create a new case when existing one is closed" this might not work end to end.

If you look at this line: https://github.com/DFE-Digital/buy-for-your-school/blob/16f38c00286be105e757319b94cbc78b5f0e2ab7/app/models/support/email.rb#L51 

The assignment of the case will be bypassed because it already has a case_id (although closed).

ryantk (changes requested):

app/services/support/incoming_emails/case_assignment.rb
PiperS52: Good spot. @ryantk is it possible to turn line 51 into `return if case_id.present? && !self.case.closed?`

PiperS52:
[Able to change contact and Org on a 'Closed' case. Should be disabled.](https://trello.com/c/HSL2NKac/380-able-to-change-contact-and-org-on-a-closed-case-should-be-disabled)

EdwinKruglov (approved):

d-a-v-e (approved):

ryantk (approved):

PR #903, opened 2022-02-28
Fix bug: graceful validation of duplicate attachments

ryantk:
## Changes in this PR

Previously the first attachment saves and then throws a 500 error before saving the other.
Now it will not save any of them and warn the user instead.

Screenshot 2022-02-28 at 12 00 28


peterdavidhamilton (approved):
Flagging that if the spec prerequisites could be dry-er and more composable it's always good.

ryantk:
> @ryantk nearly forgot to add a changelog reminder

That was added with the main ticket.

PR #908, opened 2022-03-01
FaF support journey start

peterdavidhamilton:
- allow a FaF support request journey to be started
- allow subsequent non-FaF support requests from the dashboard via the profile page
- only clear the faf journey session variable once the first authenticated answer has been posted


peterdavidhamilton:
[Specify support requests go to (FaF) procurement support instead of support-requests from the profile page](https://trello.com/c/3ynZjoVQ/396-specify-support-requests-go-to-faf-procurement-support-instead-of-support-requests-from-the-profile-page)

PR #945, opened 2022-03-30
499 The timestamp in CM is an hour behind

AndrewCowle:
## Changes in this PR

The timezone was not set in application config so that timestamps had been showing incorrectly. 

- Added 'London' as the default timezone.

Timestamps are correct with daylight savings now.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -49,5 +49,8 @@ module BuyForYourSchool
     config.exceptions_app = routes
 
     config.feature_flags = config_for(:feature_flags)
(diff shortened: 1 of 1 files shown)


kennyevil (approved):

d-a-v-e (approved):

ryantk:
[The timestamp in CM is an hour behind](https://trello.com/c/StplAzju/499-the-timestamp-in-cm-is-an-hour-behind)

PR #975, opened 2022-04-21
Case navigation back link

EdwinKruglov:
## Changes in this PR

- add `back_to` parameter to case paths that contains the encoded current URL when going from cases `index` action to `show` action so it can be used as the back link

### DIFF
--- a/app/controllers/support/application_controller.rb
+++ b/app/controllers/support/application_controller.rb
@@ -30,8 +30,8 @@ module Support
       true
     end
 
-    def current_url_b64
-      Base64.encode64(request.fullpath)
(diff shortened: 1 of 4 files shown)


EdwinKruglov:
[Users are being returned to the first page rather than to the page they were on. When looking at a list of cases.](https://trello.com/c/WGS4LfmD/533-users-are-being-returned-to-the-first-page-rather-than-to-the-page-they-were-on-when-looking-at-a-list-of-cases)

ryantk (approved):

serenaabbott11 (approved):

PR #978, opened 2022-04-25
Split out messaging from case history

ryantk:
## Changes in this PR

- add the Messages tab to the case view that contains sent and received Outlook emails, sent Notify emails and manually logged emails and phone calls
- simplify the Case History tab design and exclude all emails and phone calls

image
image


app/views/support/cases/messages/_message.html.erb
```diff
@@ -0,0 +1,17 @@
...
+    <% if message.email_to_school? %>
+      <%= I18n.t("support.case.label.messages.table.caseworker") %>
+    <% elsif message.email_from_school? %>
+      <%= I18n.t("support.case.label.messages.table.from_school") %>
+    <% else %>
+      <%= I18n.t("support.interaction.type.#{message.event_type}") %>
+    <% end %>
+  </th>
+  <td class="govuk-table__cell">
+    <%= message.body.html_safe %>
```
ryantk, line 14: As per my other comment, we need to differentiate between "Log contact with schools" interactions where the message has just been copied in, and emails sent by notify which will be in markdown and need to be rendered as such.

ryantk (commented):
Really good work on this one, it is basically done just need to address my comments around messages.

PR #1009, opened 2022-05-26
Update Actions to reflect new journey for sending emails

EdwinKruglov:
## Changes in this PR

- remove the email type page where the user chooses whether to send a templated or non-templated email
- rename the "Send email" action to "Send template email" in CMS and take the user straight to the template page

image


ryantk:
Ahh, I might have doubled this work up on the other branch.

PR #1036, opened 2022-06-14
SubmitFrameworkRequest bugfix

EdwinKruglov:
Rollbar error: https://rollbar.com/dfe-digital/GHBS/items/467

## Changes in this PR

- revert changes made to `SubmitFrameworkRequest` where the `request` was being cast from `FrameworkRequest` into `FrameworkRequestPresenter`, thus providing a user struct for non-DSI users.


ryantk (approved):
Good catch

PR #1055, opened 2022-07-25
PWNN-303 Exit survey

ryantk:
## Changes in this PR

- add the exit survey
- update the exit survey email to include a link to internal exit surveys

### Before merge

- [x] Revert exit survey email template ID in `email_templates.rb` once testing is done

### After deploy

- [ ] Update "Supported buying exit survey to closed cases" template in GOV.UK Notify to match the "exit survey test" template


app/services/support/create_case.rb
```diff
@@ -39,6 +39,11 @@ def call
         action: "create_case",
       ).call
 
+      CreateExitSurveyResponse.new(
```
ryantk, line 42: Please move this so that it is created when an exit survey is sent out. 
This would allow us to have many exit surveys per case.

app/forms/exit_survey/form.rb
```diff
@@ -0,0 +1,45 @@
+#
+# Base form object
+#
+class ExitSurvey::Form
```
ryantk, line 4: Is this necessary? It seems to duplicate BaseForm a lot.

ryantk (commented):
This looks really good so far, just a few comments.

EdwinKruglov (approved):

PR #1185, opened 2022-10-14
Ensure pages within the service can be differentiated in Google Analytics

EdwinKruglov:
## Changes in this PR

- give the pages that make up the procurement help and support journey their own routes and page titles


app/controllers/framework_requests/confirm_organisations_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+        session[:faf_group] = @form.org_id
+        session.delete(:faf_school)
+      else
+        session[:faf_school] = @form.org_id
+        session.delete(:faf_group)
+      end
+    end
+
+    def query_organisation!
+      if form.group
```
ryantk, line 58: Consider moving this logic to the form itself, have a `#form_or_group` method that handles this work, then call it from `app/views/framework_requests/confirm_organisations/_form.html.erb`

PR #1336, opened 2023-04-04
feat(cms): sortable tables

EdwinKruglov:
## Changes in this PR
- Case tables in CMS can now be sorted.

Jira: PWNN-1214


app/controllers/support/towers_controller.rb
```diff
@@ -7,7 +7,7 @@ class TowersController < ::Support::ApplicationController
 
     def show
       @filter_form = Support::CaseFilterForm.new(
-        **filter_cases_params.to_h.merge(base_cases: @tower.cases),
+        **filter_cases_params.to_h.merge(base_cases: @tower.cases).merge(defaults: { state: "live" }),
```
ryantk, line 10: As merge takes a hash you dont need to merge twice
```
**filter_cases_params.to_h.merge(base_cases: @tower.cases, defaults: { state: "live" })
```

app/javascript/components/sortable-th.js
```diff
@@ -0,0 +1,35 @@
...
+    case "asc_desc":
+      return currentSort === "none" || currentSort === "descending" ? "ascending" : "descending";
+    case "desc_asc":
+      return currentSort === "none" || currentSort === "ascending" ? "descending" : "ascending";
+    default:
+      return "ascending";
+  }
+}
+
+const sortableTh = (config) => {
```
ryantk, line 16: Good solution, might have been nice to use a stimulus controller, maybe one to have a go at refactoring another time.

PR #1361, opened 2023-04-13
1004 - feat: add referral routes

EdwinKruglov:
## Changes in this PR
- add referral routes and track them

Jira: PWNN-1004

### DIFF
--- a/app/controllers/referrals/referrals_controller.rb
+++ b/app/controllers/referrals/referrals_controller.rb
@@ -1,0 +1,28 @@
+module Referrals
+  class ReferralsController < ApplicationController
+    skip_before_action :authenticate_user!
+
+    def rfh
+      redirect_to framework_requests_path
+    end
+
+    def specify
+      redirect_to root_path
+    end
+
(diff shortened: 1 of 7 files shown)


ryantk (commented):
Good overall approach with this, I would like to make a change to simplify it a bit.

1, with your routes, flip it around so we can parameterize the step description rather than hard code them - this will allow them to make new ones without dev input:
```
get "/rfh/:referral_path", to: "referrals#rfh"

2, Tweak UserJourneyTracking initializer somewhat
```
  def initialize(app, path_root, product_section, step_description_source: :request_path, path_root_strict: false)
```
So we can set `step_description_source` to be a proc / request_path symbol - the proc should be passed the request object so we can get the params

```
config.middleware.use UserJourneyTracking, "/referrals", :govuk, :partial_path, step_description_source: ->(request) { request.get_params???("referral_path") }
```
(And change the step_description method to occomdate!)

But other than that, great stuff

ryantk (approved):

PR #1414, opened 2023-05-22
Record interactions for additional changes

TimWilde:
## Changes in this PR

Record interations for:
- Organisation changes
- Contact changes
- Support level changes

### DIFF
--- a/app/controllers/support/cases/contact_details_controller.rb
+++ b/app/controllers/support/cases/contact_details_controller.rb
@@ -7,7 +7,7 @@ module Support
     def update
       @case_contact_details_form = CaseContactDetailsForm.from_validation(validation)
 
-      if validation.success? && @case_contact_details_form.update_contact_details(current_case)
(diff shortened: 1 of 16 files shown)


ryantk (changes requested):
This will do the job but I would like to take the chance to move all case history logging into app/cores/case_history/x.rb

Take a look at the relationship between
`app/cores/case_management/assign_case_to_agent.rb`
and
`app/cores/case_history/handle_case_state_changes.rb`
taking note of `agent_assigned_to_case`.

We could either make the CaseXForm publish some events for the case_history core to listen to, or create a new object in the app/cores/case_management and delegate from the CaseXForms.

Please slack me if that doesn't make sense.

TimWilde:
Sounds good - I'll convert one of the three I've added then ask for another review to make sure I'm on the right track, before doing the rest :)

ryantk (approved):

EdwinKruglov (approved):

PR #1539, opened 2023-08-01
1432 - feat(cms): wide panel view

EdwinKruglov:
## Changes in this PR
- enable widescreen view on cases index page
- replace case tables with panels
- update case filters to allow multiple choices

Jira: PWNN-1432

### DIFF
--- a/app/models/checkbox_option.rb
+++ b/app/models/checkbox_option.rb
@@ -1,0 +1,24 @@
+class CheckboxOption
+  attr_reader :id, :title, :exclusive
+
+  def self.from(collection, id_field: :id, title_field: :title, include_all: false, include_unspecified: false)
+    result = collection.map { |key| new(id: key.send(id_field), title: key.send(title_field)) }
+    result.unshift(unspecified) if include_unspecified
+    result.unshift(all) if include_all
+    result
+  end
+
+  def self.all
+    new(id: "all", title: "All", exclusive: true)
+  end
+
+  def self.unspecified
+    new(id: "unspecified", title: "Unspecified", exclusive: true)
+  end
+
+  def initialize(id:, title:, exclusive: false)
+    @id = id
+    @title = title
+    @exclusive = exclusive
+  end
+end
--- a/app/models/support/case/filterable.rb
+++ b/app/models/support/case/filterable.rb
@@ -2,13 +2,21 @@ module Support::Case::Filterable
   extend ActiveSupport::Concern
 
   included do
-    scope :by_agent, ->(agent_id) { where(agent_id:) }
+    scope :by_agent, ->(agent_ids) { by_filter(:agent_id, agent_ids) }
 
-    scope :by_state, ->(state) { state == "live" ? live : where(state:) }
+    scope :by_state, lambda { |states|
+      states = Array(states)
+      if states.include?("all")
+        all
+      else
+        base = where(state: states.excluding("live", "all"))
+        states.include?("live") ? base.merge(live) : base
+      end
+    }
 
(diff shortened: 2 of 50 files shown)


app/models/checkbox_option.rb
```diff
@@ -0,0 +1,24 @@
...
+    result.unshift(none) if include_none
+    result.unshift(all) if include_all
+    result
+  end
+
+  def self.all
+    new(id: "all", title: "All", exclusive: true)
+  end
+
+  def self.none
```
ryantk, line 15: Should probably call this "def self.unspecified"

app/models/support/case/filterable.rb
```diff
@@ -2,24 +2,60 @@ module Support::Case::Filterable
   extend ActiveSupport::Concern
 
   included do
-    scope :by_agent, ->(agent_id) { where(agent_id:) }
-
-    scope :by_state, ->(state) { state == "live" ? live : where(state:) }
+    scope :by_agent, lambda { |agent_id|
+      agent_id = Array(agent_id)
+      return if agent_id.include?("all")
```
ryantk, line 7: Try to avoid returns in procs/lambdas, is there a way to make this a single if/case statement?

app/models/support/case/filterable.rb
```diff
@@ -2,24 +2,60 @@ module Support::Case::Filterable
...
+        base
+      end
+    }
 
     scope :live, -> { where(state: %w[initial opened on_hold]) }
 
-    scope :by_category, ->(category_id) { where(category_id:) }
+    scope :by_category, lambda { |category_id|
+      category_id = Array(category_id)
+      return if category_id.include?("all")
```
ryantk, line 29: Same as above, avoid returns here.

app/models/support/case/filterable.rb
```diff
@@ -42,9 +78,14 @@ def valid_scopes_for_filtering(filtering_criteria)
 
         next unless respond_to?(potential_scope)
```
ryantk, line 80: If we use the `Array(x)` pattern we can assume everything is an array (Check out https://apidock.com/ruby/Kernel/Array/instance)

So we can remove the if statement and modify the "each" part to keep the behaviour you added.

Something like:
```ruby
Array(criteria)
  .select { |criterion| criterion.present? || criterion == false }
  .each   { |criterion| (valid_scopes[potential_scope] ||= []) = criterion }
```

The `select` part should cover the  criteria.reject!(&:blank?)` logic

app/models/support/case/filtering.rb
```diff
@@ -22,17 +25,21 @@ class Support::Case::Filtering
...
+
+  def sorting_criteria
+    { sort:, sort_by:, sort_order: }
   end
 
 private
 
   def closed_cases_hidden_by_default(cases)
-    return cases.not_closed unless state == "closed"
+    return cases.not_closed unless state&.include?("closed") || search_term.present?
```
ryantk, line 42:
```
Array(state).include?("closed")
```
Seems more "confident" code

app/models/support/case/filtering.rb
```diff
@@ -11,7 +11,10 @@ class Support::Case::Filtering
     :level,
     :has_org,
     :search_term,
+    :procurement_stage,
     :sort,
```
ryantk, line 15: Is there anything which still uses "sort"?

It would be good to start using only the sort_X fields and refactor:

sortable.rb
```
 def sorted_by(field, order)
```

To simplifiy it's code.

ryantk (changes requested):
This is a nice change overall, just some tweaks to make.

PR #1577, opened 2023-08-24
1539 - feat(cms): persist procurement schools on case

EdwinKruglov:
## Changes in this PR
- persist schools selected in request on the case via the `support_case_organisations`table

Jira: PWNN-1539

### DIFF
--- a/app/models/support/case.rb
+++ b/app/models/support/case.rb
@@ -50,6 +50,8 @@ module Support
 
     belongs_to :procurement_stage, class_name: "Support::ProcurementStage", optional: true
 
+    has_one :framework_request, class_name: "FrameworkRequest", foreign_key: :support_case_id
+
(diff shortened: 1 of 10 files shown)


app/models/support/case.rb
```diff
@@ -42,10 +42,16 @@ class Case < ApplicationRecord
 
     has_many :energy_bills, class_name: "EnergyBill", foreign_key: :support_case_id
 
+    has_many :case_organisations, class_name: "Support::CaseOrganisation", foreign_key: :support_case_id
+
+    has_many :organisations, through: :case_organisations
```
ryantk, line 47: Even though this is technically a link to an organisation, I think this will get confused with the other relationship "organisation". We need something more explicit

Something like:
```
has_many :schools_invovled_in_procurement, through: :case_organisations, source: :organisation
```

ryantk (changes requested):

ryantk (approved):

PR #1620, opened 2023-09-22
[PWNN-1653] Request case source from E&O during create a case

cocoaken:


app/assets/config/manifest.js
```diff
@@ -2,4 +2,5 @@
 //= link_tree ../../javascript/faf
 //= link_tree ../../javascript/misc
 //= link_tree ../../javascript/request
+//= link_tree ../../javascript/discovery_method
```
ryantk, line 5: I think this line is causing the build to fail

db/schema.rb
```diff
@@ -725,7 +726,7 @@
     t.string "ukprn"
     t.string "telephone_number"
     t.jsonb "local_authority"
-    t.datetime "opened_date", precision: nil
+    t.datetime "opened_date"
```
ryantk, line 729: which version of postgres do you have installed, there is a lot of changes re "precision: nil"

cocoaken: postgres 15.4 / PostGIS 3.3.4

app/assets/config/manifest.js
cocoaken: Well spotted 👍

Gemfile.lock
```diff
@@ -124,6 +124,7 @@ GEM
     bootsnap (1.16.0)
       msgpack (~> 1.2)
     brakeman (6.0.1)
+    brow (0.4.1)
```
ryantk, line 127: These changes might have something to do with the random test failures.
Undo this change and make sure you are up to date with main

app/views/support/cases/discovery_method/_discovery_method.html.erb
```diff
@@ -0,0 +1,17 @@
+<%= form.govuk_radio_buttons_fieldset :discovery_method, class: "discovery-method-radios", legend: { size: 'm', text: I18n.t("support.case.label.discovery_method.legend") }, "data-controller": "discovery-method" do %>
```
EdwinKruglov, line 1: Is the reference to that `data-controller` still necessary here?

cocoaken: No you're right, it isn't!

EdwinKruglov (approved):

PR #1638, opened 2023-10-12
1657 - fix(cms): duplicate email attachments not displaying correctly

cocoaken:
## Changes in this PR

Fixes 3 bugs:
- Duplicate e-mail attachments not showing correctly (should be 1 of each duplicate within a case) in CMS -> Attachments -> All Files
- Duplicate e-mail attachments missing in CMS -> Attachments -> Received Files / Sent Files even when they're correctly listed in that case's All Files
- Duplicate e-mail attachments going missing between cases

### DIFF
--- a/app/models/support/email_attachment/de_dupable.rb
+++ b/app/models/support/email_attachment/de_dupable.rb
@@ -2,37 +2,22 @@ module Support::EmailAttachment::DeDupable
   extend ActiveSupport::Concern
 
   included do
-    scope :unique_files, lambda { |first_instance_only: true|
-      joins(
-        <<-SQL,
-        INNER JOIN (
-          SELECT
-            support_email_attachments.id,
-            active_storage_blobs.checksum,
-            ROW_NUMBER() OVER(
-              PARTITION BY active_storage_blobs.checksum, active_storage_blobs.filename
-              ORDER BY support_email_attachments.updated_at DESC
-            ) AS rank
-          FROM support_email_attachments
-          JOIN active_storage_attachments
-            ON active_storage_attachments.record_type = 'Support::EmailAttachment'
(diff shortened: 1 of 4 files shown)


app/models/support/email_attachment/de_dupable.rb
```diff
@@ -2,37 +2,30 @@ module Support::EmailAttachment::DeDupable
...
+      unique_attachments = []
+      order("created_at DESC")
+      .each { |attachment| unique_attachments << most_recent_instance(attachment) }
+      where(id: unique_attachments.uniq)
+    }
+
+    scope :most_recent_instance, lambda { |email_attachment|
+      all_instances(email_attachment)
+      .order("created_at DESC")
+      .first
```
ryantk, line 15: I would not return only 1 result from the scope as it violates the rules of scopes (should be chainable)

cocoaken: Doh! You're right.

app/models/support/email_attachment/de_dupable.rb
```diff
@@ -2,37 +2,30 @@ module Support::EmailAttachment::DeDupable
...
-            ON active_storage_blobs.id = active_storage_attachments.blob_id
-        ) attachment_files
-        ON attachment_files.id = support_email_attachments.id
-        #{'AND attachment_files.rank = 1' if first_instance_only}
-      SQL
-      )
+    scope :unique_files, lambda {
+      unique_attachments = []
+      order("created_at DESC")
+      .each { |attachment| unique_attachments << most_recent_instance(attachment) }
```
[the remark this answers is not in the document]
cocoaken: Good point, well made.

PR #1678, opened 2023-12-13
ci: change docker base image to alpine

ryantk:

Change docker base image to alpine to speed up builds


ryantk:
Although the docker image is much neater, it does not actually improve speed all that much

PR #1680, opened 2023-12-18
1678 - feat(e&o): add questions to create a case

serenaabbott11:
## Changes in this PR
- Added same supplier and contract start date known to the E&O create a case form
- Energy and services show contract start date
- MAT picker more than one school picked shows same supplier
- New fields in CYA page and support case request details tab

### DIFF
--- a/app/controllers/engagement/case_requests/same_suppliers_controller.rb
+++ b/app/controllers/engagement/case_requests/same_suppliers_controller.rb
@@ -1,0 +1,41 @@
+module Engagement
+  module CaseRequests
+    class SameSuppliersController < ApplicationController
+      before_action :case_request, only: %i[edit update]
+      before_action :back_url
+
+      def edit
+        @same_supplierer = @case_request.same_supplierer
+      end
+
+      def update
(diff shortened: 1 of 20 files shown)


ryantk (approved):

app/controllers/engagement/case_requests/same_suppliers_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  module CaseRequests
+    class SameSuppliersController < ApplicationController
+      before_action :case_request, only: %i[edit update]
+
+      def edit
+        @back_url = params[:source] == "change_link" ? engagement_case_request_path(@case_request) : edit_engagement_case_request_school_picker_path(@case_request)
+        @same_supplierer = @case_request.same_supplierer
+      end
+
+      def update
```
EdwinKruglov, line 11: Small thing, but it might be a good idea to set the `back_url` for the `update` action as well, using `before_action :back_url` as you've done in the other controller. Otherwise if validation fails here there won't be a back link.

EdwinKruglov (approved):

PR #1681, opened 2023-12-28
ci: add test coverage reporting

EdwinKruglov:
## Changes in this PR
- extend the `rspec` job to upload coverage reports
- add `collate_test_coverage` job to collate test coverage reports from multiple nodes into an HTML artifact


ryantk:
@EdwinKruglov one thing to think about with this, is make sure the coverage report is not wiped out when moving from rspec test run to the retry failed tests run.

cocoaken (approved):

PR #1688, opened 2024-01-10
build: update sidekiq to v7

ryantk:
## Changes in this PR

updated sidekiq gem to major version 7.

NOTE: Read my comment below


ryantk:
This gem requires redis 6.2 at least.
Microsoft Azure does not yet provide any redis version over 6.0.2 so this work cannot be carried out.

threedaymonk:
As we still can't do anything to put this into practice (Azure still doesn't offer Redis 7) and the changes are minimal, I propose that we close this PR for the time being.

PR #1719, opened 2024-02-05
1841 - feat: end-of-journey survey

EdwinKruglov:
## Changes in this PR
- add the `end_of_journey_survey_responses` table to store user feedback collected at the end of the RfH and FaF journeys
- allow JSON POST requests to come in from FaF and store user feedback

### DIFF
--- a/app/views/framework_request_submissions/show.html.erb
+++ b/app/views/framework_request_submissions/show.html.erb
@@ -18,6 +18,14 @@
       <% else %>
         <%= render "default" %>
       <% end %>
+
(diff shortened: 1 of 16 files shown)


ryantk (approved):

app/views/framework_request_submissions/show.html.erb
```diff
@@ -18,6 +18,12 @@
       <% else %>
         <%= render "default" %>
       <% end %>
+
+      <div class="divider-bottom"></div>
```
ryantk, line 22: I think the whole PR looks good to merge, perhaps with a feature flag here until the ticket is reviewed

PR #1722, opened 2024-02-06
1894 - fix(cms): add ability to remove framework

serenaabbott11:
## Changes in this PR
- Ability to clear framework from case
- If there is no framework in the case then there is no clear framework link, if there is a framework then there is a clear framework link that removes the value, clears the box, and removes the clear framework link

Screenshot 2024-02-14 at 09 54 29

### DIFF
--- a/app/javascript/controllers/framework_edit_controller.js
+++ b/app/javascript/controllers/framework_edit_controller.js
@@ -1,0 +1,36 @@
+import { Controller } from "@hotwired/stimulus"
+
+// Connects to data-controller="framework-edit"
+export default class extends Controller {
+    static targets = [ "input", "output", "hidden" ]
+  
+    connect() {
+      if (this.hiddenTarget.value == "") {
+        this.hideClearFrameworkLink();
+      }
+      this.inputTarget.addEventListener("change", this.toggle.bind(this));
+    }
+
+    hideClearFrameworkLink() {
+      this.outputTarget.classList.add("govuk-!-display-none");
+      this.outputTarget.setAttribute("disabled", true);
+    }  
+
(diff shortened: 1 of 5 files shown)


app/javascript/controllers/framework_edit_controller.js
```diff
@@ -0,0 +1,33 @@
+import { Controller } from "@hotwired/stimulus"
+
+// Connects to data-controller="framework-edit"
+export default class extends Controller {
+    static targets = [ "input", "output" ]
+  
+    connect() {
+      this.inputTarget.querySelector("[name='case_procurement_details_form[framework_name]']").addEventListener("change", this.toggle.bind(this));
```
ryantk, line 8: Instead of using querySelector, make a new target and refer to it that way

app/javascript/controllers/framework_edit_controller.js
```diff
@@ -0,0 +1,33 @@
...
+  
+    toggle() {
+      if (this.inputTarget.querySelector("[name='case_procurement_details_form[frameworks_framework_id]']").value != null) {
+        this.showClearFrameworkLink();
+      }
+    }
+
+    clear(e) {
+      e.preventDefault()
+      document.getElementsByName("case_procurement_details_form[framework_name]")[0].value = "";
```
ryantk, line 29: Is there a way we can avoid using the form name "case_procurement_details_form", perhaps a target?

ryantk (changes requested):
Looks good overall, just put a comment about avoiding "hard coding" the field names in the js controller which could make it brittle.

cocoaken (approved):

PR #1774, opened 2024-03-07
1487 - fix(specs): flaky tests

EdwinKruglov:
This is an attempt at fixing as many flaky tests as possible. This does not guarantee that every flaky test has been fixed, but many that were order-dependent should have been fixed.

### Main findings
- Some of the flakiness was being caused by the `display-cc-bcc.js` JS component. I've moved its functionality to the `draft_email_controller.js` Stimulus controller which seems to have helped.
- A number of issues were being caused by sequence fields on FactoryBot factories which would not reset between tests. This has been added to `database_cleaner.rb`.
- Timing issues were being caused by Capybara failing to select an option from the autocomplete dorpdown component. There were multiple `sleep` calls in different tests. I've extracted this into the single method `select_autocomplete_option` in `autocomplete_helpers.rb` and increased its timeout.
- Some `sleep` calls stil remain and may have to be refactored in the future.
- The code block
    ```
    around do |example|
      Current.set(actor: agent) do
        example.run
      end
    end
    ```
    was being used in mulitple tests. For some reason the agents created by it were not being removed between tests. I've replaced these with a simple `Current.actor = agent`.
- A few usages of `allow_any_instance_of` had effects on subsequent tests. This kind of stubbing is generally discouraged as it tends to be buggy. I've removed the ones that were causing obvious issues.


app/javascript/controllers/index.js
```diff
@@ -7,6 +7,9 @@ import { application } from "./application"
 import AutoFilterController from "./auto_filter_controller"
 application.register("auto-filter", AutoFilterController)
 
+import CaseAttachmentController from "./case_attachment_controller"
+application.register("case-attachment", CaseAttachmentController)
+
```
cocoaken, line 12: What was the benefit of moving it up here, out of interest? Genuine question / not trying to sound like a smart***e!

cocoaken (approved):

app/javascript/controllers/index.js
EdwinKruglov: No benefit, just regenerated this file when I was trying things out and it re-ordered things alphabetically.

PR #1941, opened 2024-10-22
2025 - feat(cms): add proposed open and closed status and date to establishm…

serenaabbott11:
## Changes in this PR
- Adds proposed to open and proposed to close status to the organisation drop down alongside the open/closed date if there is one
- Excludes archived records from being shown
- Orders by organisation status

Screenshot 2024-10-28 at 12 01 09

### DIFF
--- a/app/models/support/establishment_search/presentable.rb
+++ b/app/models/support/establishment_search/presentable.rb
@@ -10,7 +10,17 @@ private
   def autocomplete_template_vars
     case source
     when "LocalAuthority" then { name:, establishment_type:, code: }
-    else { name:, postcode:, establishment_type:, urn:, ukprn: }
+    else { name:, postcode:, establishment_type:, proposed_open_or_closed:, urn:, ukprn: }
+    end
+  end
+
+  def proposed_open_or_closed
+    case organisation_status
+    when "opening"
+      I18n.t(organisation_status, scope: "support.organisation_statuses", date: I18n.l(opened_date, format: :compact))
+    when "closing"
+      I18n.t(organisation_status, scope: "support.organisation_statuses", date: I18n.l(closed_date, format: :compact))
+    else ""
     end
   end
 end
--- a/app/views/support/establishment_search/support/organisation/_autocomplete_template.html.erb
+++ b/app/views/support/establishment_search/support/organisation/_autocomplete_template.html.erb
@@ -1,1 +1,6 @@
-<strong><%= name %></strong>, <%= postcode %><br /><%= establishment_type %><br />URN: <%= urn %> - UKPRN: <%= ukprn %>
+<%= safe_join([
+  "<strong>#{h(name)}</strong>, #{h(postcode)}".html_safe,
+  establishment_type,
+  proposed_open_or_closed,
+  "URN: #{urn} - UKPRN: #{ukprn}"
+].compact_blank, "<br />".html_safe) %>
--- a/spec/models/support/establishment_search/presentable_spec.rb
+++ b/spec/models/support/establishment_search/presentable_spec.rb
@@ -3,27 +3,43 @@ require "rails_helper"
 describe Support::EstablishmentSearch::Presentable do
   subject(:presentable) { Support::EstablishmentSearch }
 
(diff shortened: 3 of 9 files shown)


app/models/support/establishment_search/presentable.rb
```diff
@@ -10,7 +10,15 @@ def autocomplete_template
...
-    else { name:, postcode:, establishment_type:, urn:, ukprn: }
+    else { name:, postcode:, establishment_type:, proposed_open_or_closed:, urn:, ukprn: }
+    end
+  end
+
+  def proposed_open_or_closed
+    if organisation_status == "opening"
+      "#{I18n.t("support.organisation_statuses.#{organisation_status}")} #{opened_date.strftime('%d-%m-%Y')}"
+    elsif organisation_status == "closing"
+      "#{I18n.t("support.organisation_statuses.#{organisation_status}")} #{closed_date.strftime('%d-%m-%Y')}"
```
threedaymonk, line 21: A `case` statement might be more readable here: it would eliminate the repetition of `organisation_status`. Should there be an `else` condition, if neither is true?

Rather than concatenating strings, it's preferable to [use a placeholder variable in the locale string](https://guides.rubyonrails.org/i18n.html#passing-variables-to-translations). You can also use a scope to eliminate the interpolation in the key, and if we're using this date pattern in multiple places it might be nice to [name it](https://guides.rubyonrails.org/i18n.html#adding-date-time-formats):

```ruby
I18n.t(organisation_status, scope: "support.organisation_statuses", 
  date: I18n.l(opened_date, format: :short))
```

app/views/support/establishment_search/support/organisation/_autocomplete_template.html.erb
```diff
@@ -1 +1 @@
-<strong><%= name %></strong>, <%= postcode %><br /><%= establishment_type %><br />URN: <%= urn %> - UKPRN: <%= ukprn %>
+<strong><%= name %></strong>, <%= postcode %><br /><%= establishment_type %><% if proposed_open_or_closed.present? %><br /><%= proposed_open_or_closed %><%end%><br />URN: <%= urn %> - UKPRN: <%= ukprn %>
```
threedaymonk, line 1: A pattern like `[a, b, c].compact_blank.join("<br />")` might obviate the need for the presence check

spec/models/support/establishment_search/presentable_spec.rb
```diff
@@ -3,14 +3,30 @@
...
+    context "when the search result is an opening school" do
       it "returns the correct autocomplete template" do
-        search_result = presentable.omnisearch(school.urn).to_a.first
+        search_result = presentable.omnisearch(opening_school.urn).to_a.first
+        expect(search_result.autocomplete_template).to eq("<strong>Test Opening School</strong>, EC3A 5DE<br />Foundation school<br />Proposed to open 30-07-2024<br />URN: 3 - UKPRN: 333\n")
+      end
+    end
+
+    context "when the search result is a closing school" do
+      it "returns the correct autocomplete template" do
```
threedaymonk, line 21: I realise that this vocabulary has been inherited from the existing code, but can we replace "correct" with an explanation of what correctness means here (and in the examples above and below)?

threedaymonk (changes requested):
I think there are a couple of opportunities to make the code a bit more readable, which I've noted.

The database views seem rather complex. Is there anything in the test suite that exercises them?

app/models/support/establishment_search/presentable.rb
```diff
@@ -10,7 +10,17 @@ def autocomplete_template
...
     when "LocalAuthority" then { name:, establishment_type:, code: }
-    else { name:, postcode:, establishment_type:, urn:, ukprn: }
+    else { name:, postcode:, establishment_type:, proposed_open_or_closed:, urn:, ukprn: }
+    end
+  end
+
+  def proposed_open_or_closed
+    case organisation_status
+    when "opening"
+      I18n.t(organisation_status, scope: "support.organisation_statuses", date: I18n.l(opened_date, format: "%d %b %y"))
```
threedaymonk, line 20: Can you give the format a name, as here: https://guides.rubyonrails.org/i18n.html#adding-date-time-formats

threedaymonk:
Looks great!

threedaymonk (approved):

PR #1971, opened 2024-11-05
PWNN-2066 Don't create duplicate support agents with the same email address

threedaymonk:
This addresses PWNN-2066, "Proc-ops user couldn't sign when two agent accounts are created with the same email id":

> Proc-ops users experience trouble signing into the CMS when there are two agent accounts created for the same email address one with roles assigned and another with roles unassigned.
> 
> Temporary fix:  the two accounts have been assigned with the same privileges to work around.
> 
> Expected Outcome: When the proc-ops manager tries to create a new agent account with the email address that already belongs to another agent (  current staff or former staff) , then instead of creating a new agent account, it should owerwrite the former agent’s account with the selected roles for the new agent .
> 
> If the first name/ last name or both are different for the new agent and only the email address is common between new agent and former agent, then when trying to add the new agent, it should revive the former agent account and overwrite the former agent’s first name and last name with the new agent’s first name and last name.

## Changes in this PR

- The ManageSupportAgents concern that shared code between two similar controllers is removed.
- AgentForm is now a subclass of Support::Agent, so that the controllers are simpler. All it deals with is form-related validation and policy-guarded role assignment.
- When adding a support agent, if an existing record has the same email, that record will be updated with the name and roles entered.
- Duplicate support agents are removed from the database. The newer records are removed.
- Uniqueness validation and constraints are added to the email field.


threedaymonk:
The Sonar complaint about duplicated code is because there are, indeed, two duplicated controllers. Someone "worked around" this in the past by moving the duplicated code into a Concern, which is really a misuse of those. I've eliminated most of the complexity from the controllers, so it feels better to simply have the same code in each. (A question for another time is why on earth do we have two controllers that do the same thing? but let's not dive into that right now.)

serenaabbott11 (approved):

PR #1977, opened 2024-11-09
PWNN-2097 Ability to select DfE approved or CAB approved framework on a case

edwin-jebaraj:
## Changes in this PR

1. app/controllers/support/frameworks_controller.rb - added condition to show only DfE approved or CAB approved framework

### DIFF
--- a/app/controllers/support/frameworks_controller.rb
+++ b/app/controllers/support/frameworks_controller.rb
@@ -12,6 +12,7 @@ module Support
 
           render json: ::Frameworks::Framework
             .not_evaluating
+            .where(status: %w[dfe_approved cab_approved])
             .omnisearch(params[:q])
             .order("reference DESC")
             .as_json(methods:)


app/controllers/support/frameworks_controller.rb
```diff
@@ -12,6 +12,7 @@ def index
 
           render json: ::Frameworks::Framework
             .not_evaluating
+            .where(status: %w[dfe_approved cab_approved])
```
threedaymonk, line 15: I feel that this would be better as a scope on `Framework` so that we could have `.approved` here (assuming that there are no other `???_approved` statuses).

I know this is a trivial change, but I still want to see a test for this. It only needs to be a short one in `frameworks_controller_spec.rb` that creates a few frameworks and asserts that ones that should be there are, and the ones that shouldn't aren't.

threedaymonk (changes requested):

PR #1990, opened 2024-11-18
2012 - feat(cms): Remove archived establishments from selection options

serenaabbott11:
Archived organisations and establishment groups are those that are in our databases but are no longer in the GIAS data download for various reasons, such as a school never opening. We do not want Procurement Operations or school users to be able to select archived schools from any of the dropdowns or school pickers as there should be an alternate correct establishment to choose from.

## Changes in this PR
- Removes archived schools from the selection in the multi school picker
- Removes archived school from the organisation and establishment group searches

### DIFF
--- a/app/controllers/support/establishment_groups_controller.rb
+++ b/app/controllers/support/establishment_groups_controller.rb
@@ -6,9 +6,10 @@ module Support
 
     def index
       query = <<-SQL
-        ukprn LIKE :q OR
+        (ukprn LIKE :q OR
         uid LIKE :q OR
-        lower(name) LIKE lower(:q)
+        lower(name) LIKE lower(:q)) AND
+        archived != true
       SQL
 
       respond_to do |format|
--- a/app/controllers/support/establishment_groups_controller.rb
+++ b/app/controllers/support/establishment_groups_controller.rb
@@ -6,15 +6,15 @@ module Support
 
     def index
       query = <<-SQL
-        (ukprn LIKE :q OR
+        ukprn LIKE :q OR
         uid LIKE :q OR
-        lower(name) LIKE lower(:q)) AND
-        archived != true
+        lower(name) LIKE lower(:q)
       SQL
 
       respond_to do |format|
         format.json do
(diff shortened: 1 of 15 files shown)


app/models/local_authority.rb
```diff
@@ -9,7 +9,7 @@ def eligible_for_school_picker?
   end
 
   def organisations_for_multi_school_picker
-    organisations.local_authority_maintained
+    organisations.local_authority_maintained.where.not(archived: true)
```
threedaymonk, line 12: I wonder whether this is pointing to a smell in the database: we have a mixture of `archived` fields, some of which allow null, some of which don't. It might be nice to write a migration that sets all nulls to false and then adds default false and a non-null constraint on every table that doesn't currently do that.

Then we can say `where(archived: false)` instead, and in fact for consistency we should add an `active` scope (see `category.rb` and `email_template.rb`) which could also then be a concern as it's the same everywhere there's an archive attribute.

threedaymonk (changes requested):

app/controllers/support/establishment_groups_controller.rb
```diff
@@ -6,9 +6,10 @@ class EstablishmentGroupsController < ApplicationController
 
     def index
       query = <<-SQL
-        ukprn LIKE :q OR
+        (ukprn LIKE :q OR
         uid LIKE :q OR
-        lower(name) LIKE lower(:q)
+        lower(name) LIKE lower(:q)) AND
+        archived != true
```
threedaymonk, line 12: It might be better (rather, more uniform and consistent) to leave `archived` out here and add an extra constraint below:

```ruby
          render json: EstablishmentGroup
            .active
            .where(query, q: "%#{params.fetch(:q)}%")
```

(Also, if you have any ideas for a better name for `query` that would better explain what this does, feel free to rename it. Nothing springs to my mind, though.)

threedaymonk (approved):
I have one further suggestion, but I'm happy with this as is.

PR #1996, opened 2024-11-20
SC-81 Add evaluators

threedaymonk:
Adds a `Support::Evaluator` model belonging to a `Support::Case` and a controller to manipulate evaluators on the case.

Implements the new confirmation page style, which works by using a `confirm` parameter on the destroy action to determine whether to show the confirmation screen or to go ahead and delete.

screenshot-2024-11-25-16-24-10 214

screenshot-2024-11-25-16-24-11 200

screenshot-2024-11-25-16-24-11 731

screenshot-2024-11-25-16-24-12 393

screenshot-2024-11-25-16-24-12 949

screenshot-2024-11-25-16-24-13 257

screenshot-2024-11-25-16-24-13 605


threedaymonk:
This really needs the contextual banner across the top to show where we are, but I don't know if there's a component for that.

app/views/support/cases/evaluators/edit.html.erb
```diff
@@ -0,0 +1,4 @@
+<span class="govuk-caption-l"><%= @current_case.organisation_name %></span>
```
serenaabbott11, line 1: The only thing I can see to possibly change is that on the design there is the case ref just before the organisation name in the captions

threedaymonk: Implemented!

serenaabbott11 (approved):

PR #1998, opened 2024-11-25
PWNN-2097 Ability to select DfE approved or CAB approved framework

edwin-jebaraj:
## Changes in this PR

**PWNN-2097 Ability to select DfE approved or CAB approved framework**

app/controllers/support/frameworks_controller.rb  - Added new condtion to filter approved framework

### DIFF
--- a/app/controllers/support/frameworks_controller.rb
+++ b/app/controllers/support/frameworks_controller.rb
@@ -12,6 +12,7 @@ module Support
 
           render json: ::Frameworks::Framework
             .not_evaluating
+            .where(status: %w[dfe_approved cab_approved])
             .omnisearch(params[:q])
             .order("reference DESC")
             .as_json(methods:)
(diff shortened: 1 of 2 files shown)


app/controllers/support/frameworks_controller.rb
```diff
@@ -12,6 +12,7 @@ def index
 
           render json: ::Frameworks::Framework
             .not_evaluating
+            .where(status: %w[dfe_approved cab_approved])
```
threedaymonk, line 15: rather than putting this in the controller, can we have an "approved" named scope instead?

threedaymonk (changes requested):

app/controllers/support/frameworks_controller.rb
edwin-jebaraj: @threedaymonk  updated

threedaymonk (approved):
All good, please rebase and then it can be merged.

PR #2009, opened 2024-12-03
SC-124 Add evaluators link and status change funtionality to tasklist

serenaabbott11:
## Changes in this PR
- Adds the correct link to the add evaluators page
- Changes the status to complete when there is an evaluator added
- Adds back links to the pages
- Adds validation for email format, and custom validation messages

Screenshot 2024-12-05 at 12 35 00

### DIFF
--- a/app/models/support/evaluator.rb
+++ b/app/models/support/evaluator.rb
@@ -2,10 +2,9 @@ module Support
   class Evaluator < ApplicationRecord
     belongs_to :support_case, class_name: "Support::Case"
 
-    include HasNormalizedEmail
-
     validates :email,
               presence: true,
+              format: { with: URI::MailTo::EMAIL_REGEXP },
               uniqueness: { case_sensitive: false, scope: :support_case_id }
     validates :first_name, :last_name, presence: true
 
--- a/app/views/support/cases/show/_tasklist.html.erb
+++ b/app/views/support/cases/show/_tasklist.html.erb
@@ -8,7 +8,7 @@
         <%= I18n.t("support.case.label.tasklist.complete_evaluation") %>
       </h3>
       <%= govuk_task_list(id_prefix: "complete-evaluation") do |task_list|
-        task_list.with_item(title: I18n.t("support.case.label.tasklist.item.add_evaluators"), href: '#', status: govuk_tag(text: I18n.t("support.case.label.tasklist.status.to_do")))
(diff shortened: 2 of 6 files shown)


app/views/support/cases/show/_tasklist.html.erb
```diff
@@ -8,7 +8,7 @@
         <%= I18n.t("support.case.label.tasklist.complete_evaluation") %>
       </h3>
       <%= govuk_task_list(id_prefix: "complete-evaluation") do |task_list|
-        task_list.with_item(title: I18n.t("support.case.label.tasklist.item.add_evaluators"), href: '#', status: govuk_tag(text: I18n.t("support.case.label.tasklist.status.to_do")))
+        task_list.with_item(title: I18n.t("support.case.label.tasklist.item.add_evaluators"), href: support_case_evaluators_path(@current_case), status: @current_case.evaluators.present? ? govuk_tag(text: I18n.t("support.case.label.tasklist.status.complete"), colour: "green") : govuk_tag(text: I18n.t("support.case.label.tasklist.status.to_do")))
```
threedaymonk, line 11: In the long run I think we probably want to find a briefer and more uniform abstraction for enumerating task list statuses, and to push some of this logic into the model layer. This is good for now, though.

threedaymonk (approved):

app/models/support/evaluator.rb
```diff
@@ -2,10 +2,9 @@ module Support
   class Evaluator < ApplicationRecord
     belongs_to :support_case, class_name: "Support::Case"
 
-    include HasNormalizedEmail
-
     validates :email,
               presence: true,
+              format: { with: URI::MailTo::EMAIL_REGEXP },
```
threedaymonk, line 7: This is a reasonable approach, but if we end up with email validation in multiple places it would be worth writing a custom validator so that we can have a universal error message and write something like

```ruby
validates :email, email_address: true
```

E.g.: 

```ruby
class EmailAddressValidator < ActiveModel::EachValidator
  def validate_each(record, attribute, value)
    return if valid_email?(value) # TODO

    record.errors.add attribute, :invalid_email_address, **options
  end
end
```

and

```yaml
en:
  errors:
    messages:
      invalid_email_address: "Enter an email address in the correct format, like name@example.com"
```

For validation you can match a regexp or use something more robust and comprehensive like [this gem](https://github.com/afair/email_address).

threedaymonk (approved):

app/models/support/evaluator.rb
serenaabbott11: Thanks Paul, as we chatted about, I'll be doing this work in another ticket

PR #2012, opened 2024-12-04
Add feature flag documentation

serenaabbott11:
## Changes in this PR
- Adds documentation about feature flags - what they are, current flags, and where to find the username and passwords


threedaymonk:
This is great, but I think a few more details would be helpful:

- how do I add/use a feature flag in code?
- how do I remove one when it's no longer needed?
- a link to Flipper so we know where to look for further documentation.

threedaymonk (approved):

serenaabbott11:
Thanks for the comment @threedaymonk, I've added some extra info if you wouldn't mind having a look?

threedaymonk (approved):

PR #2015, opened 2024-12-06
SC-84 Evaluator task list stub

threedaymonk:
This is a very basic stub page for the evaluator task list so that we have a place to take people when they sign in.

The due date is not filled in as this doesn't yet exist, and there are no task list items.

When you visit the page and you're signed in and an evaluator:

screenshot-2024-12-06-16-44-30 166

When you are signed in but not an evaluator, you are dumped back to the home page:

screenshot-2024-12-06-16-44-31 654

We might have to do further work to make the authentication journey work properly end to end once we are emailing evaluators, as the tests don't accurately reflect the real-world sign-in experience.

### DIFF
--- a/app/controllers/evaluation/tasks_controller.rb
+++ b/app/controllers/evaluation/tasks_controller.rb
@@ -1,13 +1,24 @@
 module Evaluation
   class TasksController < ApplicationController
+    before_action :set_current_case
     before_action :check_user_is_evaluator
+
+    def edit
+      session[:email_evaluator_link] = evaluation_task_path(@current_case, host: request.host)
+      redirect_to evaluation_task_path(@current_case)
+    end
 
     def show; end
 
   private
 
     helper_method def current_evaluator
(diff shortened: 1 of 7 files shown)


app/controllers/evaluation/tasks_controller.rb
```diff
@@ -1,13 +1,24 @@
...
+      redirect_to evaluation_task_path(@current_case)
+    end
+
     def show; end
 
   private
 
     helper_method def current_evaluator
-      @current_evaluator ||= Support::Evaluator.find(params[:id])
+      @current_evaluator ||= Support::Evaluator.find_by(support_case_id: params[:id], email: current_user.email)
```
threedaymonk, line 16: What happens if it's not found? Should this be `find_by!` to raise an exception?

edwin-jebaraj: updated

edwin-jebaraj (approved):

PR #2017, opened 2024-12-12
SC-83 Invite evaluators to start evaluation tasks

edwin-jebaraj:
## Changes in this PR
1. app/controllers/support/cases/email_evaluators_controller.rb - Added new controller
2. app/models/email/template_parser.rb - Added new parse variable
3. app/views/support/cases/email_evaluators/_document_list.html.erb - Added new view file to list uploaded documents
4. app/views/support/cases/email_evaluators/_email_list.html.erb - Added new view file to list evaluators
5. app/views/support/cases/email_evaluators/_form.html.erb - Added new form file
6. app/views/support/cases/email_evaluators/_form_template.html.erb - Added new default template file
7.  app/views/support/cases/email_evaluators/edit.html.erb - Added new view file
8. app/views/support/cases/show/_tasklist.html.erb - Updated email evaluator link in task list
9. config/locales/en.yml - added new language variable
10. config/routes.rb - Added new route for email_evaluators
11. db/migrate/20241211171744_add_sent_email_to_evaluators_to_case.rb - Added new migration file
12. spec/features/support/agent_can_email_to_evaluators_spec.rb - Added new spec file for testing

### DIFF
--- a/app/controllers/support/cases/email_evaluators_controller.rb
+++ b/app/controllers/support/cases/email_evaluators_controller.rb
@@ -1,0 +1,89 @@
+module Support
+  class Cases::EmailEvaluatorsController < Cases::ApplicationController
+    helper_method :back_to_url_b64
+
+    content_security_policy do |policy|
+      policy.style_src_attr :unsafe_inline
+    end
+
+    before_action :set_current_case
+    before_action :set_email_addresses
+    before_action :set_documents
+    before_action :set_template
+    before_action { @back_url = support_case_path(@current_case, anchor: "tasklist") }
+
+    def edit
+      @draft = Email::Draft.new(
+        default_content: default_template,
+        default_subject:,
+        template_id: @template_id,
+        ticket: current_case.to_model,
+        to_recipients: @to_recipients,
+      ).save_draft!
+
+      @support_email_id = @draft.id
+      @email_evaluators = Email::Draft.find(@support_email_id)
+      parse_template
+    end
+
+    def update
+      @email_evaluators = Email::Draft.find(params[:id])
+      @email_evaluators.attributes = form_params
+      parse_template
+      if @email_evaluators.valid?(:new_message)
+        @email_evaluators.save_draft!
+        @email_evaluators.deliver_as_new_message
+
+        @current_case.update!(sent_email_to_evaluators: true)
+
+        redirect_to @back_url
+      else
+        render :edit
+      end
+    end
+
+    def back_to_url_b64
+      return Base64.encode64(edit_support_case_message_thread_path(case_id: current_case.id, id: params[:id])) if action_name == "submit"
+
+      current_url_b64
+    end
+
+  private
+
+    def set_current_case
+      @current_case = Support::Case.find(params[:case_id])
+    end
+
+    def set_email_addresses
+      @evaluators = @current_case.evaluators.all
+      @email_addresses = @evaluators.map(&:email)
+      @to_recipients = @email_addresses.to_json
+    end
+
+    def set_documents
+      @documents = @current_case.upload_documents
+    end
+
+    def form_params
+      params.require(:email_evaluators).permit(:html_content)
+    end
+
+    def draft_email_params
+      params.require(:email_evaluators).permit(:id)
+    end
+
+    def default_subject = "Case #{current_case.ref} - invitation to complete procurement evaluation"
+
+    def default_template = render_to_string(partial: "support/cases/email_evaluators/form_template")
+
+    def set_template
+      template = Support::EmailTemplate.find_by(title: "Invitation to complete procurement evaluation")
+      @template_id = template.id if template
+      @unique_link = evaluation_verify_evaluators_unique_link_path(@current_case, host: request.host)
+    end
+
+    def parse_template
+      @email_evaluators.html_content = Support::EmailEvaluatorsVariableParser.new(@current_case, @email_evaluators, @unique_link).parse_template
+    end
+  end
+end
--- a/app/models/email/template_parser.rb
+++ b/app/models/email/template_parser.rb
@@ -2,6 +2,10 @@ class Email::TemplateParser
   def initialize(variables: {})
     @variables = {
       "caseworker_full_name" => Current.actor.try(:full_name),
+      "organisation_name" => "{{organisation_name}}",
+      "sub_category" => "{{sub_category}}",
+      "unique_case_specific_link" => "{{unique_case_specific_link}}",
+      "evaluation_due_date" => "{{evaluation_due_date}}",
     }.merge(variables)
   end
 
--- a/app/models/support/case.rb
+++ b/app/models/support/case.rb
@@ -209,7 +209,7 @@ module Support
       agent.notify_assigned_to_case(support_case: self, assigned_by:)
     end
 
-    def sub_category_aeiou_format
+    def sub_category_with_indefinite_article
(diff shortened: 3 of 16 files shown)


app/controllers/support/cases/email_evaluators_controller.rb
```diff
@@ -0,0 +1,97 @@
...
+      template = Support::EmailTemplate.find_by(title: "Invitation to complete procurement evaluation")
+      @template_id = template.id if template
+    end
+
+    def parse_template
+      unique_link = evaluation_verify_evaluators_unique_link_path(@current_case, host: request.host)
+      variables = {
+        "organisation_name" => current_case.organisation_name || current_case.email,
+        "sub_category" => current_case.sub_category || "[sub_category]",
+        "unique_case_specific_link" => "<a target='_blank' rel='noopener noreferrer' href='#{unique_link}'>unique case-specific link</a>",
```
threedaymonk, line 89: This needs to be properly escaped. Ideally, we'd use [link_to].

It might be nice to extract the all the email sending into a separate object, maybe somewhere in `app/services`, because it feels a bit out of scope for the controller.

[link_to]: https://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to

app/models/email/template_parser.rb
```diff
@@ -2,6 +2,10 @@ class Email::TemplateParser
   def initialize(variables: {})
     @variables = {
       "caseworker_full_name" => Current.actor.try(:full_name),
+      "organisation_name" => "{{organisation_name}}",
+      "sub_category" => "{{sub_category}}",
+      "unique_case_specific_link" => "{{unique_case_specific_link}}",
+      "evaluation_due_date" => "{{evaluation_due_date}}",
```
threedaymonk, line 8: do you need the interpolation? Would this work?

```ruby
"organisation_name" => organisation_name,
# etc.
```

spec/features/support/agent_can_email_to_evaluators_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    visit support_case_path(support_case, anchor: "tasklist")
+
+    expect(find("#complete-evaluation-4-status")).to have_text("To do")
+
+    visit edit_support_case_email_evaluators_path(support_case)
+
+    expect(page).to have_content("Email evaluators")
+
+    create(:support_email, :inbox, ticket: support_case, outlook_conversation_id: "OCID1", subject: "Email Vealuators", recipients: [{ "name" => "Test 1", "address" => "test1@email.com" }], unique_body: "Email 1", is_read: false)
```
threedaymonk, line 45: Typo: Veaulators => Evaluators

spec/models/support/case_spec.rb
```diff
@@ -59,7 +59,7 @@
...
+        "id,ref,category_id,request_text,support_level,status,state,created_at,updated_at,agent_id,first_name,last_name,email,phone_number,source,organisation_id,existing_contract_id,new_contract_id,procurement_id,savings_status,savings_estimate_method,savings_actual_method,savings_estimate,savings_actual,action_required,organisation_type,value,closure_reason,extension_number,other_category,other_query,procurement_amount,confidence_level,special_requirements,query_id,exit_survey_sent,detected_category_id,creation_source,user_selected_category,created_by_id,procurement_stage_id,initial_request_text,with_school,next_key_date,next_key_date_description,discovery_method,discovery_method_other_text,project,other_school_urns,is_evaluator,evaluation_due_date,has_uploaded_documents,sent_email_to_evaluators\n",
```
threedaymonk, line 62: Is this generated CSV ingested anywhere? Do we really need every attribute in it? Can we delete it entirely?

threedaymonk (changes requested):
Can we also test that the email is sent and that the link in the email goes somewhere useful, so that we have confidence in the whole process from end to end?

spec/features/support/agent_can_email_to_evaluators_spec.rb
edwin-jebaraj: Updated

app/controllers/support/cases/email_evaluators_controller.rb
edwin-jebaraj: Moved to new service file

spec/models/support/case_spec.rb
edwin-jebaraj: removed

app/models/email/template_parser.rb
edwin-jebaraj: We need this for email template parsing

app/controllers/support/cases/email_evaluators_controller.rb
threedaymonk: I was thinking more that the whole of the template parsing should be extracted, so that the `parse_template` directives in the controller would be more like:

```ruby
@parse_template = SomeService.call(...)
```

app/services/support/email_evaluators_variable_parser.rb
```diff
@@ -2,18 +2,24 @@ module Support
...
+    def initialize(current_case, email_evaluators, unique_link)
       @current_case = current_case
+      @email_evaluators = email_evaluators
       @unique_link = unique_link
     end
 
     def variables
       {
         "organisation_name" => @current_case.organisation_name || @current_case.email,
+        "a_or_an" => @current_case.sub_category.starts_with?("A", "E", "I", "O", "U") ? "an" : "a",
```
threedaymonk, line 14: This is unfortunately much more complicated!

- A universal method
- An honour

It would be better to rewrite these emails so that they don't need an indefinite article, I think.

app/models/support/case.rb
```diff
@@ -208,5 +208,9 @@ def assign_to_agent(agent, assigned_by: Current.agent)
       update!(agent:)
       agent.notify_assigned_to_case(support_case: self, assigned_by:)
     end
+
+    def sub_category_aeiou_format
```
threedaymonk, line 212: I think I'd probably choose something that describes the outcome here - "sub category with indefinite article" - rather than the mechanism

app/views/support/cases/show/_tasklist.html.erb
```diff
@@ -20,9 +20,17 @@
         end
 
         task_list.with_item(title: I18n.t("support.case.label.tasklist.item.upload_documents"), href: edit_support_case_document_uploads_path(@current_case), status: document_upload_Status)
-        task_list.with_item(title: I18n.t("support.case.label.tasklist.item.email_evaluators")) do | item | 
-        item.with_status(text: I18n.t("support.case.label.tasklist.status.cannot_start"), cannot_start_yet: true)
+
+        if (@current_case.evaluators.count > 0 && @current_case.evaluation_due_date && @current_case.has_uploaded_documents && @current_case.sent_email_to_evaluators)
```
threedaymonk, line 24: instead of `count > 0`, `any?` would explain the intention better.

config/routes.rb
```diff
@@ -336,6 +337,11 @@
     end
   end
 
+  namespace :evaluation do
+    resources :tasks, only: %i[show edit]
+    get "verify/evaluator/link/:id", to: "tasks#edit", as: :verify_evaluators_unique_link
```
threedaymonk, line 342: I was wondering whether this was going to end up unreasonably long with the already-long domain name in front, but I think it's OK as we're sending HTML emails.

threedaymonk (commented):
I think this is OK, but I'm going to ask @serenaabbott11 to take a look as she's working more directly on this and might spot something I've missed.

threedaymonk (approved):

PR #2022, opened 2025-01-09
SC-78 Evaluator journey: accessing evaluation from a link 

edwin-jebaraj:
## Changes in this PR

1. app/controllers/cms_entry_points_controller.rb - check  email_evaluator_link session and redirect to evaluator tasklist page
2. app/services/create_user.rb - updated dsi_uid to the support evaluators


threedaymonk (changes requested):
If this behaviour is worth adding, it's worth testing, so that we know what it's meant to do!

threedaymonk (approved):

PR #2025, opened 2025-01-23
SC-78 fixed parsing issue when sub category is nil

edwin-jebaraj:
## Changes in this PR

SC-78 fixed parsing issue when sub category is nil


threedaymonk (changes requested):
All bugs fixed should have tests. Ideally, the test should be written first, and fail, and then pass when the code is fixed.

threedaymonk (approved):
I think this solves the immediate problem (error instead of page loading) but instead of "[sub_category]", maybe we shouldn't be interpolating the category at all if it's missing.

PR #2028, opened 2025-01-29
SC-150 - Refactor email address validation

serenaabbott11:
## Changes in this PR
- Refactored email validation for adding evaluators, framework requests, and creating a new case
- Other similar validation has NOT been changed due to complications with using forms and schemas

### DIFF
--- a/app/forms/framework_requests/email_form.rb
+++ b/app/forms/framework_requests/email_form.rb
@@ -1,40 +1,14 @@
 module FrameworkRequests
   class EmailForm < BaseForm
-    validates :email, presence: true, format: { with: URI::MailTo::EMAIL_REGEXP }
-    validate :email_is_not_personal
+    validates :email, presence: true
+    validates :email, email_address: { format: true }
+    validates :email, email_address: { school_email: true }
 
     attr_accessor :email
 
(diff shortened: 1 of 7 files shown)


app/forms/framework_requests/email_form.rb
```diff
@@ -1,40 +1,14 @@
 module FrameworkRequests
   class EmailForm < BaseForm
-    validates :email, presence: true, format: { with: URI::MailTo::EMAIL_REGEXP }
-    validate :email_is_not_personal
+    validates :email, presence: true
+    validates :email, email_address: { format: true }
+    validates :email, email_address: { school_email: true }
```
threedaymonk, line 5: I think these can be merged into one line:

```ruby
validates :email, email_address: { format: true, school_email: true }
```

but don't let it get in the way of getting this finished and merged!

threedaymonk (approved):

app/forms/framework_requests/email_form.rb
serenaabbott11: thank you! I tried this and it weirdly didn't work - but it may have been due to the mess of other files i had going on at the time! will try again tomorrow then pass over to QA 👍

PR #2035, opened 2025-02-17
Added fix for tab list overflow and updated tablist label

edwin-jebaraj:
## Changes in this PR
1. app/assets/stylesheets/application.sass.scss - Added css to override procurement tabs screen
2. app/views/support/cases/show/_tabs.html.erb - Added new class to override style
3. config/locales/en.yml - Updated the label name

**Before:**
image

**After:** 
image


threedaymonk (approved):
This will do for now, but I'm concerned that we've already hit the limits of a tabbed interface and need to push back and ask for a better design.

PR #2038, opened 2025-02-19
SC-85 Evaluator journey: Download documents - Updated mailto subject

edwin-jebaraj:
## Changes in this PR

app/views/evaluation/download_documents/show.html.erb - Updated mailto subject


app/views/evaluation/download_documents/show.html.erb
```diff
@@ -3,7 +3,7 @@
 <h1 class="govuk-heading-l"><%= I18n.t("evaluation.download_documents.header") %></h1>
 <p class="govuk-body">Use these documents to complete your evaluation of each supplier.</p>
 <p class="govuk-body govuk-!-static-margin-bottom-8">If you have a problem or conflict of interest, 
-  <% mail_to_address = "mailto:#{ENV.fetch('MS_GRAPH_SHARED_MAILBOX_ADDRESS')}?subject= Case #{ @current_case.ref} - query" %>
+  <% mail_to_address = "mailto:#{ENV.fetch('MS_GRAPH_SHARED_MAILBOX_ADDRESS')}?subject=Case #{ @current_case.ref} - Evaluation query from #{ current_evaluator.name}" %>
   <%= link_to "contact us", mail_to_address , class: "govuk-link" %>
```
threedaymonk, line 7: I think we need to look at escaping here, in case the name contains punctuation. I think the spaces probably should have been escaped already, but the best thing would be to use the [`mail_to`][mail_to] helper instead of `link_to`.

[mail_to]: https://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-mail_to

threedaymonk (changes requested):

app/views/evaluation/download_documents/show.html.erb
edwin-jebaraj: updated

threedaymonk (approved):

PR #2044, opened 2025-02-26
SC-213 Add notification list entry Proc Ops when an evaluation is submitted

edwin-jebaraj:
## Changes in this PR
1. app/controllers/evaluation/upload_completed_documents_controller.rb - Added methods to update notification details
2. app/models/support/case/notifiable.rb - Added a method to update notification details
3. app/models/support/notification.rb - Added new topic enum value
4.  app/views/support/notifications/topics/_evaluation_documents_uploaded.html.erb - Added new file for notification
5. spec/features/evaluation/upload_completed_documents_spec.rb - Added test case to check notification count when evaluation is submitted

### DIFF
--- a/app/controllers/evaluation/upload_completed_documents_controller.rb
+++ b/app/controllers/evaluation/upload_completed_documents_controller.rb
@@ -14,6 +14,7 @@ module Evaluation
         @document_uploader.save_evaluation_document!(current_user.email)
         current_evaluator.update!(case_document_uploader_params)
         update_action_required
+        notify_procops_document_uploaded
         redirect_to @back_url
       else
         render :show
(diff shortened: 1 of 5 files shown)


app/controllers/evaluation/upload_completed_documents_controller.rb
```diff
@@ -81,5 +86,21 @@ def notify_procops_action_required?
       pending_evaluations = Support::Evaluator.where(support_case_id: params[:case_id], has_uploaded_documents: true, evaluation_approved: false).any?
       unread_emails || pending_evaluations
     end
+
+    def agent_present_and_documents_uploaded?
+      @current_case.agent.present? && case_document_uploader_params[:has_uploaded_documents] == "true"
```
threedaymonk, line 91: Why is this the literal string "true"? I'd have expected "1" from an HTML form.

This scenario doesn't seem to be tested.

edwin-jebaraj: updated condition

threedaymonk (approved):

PR #2047, opened 2025-03-03
Make delivery of email a background job

doctorpod:
Emails are sent via the Microsoft Graph API, which is unreliable and may fail due to timeouts or errors; currently, failures are silent. To improve reliability, this moves email sending to a background job with retries

## Changes in this PR

- Add new method `Email::Draft#queue_delivery`
- 7 controllers now call this instead of `Email::Draft.deliver_as_(new_message/reply)`

### DIFF
--- a/app/jobs/deliver_email_job.rb
+++ b/app/jobs/deliver_email_job.rb
@@ -1,0 +1,13 @@
+class DeliverEmailJob < ApplicationJob
+  queue_as :emailing # FIXME: Set this queue up?
+
+  def perform(id, send_type)
+    draft = Email::Draft.find(id)
+
+    if send_type == :as_new_message
+      draft.deliver_as_new_message
+    elsif send_type == :as_reply
+      draft.delivery_as_reply
+    end
+  end
+end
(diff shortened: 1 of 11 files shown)


app/jobs/deliver_email_job.rb
```diff
@@ -0,0 +1,13 @@
+class DeliverEmailJob < ApplicationJob
+  queue_as :emailing # FIXME: Set this queue up?
```
threedaymonk, line 2: This looks like it's set up in `sidekiq.yml`

app/jobs/deliver_email_job.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  def perform(id, send_type)
+    draft = Email::Draft.find(id)
```
threedaymonk, line 5: Email::Draft isn't backed by a database, so we'll need to:

- create a database table, with blob storage for attachments - use `has_attached_file`?
- ensure that the object is saved
- (probably) delete the record and attachments when it's been successfully sent, as they are potentially large

(yes, we are recreating a wheel already solved by SMTP, ain't the cloud grand)

PR #2053, opened 2025-03-11
SC-116 Fixed handover task list status error

edwin-jebaraj:
## Changes in this PR
1. app/views/support/cases/show/_tasklist.html.erb - Updated condition to display handover tasklist status
2. spec/features/support/case_contract_handover_tasklist_spec.rb -  Added test file to check the status


app/views/support/cases/show/_tasklist.html.erb
```diff
@@ -90,9 +90,9 @@
           end
         end
 
-        if @current_case.downloaded_all_contract_handover_pack?
+        if @current_case.contract_recipients.any? && @current_case.upload_contract_handover_complete? && @current_case.downloaded_all_contract_handover_pack?
           download_handover_status = govuk_tag(text: I18n.t("support.case.label.tasklist.status.complete"), colour: "green")
-        elsif @current_case.downloaded_any_contract_handover_pack?
+        elsif @current_case.contract_recipients.any? && @current_case.upload_contract_handover_complete? && @current_case.downloaded_any_contract_handover_pack?
```
threedaymonk, line 95: This logic is pretty complicated, so it might be better to move these onto a couple of methods on `@current_case`, or possibly onto a presenter object for this scenario. These could then be tested in isolation.

threedaymonk (changes requested):

app/views/support/cases/show/_tasklist.html.erb
threedaymonk: Generally, my feeling is that logic in views should be as simple as possible, and any complexity should be pushed elsewhere, either the controller, the model, or a presenter.

edwin-jebaraj: updated

threedaymonk (approved):

PR #2063, opened 2025-03-27
SC-227 Evaluator Login - Session independence

edwin-jebaraj:
## Changes in this PR
1. app/controllers/evaluation/signin_controller.rb - Added new sign in controller for evaluator login
2. app/controllers/evaluation/tasks_controller.rb - Updated file for evaluator sign in page redirect
3. app/controllers/sessions_controller.rb - Updated file for evaluator sign in page redirect
4. app/controllers/specify/home_controller.rb - Updated file for root path redirect
5. app/views/specify/home/show.html.erb - Updated file for evaluator sign in page redirect
6. config/routes.rb -Added new route path for evaluator sign in page 
7. app/views/evaluation/signin/show.html.erb - Added new view file for evaluator sign in page 
8. spec/features/evaluation/signin_spec.rb - Added test case for evaluator sign in page


app/controllers/evaluation/signin_controller.rb
```diff
@@ -0,0 +1,10 @@
+module Evaluation
+  class SigninController < ApplicationController
+    skip_before_action :authenticate_user!
+    def show
+      @start_now_button_route = current_user.guest? ? "/auth/dfe" : dashboard_path
```
threedaymonk, line 5: I note that `/auth/dfe` doesn't appear in the output of `rails routes`. Should or could it be a named route?

threedaymonk (approved):

app/controllers/evaluation/signin_controller.rb
edwin-jebaraj: keeping same as existing pattern

PR #2065, opened 2025-04-01
Make evaluator upload docs task clickable but not editable when evalu…

serenaabbott11:
## Changes in this PR
Allow the upload docs task to be clickable from the tasklist, and allow it to be edited when the evaluation has **not** been approved by Proc Ops.


threedaymonk (approved):
One suggestion, but otherwise it's good!

PR #2068, opened 2025-04-02
SC-312 Procurement task list - Add users new 100 limit

edwin-jebaraj:
## Changes in this PR
1. app/models/support/contract_recipient.rb - Added new validation
2. app/models/support/evaluator.rb - Added new validation
3. config/locales/en.yml - Added new language variable
4. spec/models/support/contract_recipient_spec.rb - Added new test case for validation
5. spec/models/support/evaluator_spec.rb -  Added new test case for validation

### DIFF
--- a/spec/models/support/contract_recipient_spec.rb
+++ b/spec/models/support/contract_recipient_spec.rb
@@ -1,0 +1,52 @@
+RSpec.describe Support::ContractRecipient, type: :model do
+  describe "validation" do
+    it "enforces case insensitive email uniqueness" do
+      support_case = create(:support_case)
+      create(:support_contract_recipient, support_case:, email: "test@example.com")
+      contract_recipient = build(:support_contract_recipient, support_case:)
+
+      contract_recipient.email = "TEST@example.com"
+      expect(contract_recipient).not_to be_valid
+
+      contract_recipient.email = "different@example.com"
+      expect(contract_recipient).to be_valid
+    end
+
+    it "scopes email uniqueness by case" do
+      case_a = create(:support_case)
+      case_b = create(:support_case)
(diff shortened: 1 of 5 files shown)


spec/models/support/contract_recipient_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      contract_recipient.email = "example"
+      expect(contract_recipient).not_to be_valid
+    end
+
+    it "does not allow more than 100 contract recipients per support case" do
+      support_case = create(:support_case)
+
+      100.times do |i|
+        create(:support_contract_recipient, support_case:, email: "contract_recipient#{i}@example.com")
+      end
```
threedaymonk, line 46: it's much more efficient if you use `create_list` here, and you shouldn't need to specify the email address as it's already using a sequence in the factory.

```ruby
create_list(:support_contract_recipient, 100)
```

spec/models/support/evaluator_spec.rb
```diff
@@ -37,5 +37,18 @@
       evaluator.email = "example"
       expect(evaluator).not_to be_valid
     end
+
+    it "does not allow more than 100 evaluators per support case" do
+      support_case = create(:support_case)
+
+      100.times do |i|
+        create(:support_evaluator, support_case:, email: "evaluator#{i}@example.com")
+      end
```
threedaymonk, line 46: as above

threedaymonk (changes requested):

spec/models/support/evaluator_spec.rb
edwin-jebaraj: updated

PR #2079, opened 2025-04-15
Version updates

threedaymonk:
Update versions of Ruby, Rails, and other gems in preparation for IT health check.


dinesh16 (approved):
LGTM 👍 (There are some existing warning for mixing case styles in Dockerfile, probably we could change it later.)

threedaymonk:
@dinesh16 I have a separate PR for the Dockerfile which I'll publish shortly.

PR #2099, opened 2025-04-28
Post sign-in redirect to school picker screen

doctorpod:
## Changes in this PR
New session variable to trigger post sign-in redirect to first screen of the onboarding flow


app/controllers/sessions_controller.rb
```diff
@@ -11,6 +11,11 @@ def create
     user_session.invalidate_other_user_sessions(auth: auth_hash)
     user = CreateUser.new(auth: auth_hash).call
 
+    if session[:energy_onboarding] == true
```
edwin-jebaraj, line 14: can you move this to entry_path. We need to test both internal and external user

edwin-jebaraj (changes requested):

edwin-jebaraj (approved):

PR #2114, opened 2025-05-06
Add Gas single or multi screen

doctorpod:
Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/ES-164


app/controllers/energy/gas_single_multis_controller.rb
```diff
@@ -0,0 +1,36 @@
+module Energy
+  class GasSingleMultisController < ApplicationController
+    before_action :organisation_details
```
edwin-jebaraj, line 3: call form method before update action 
`before_action :form, only: %i[update]`

edwin-jebaraj (changes requested):

doctorpod:
Closed due to suddenly having 70 commits after rebase to latest `feature/energy` -- new PR made by cherry picking the correct commits onto a new branch

PR #2127, opened 2025-05-12
ES-163 | Add energy Task list

dinesh16:
## Changes in this PR
This is the task list for information about your schools. Its not the final version as we haven't completed all the screens. 
There will be more work to this task list on later stages,
Currently Tasklist class has tests for only available screens, Need to add more tests later stage including feature test.

image


doctorpod (approved):
LGTM - one suggestion but not urgent

PR #2130, opened 2025-05-13
ES-287 cec homepage

serenaabbott11:
## Changes in this PR
- Add all cases and my cases to the homepage
- Show list of support cases with a sub-category of DfE Energy for Schools
- Include filters and sort
- Look similar to procops support cases list

Screenshot 2025-05-13 at 15 50 55

### DIFF
--- a/app/models/support/case/filterable.rb
+++ b/app/models/support/case/filterable.rb
@@ -65,6 +65,8 @@ module Support::Case::Filterable
     scope :created_by_e_and_o, -> { where(creation_source: :engagement_and_outreach_team) }
 
     scope :created_by, ->(created_by_ids) { where(created_by_id: created_by_ids) }
+
+    scope :dfe_energy, -> { by_category(Support::Category.find_by(title: "DfE Energy for Schools service")) }
   end
 
   class_methods do
--- a/app/views/cec/onboarding_cases/case_list/_case_list.html.erb
+++ b/app/views/cec/onboarding_cases/case_list/_case_list.html.erb
@@ -1,0 +1,19 @@
+<div class="case-list govuk-grid-row">
+  <div class="govuk-grid-column-one-quarter">
+    <%= render partial: "cec/onboarding_cases/case_list/filters", locals: { tab:, filter_form:, filter_form_scope:, filter_url:, clear_url: } %>
+  </div>
+  <div class="govuk-grid-column-three-quarters">
+	
+    <% if cases.none? %>
+      <h2 class="govuk-heading-m"><%= I18n.t("support.case.list.no_cases") %></h2>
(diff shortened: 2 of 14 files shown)


app/models/support/case/filterable.rb
```diff
@@ -65,6 +65,8 @@ module Support::Case::Filterable
     scope :created_by_e_and_o, -> { where(creation_source: :engagement_and_outreach_team) }
 
     scope :created_by, ->(created_by_ids) { where(created_by_id: created_by_ids) }
+
+    scope :dfe_energy, -> { by_category(Support::Category.find_by(title: "DfE Energy for Schools service")) }
```
threedaymonk, line 69: Matching on literal text feels a bit fragile, but I don't have a concrete suggestion for improvement.

app/views/cec/onboarding_cases/case_list/_case_list.html.erb
```diff
@@ -0,0 +1,19 @@
...
+  <div class="govuk-grid-column-three-quarters">
+	
+    <% if cases.none? %>
+      <h2 class="govuk-heading-m"><%= I18n.t("support.case.list.no_cases") %></h2>
+      <p class="govuk-body"><%= I18n.t("support.case.list.no_cases_advice") %></p>
+    <% else %>
+      <ul>
+				<li>
+	        <%= render partial: "cec/onboarding_cases/case_list/case_panel/case_panel", collection: Support::CasePresenter.wrap_collection(cases), as: :kase, locals: { tab: }  %>
+				</li>
```
threedaymonk, line 14: It's not the most important thing, but the indentation has gone a bit wacky here!

app/views/cec/onboarding_cases/case_list/case_panel/_top_left_pane.html.erb
```diff
@@ -0,0 +1,37 @@
...
+    <dd class="govuk-summary-list__value">
+      <%= kase.organisation_name %>
+    </dd>
+  </div>
+  <div class="govuk-summary-list__row">
+    <dt class="govuk-summary-list__key">
+      <%= I18n.t("support.case.label.level") %>
+    </dt>
+    <dd class="govuk-summary-list__value">
+      <%= kase.support_level.nil? ? "Not set" : 1 + kase.read_attribute_before_type_cast(:support_level) %>
```
threedaymonk, line 26: I wonder if this should be an attribute of the case; `presentational_level` or something like that perhaps.

threedaymonk (approved):
I have a few thoughts, but they're not blockers.

app/models/support/case/filterable.rb
serenaabbott11: Yeah I also had this thought but couldn't think of what to do, as the title is the only thing that is supposed to be matched with

PR #2136, opened 2025-05-14
ES-360 Onboarding case - return rule (allow more than one case per Org)

edwin-jebaraj:
## Changes in this PR
[68224b1](https://github.com/DFE-Digital/buy-for-your-school/pull/2136/commits/68224b1441194f6a10ceb5897654995f78b8a3e5) - Added Onboarding case - return rule

### DIFF
--- a/app/controllers/energy/authorisation_controller.rb
+++ b/app/controllers/energy/authorisation_controller.rb
@@ -15,7 +15,7 @@ module Energy
 
   private
 
-    def existing_onboarding_case?
+    def existing_onboarding_organisations
       Energy::OnboardingCaseOrganisation.where(onboardable: @support_organisation)
     end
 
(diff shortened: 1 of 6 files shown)


app/controllers/energy/authorisation_controller.rb
```diff
@@ -1,27 +1,58 @@
...
       return unless params[:type] == "single"
 
-      onboarding_case = existing_onboarding_case? || create_onboarding_case
+      onboarding_case = create_onboarding_case
       redirect_to energy_case_switch_energy_path(case_id: onboarding_case.energy_onboarding_case_id)
     end
 
   private
 
     def existing_onboarding_case?
```
doctorpod, line 18: If this may bring back multiple onboarding organisations, this should probably be named something like `existing_onboarding_organisations` without the question mark

app/controllers/energy/authorisation_controller.rb
```diff
@@ -1,27 +1,58 @@
...
-      Energy::OnboardingCaseOrganisation.find_by(onboardable: @support_organisation)
+      Energy::OnboardingCaseOrganisation.where(onboardable: @support_organisation)
     end
 
     def create_onboarding_case
       Energy::CaseCreatable.create_case(current_user, @support_organisation)
     end
 
+    def check_active_onboarding_case
+      onboarding_case_organisation = existing_onboarding_case?
```
doctorpod, line 27: Pluralise `onboarding_case_organisations`

doctorpod (approved):
I think I understand what the ticket wanted :/ -- just a couple of re-naming suggestions

app/controllers/energy/authorisation_controller.rb
edwin-jebaraj: updated

PR #2156, opened 2025-05-21
ES 168 check your answers

doctorpod:
Ticket:: https://dfedigital.atlassian.net.mcas.ms/browse/ES-168

### DIFF
--- a/app/controllers/energy/application_controller.rb
+++ b/app/controllers/energy/application_controller.rb
@@ -16,11 +16,11 @@ module Energy
     end
 
     def set_routing_flags
-      @routing_flags = {tasks: params[:tasks], check: params[:check]}
+      @routing_flags = { tasks: params[:tasks], check: params[:check] }
     end
 
     def set_from_tasks_or_check
-      @from_tasks_or_check = (params[:tasks] == "1" || params[:check] == "1")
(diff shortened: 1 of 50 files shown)


dinesh16:
Hi @doctorpod 
I would suggest to have separate class for `AnswerList` and that can have list of `Answer` object similar to TaskList and Task approach rather than amending existing TaskList. So we can keep it simple and clean approach and well as follow the SRS principle and make it easy to test and east maintain/change. Then we can grab any common functionalities to module/concerns and reuse it in both AnswerList and task list. What do you think? @edwin-jebaraj any suggestions?

doctorpod:
> Hi @doctorpod I would suggest to have separate class for `AnswerList` and that can have list of `Answer` object similar to TaskList and Task approach rather than amending existing TaskList. So we can keep it simple and clean approach as well as follow the SRS principle and make it easy to test and maintain. Then we can extract any common functionalities to module/concerns and reuse it in both AnswerList and TaskList. What do you think? @edwin-jebaraj any suggestions?

Fair point @dinesh16 - However we may not now have time to split TaskList - we may have to do it after release :/

edwin-jebaraj (approved):
LGTM

edwin-jebaraj (approved):
LGTM, Let's push this dev

PR #2167, opened 2025-05-23
ES-335 | send energy onboarding email to the school

dinesh16:

### DIFF
--- a/app/controllers/energy/authorisation_controller.rb
+++ b/app/controllers/energy/authorisation_controller.rb
@@ -9,8 +9,12 @@ module Energy
     def update
       return unless params[:type] == "single"
 
-      onboarding_case = create_onboarding_case
-      redirect_to energy_case_switch_energy_path(case_id: onboarding_case.energy_onboarding_case_id)
+      @onboarding_case_organisation = create_onboarding_case
+      @current_support_case = @onboarding_case_organisation.onboarding_case.support_case
+
+      draft_and_send_onboarding_email_to_school(current_user.email.to_json)
+
+      redirect_to energy_case_switch_energy_path(case_id: @onboarding_case_organisation.energy_onboarding_case_id)
     end
 
   private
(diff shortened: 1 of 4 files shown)


app/controllers/energy/authorisation_controller.rb
```diff
@@ -76,5 +80,39 @@ def support_organisation_and_valid_urns
...
+    def email_template
+      @email_template ||= Support::EmailTemplate.find_by(title: "Energy onboarding Form started (email)")
+    end
+
+    def parse_template
+      @energy_onboarding_email.html_content = Energy::Emails::SchoolOnboardingVariableParser.new(@current_support_case, @energy_onboarding_email, onboarding_case_link).parse_template
+    end
+
+    def onboarding_case_link
+      "/energy/case/#{@onboarding_case_organisation.energy_onboarding_case_id}/switch_energy"
```
edwin-jebaraj, line 97: try this way `energy_case_switch_energy_path( host: request.host)`

app/models/email/template_parser.rb
```diff
@@ -1,7 +1,8 @@
 class Email::TemplateParser
   def initialize(variables: {})
     @variables = {
-      "caseworker_full_name" => Current.actor.try(:full_name),
```
edwin-jebaraj, line 4: don't change existing caseworker_full_name variable. use new variable name.

app/views/energy/authorisation/_onboarding_email_template.erb
```diff
@@ -0,0 +1,15 @@
...
+
+<p>To complete the form, visit <span>{{unique_case_specific_link}}</span> and complete the sections labelled ‘Not started’ or ‘In progress’. </p>
+
+<p>You can return to the form and change your answers at any point before it is submitted. </p>
+
+<p>If you need help with the form, you can contact DfE by replying to this email. </p>
+
+<p>Kind regards,</p>
+
+<p>The Energy for Schools Team </p>
```
edwin-jebaraj, line 15: You might need to add DISCLAIMER.

edwin-jebaraj (changes requested):

app/views/energy/authorisation/_onboarding_email_template.erb
dinesh16: currently, not on the template, but we can add it if we need this to the template. I think its not a blocker for now.

edwin-jebaraj (approved):
LGTM, Let's push this to _development_

PR #2175, opened 2025-05-28
Task list status derivation and outbound routing

doctorpod:
Ticket: https://dfedigital.atlassian.net.mcas.ms/browse/ES-395


app/controllers/energy/gas_meter_controller.rb
```diff
@@ -15,12 +15,14 @@ def new
     end
 
     @gas_meter_detail = @onboarding_case_organisation.gas_meters.new
+    form_url
```
edwin-jebaraj, line 18: do we need to call `form_url` here? it's already there in `before_action`

app/controllers/energy/gas_meter_controller.rb
```diff
@@ -15,12 +15,14 @@ def new
...
   end
 
   def edit; end
 
   def create
     @gas_meter_detail = @onboarding_case_organisation.gas_meters.new(form_params)
+    form_url
```
edwin-jebaraj, line 25: same as previous comment, we already done callback in before_action

edwin-jebaraj (approved):
LGTM

app/controllers/energy/gas_meter_controller.rb
serenaabbott11: @doctorpod can you confirm please?

app/controllers/energy/gas_meter_controller.rb
doctorpod: Yes. In most controllers we can get away with a before_action call but in this case, the URL depends on whether the object is new or not so it had to be defined in separate places. An alternative would be to just assign @form_url in place (in the method) rather than call the method

PR #2183, opened 2025-06-03
ES-424 amend survey link and guidance

serenaabbott11:
## Changes in this PR
- Change some of the wording in the onboarding pages and add a line above the related actions to match the prototype
- Change the service to be energy for schools for the customer satisfaction survey when coming from an energy url
- Add confirmation screen as a source so that the end of journey survey works


dinesh16 (approved):
LGTM 👍

PR #2185, opened 2025-06-03
ES-335 - fix redirection of form started email link

dinesh16:
## Changes in this PR
https://dfedigital.atlassian.net.mcas.ms/browse/ES-335

As part of the energy form started email, when the user click on the link, 
- For the signed in users: It should redirect to task list if the user has selected switching energy type, if switching energy type is not selected then the user will be redirected to switch energy screen. 
- For the non sign-in users: It should redirected to energy sign in page and then goes to task list

### DIFF
--- a/spec/controllers/sessions_controller_spec.rb
+++ b/spec/controllers/sessions_controller_spec.rb
@@ -88,7 +88,6 @@ describe SessionsController do
       end
 
       it "redirects to my_procurements_verify_unique_link_path" do
-        # binding.pry
         expect(response).to redirect_to(my_procurements_task_path(support_case))
       end
     end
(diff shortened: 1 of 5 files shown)


edwin-jebaraj:
Can you add some test cases for this?

spec/controllers/sessions_controller_spec.rb
```diff
@@ -88,8 +88,26 @@
       end
 
       it "redirects to my_procurements_verify_unique_link_path" do
+        # binding.pry
```
edwin-jebaraj, line 91: remove this

edwin-jebaraj (changes requested):

spec/controllers/sessions_controller_spec.rb
```diff
@@ -88,8 +88,26 @@
...
+          energy_onboarding: true,
+        }
+      end
+
+      it "redirects to task list" do
+        expect(response).to redirect_to(energy_task_list_path)
+      end
+    end
   end
 end
```
edwin-jebaraj, line 113: also please add user not logged in test case

PR #2187, opened 2025-06-05
ES-395 fixes

serenaabbott11:
## Changes in this PR
- Redirect from the task list to the meter type pages for gas and electric if it hasn't been selected before (clicked discard and go to task list rather than saved on the meter type screen)
- Changes logic so if vat rate declared is false then it's in progress rather than complete (so user can't progress unless boxes are ticked)
- Remove vat data relating to 5% vat rate when vat rate is changed to 20%
- Set default address when vat details are incorrect and there are no addresses to choose from on the vat contact screen


edwin-jebaraj (approved):
LGTM

PR #2200, opened 2025-06-11
Add redirect based on case submitted time stamp

doctorpod:
Ticket: https://dfedigital.atlassian.net/browse/ES-447


app/controllers/energy/application_controller.rb
```diff
@@ -15,6 +15,12 @@ def check_flag
       render "errors/not_found", status: :not_found unless Flipper.enabled?(:energy)
     end
 
+    def check_if_submitted
+      return if %w[school_selections authorisation confirmations].include? controller_name
```
threedaymonk, line 19: would it make more sense to redefine this method as a no-op in these controllers?

threedaymonk (approved):
Looks fine, but my preference would be to use method dispatch in preference to inspecting the string version of the current class where possible.

app/controllers/energy/application_controller.rb
doctorpod: Tanks Paul - I think I'll use _skip_before_action_ in the controllers

PR #2206, opened 2025-06-12
ES-481 - Add text to show user the feedback link opens in a new tab

serenaabbott11:
## Changes in this PR
- Add text to show user the feedback link opens in a new tab but the span is visually hidden within the link so that it is only picked up by assistive technology

### DIFF
--- a/spec/helpers/specify/application_helper_spec.rb
+++ b/spec/helpers/specify/application_helper_spec.rb
@@ -9,7 +9,7 @@ RSpec.describe ApplicationHelper, type: :helper do
   describe "#banner_message" do
     it "returns the beta message by default" do
       # banner.beta.message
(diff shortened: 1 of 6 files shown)


doctorpod (approved):

spec/helpers/specify/application_helper_spec.rb
```diff
@@ -9,7 +9,7 @@
   describe "#banner_message" do
     it "returns the beta message by default" do
       # banner.beta.message
-      expect(helper.banner_message).to eq "This is a new service – your <a class=\"govuk-link\" target=\"_blank\" rel=\"noopener\" href=\"#{CGI.escapeHTML(new_customer_satisfaction_survey_path(source: 'banner_link', service: 'create_a_spec'))}\">feedback</a> will help us to improve it."
+      expect(helper.banner_message).to eq "This is a new service – your <a class=\"govuk-link\" target=\"_blank\" rel=\"noopener\" href=\"#{CGI.escapeHTML(new_customer_satisfaction_survey_path(source: 'banner_link', service: 'create_a_spec'))}\">feedback</a> will help us to improve it (opens in new tab)."
```
titlescreen, line 12: I think the link needs to resolve like this…

<a href=""…>LINK TEXT <span class="visually-hidden"> opens in new tab</span></a>

doctorpod (approved):

PR #2296, opened 2025-08-19
Feature/es 597 fix msgraph crash

eanya-dfe:
#Problem

- The ghbs-dev-web container was crashing on startup due to an uninitialized constant MicrosoftGraph::*.

- Root cause: the microsoft_graph gem was referenced in our code (lib/email/message_cacheable/resync_outlook_ids.rb) before being properly loaded/configured.

- The initializer also assumed MS Graph credentials were always present, which caused hard failures when env vars were missing.

#Change

- Ensured the gem is always required so Zeitwerk can see MicrosoftGraph::* constants.

- Wrapped the configuration logic in a conditional (ENABLE_MS_GRAPH flag + required env vars).

- If credentials are missing or the feature flag is off, the app boots cleanly and logs an informational message instead of crashing.

- No business logic changes, just safer initialisation.

#Why

- Minimal fix: Only touched the initialiser, didn’t alter any core app logic.

- Unblocks development: Allows dev/test/staging environments without MS Graph configured to run the app.

- Team-friendly: Added ENABLE_MS_GRAPH flag so MS Graph can be explicitly turned on/off per environment without breaking everyone’s containers.

- Production-safe: In prod, where credentials exist, behavior remains the same.

#Impact

- Dev/staging containers won’t crash when MS Graph isn’t configured.

- Zeitwerk still loads constants correctly.

- No functional change for environments with proper credentials.

- Minimal, low-risk change: only initialiser logic touched.


eanya-dfe:
No longer required at the moment. Will reopen if necessary

PR #2470, opened 2026-03-24
SER-280 | moving controllers, routes and views

dinesh16:
## Changes in this PR

Ticket: https://dfedigital.atlassian.net/browse/SER-280

This is moving FABs rails application content to current CMS application. As part of the migration, the contents will be moved bit by bit to the feature branch and on this PR mainly focusing on Controllers, Views Routes and assets.


retmanski (commented):
We need to firm up the behaviour for static pages - currently this doesn't work as expected and will only ever show the BFYS static pages (which is fine once all of the FABS Contentful content gets migrated across).

app/controllers/fabs/pages_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+    end
+
+    trail.reverse_each do |n|
+      case n
+      when FABS::Category
+        add_breadcrumb category_breadcrumb_name, category_breadcrumb_path
+      when FABS::Page
+        add_breadcrumb n.title, page_path(n.slug)
+      end
```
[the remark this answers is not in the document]
dinesh16: We can merge, may be ideal to have sensible value for else statement.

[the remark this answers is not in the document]
retmanski: Okay, let's just ignore Sonarcube then. I don't see anything wrong with the existing code.

app/controllers/contentful_webhooks_controller.rb
```diff
@@ -0,0 +1,53 @@
+class ContentfulWebhooksController < Fabs::ApplicationController
+  skip_before_action :verify_authenticity_token
```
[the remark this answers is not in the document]
retmanski: Ignore this recommendation

app/controllers/pages_controller.rb
```diff
@@ -1,10 +1,25 @@
...
-  before_action :set_breadcrumbs, only: :show
+
+  include Breadcrumbs
 
   def show
-    @page = PagePresenter.new(page)
+    if page.present?
+      set_breadcrumbs
+      @page = PagePresenter.new(page)
+      render "pages/show", layout: "application"
```
retmanski, line 10: you can remove this line as it's the default behaviour anyway

app/views/categories/show.html.erb
```diff
@@ -0,0 +1,62 @@
+<h2> SHOW categories here</h2>
```
retmanski: Delete this line?

retmanski (commented):
Just a couple of minor comments (few lines to remove) but otherwise looks good.

retmanski (approved):
LGTM 👍

app/controllers/solutions_controller.rb
```diff
@@ -25,13 +25,13 @@ def show
     @page_title = @solution.title
     @page_description = @solution.description
     @page_header_class = "details-header"
-    add_breadcrumb home_breadcrumb_name, home_breadcrumb_path
+    add_breadcrumb_on_rails :home_breadcrumb_name, :home_breadcrumb_path
     @canonical_url = category_solution_url(@primary_category.slug, @solution.slug)
-
+    # binding.pry
```
retmanski: Remove please

app/controllers/categories_controller.rb
```diff
@@ -9,7 +9,7 @@ def index
   end
 
   def show
-    add_breadcrumb home_breadcrumb_name, root_path
+    add_breadcrumb_on_rails :home_breadcrumb_name, :root_path
```
retmanski, line 12: Could you explain why the syntax for this has changed please? It doesn't seem to match the docs for the breadcrumb gem (that just shows `add_breadcrumb` as we already had) 🤷‍♂️ 

https://github.com/weppos/breadcrumbs_on_rails

retmanski (changes requested):

app/controllers/categories_controller.rb
dinesh16: Yeh, this was bit tricky one @retmanski, It wasn't working with default method `add_breadcrumb` as we have another breadcrumb gem [loaf](https://github.com/piotrmurach/loaf), when we call add_breadcrumb by default it uses `Loaf::Crumb` object so you don;t see actual breadcrumb on the frontend actually it should create `BreadcrumbsOnRails::Breadcrumbs::Element` . Then I had to dig the `breadcrumb_on_rails` gem and use the method `add_breadcrumb_on_rails` which is the real method name and `add_breadcrumb` is an alias of it. Its quick win less code changes. Any suggestions are welcom.

retmanski (approved):

PR #2471, opened 2026-04-13
Add search and selected contentful management rake tasks from FABS

retmanski:
Add search and selected contentful management rake tasks from FABS codebase

This includes additional gems to support OpenSearch and Contentful management, the rake tasks and SearchClient model that was omitted when migrating the models previously. I've also added some documentation around the use of OpenSearch and test coverage for the migrated rake tasks.

I've deliberately excluded tasks relating to Contentful / i18n sync and Asset copying as we won't use this in the merged application.


serenaabbott11 (approved):
Looks good! Agree that these sonarcloud failures can be ignored and probably should look at adjusting the rules as you suggested. 

Only thing I can think of is adding the environment variables into the .env.example file? We haven't been good at this and probably should do a check of the .env files/env variables especially now we are picking the terraform up again so could do this separately

PR #2474, opened 2026-04-20
CORE-59 | remove fabs endpoint and api implementation

dinesh16:
Ticket: https://dfedigital.atlassian.net/browse/CORE-59

## Changes in this PR

After merging FABS and FAF, the /bfys/solutions API is no longer needed. Instead of BFYS calling its own endpoint, the SyncFrameworks job can directly use Solution model.

- Remove routes
- Remove api implementation
- Update and refactor specs

### Todo
- Remove FAF_FRAMEWORK_ENDPOINT environment variable


retmanski (commented):
Looks good - only thing I would mention is that the ENV var `FAF_FRAMEWORK_ENDPOINT` is probably now redundant, so that can be removed from any example .env files or documentation.

Also, have a look at this rake task in `data_migrations.rake` - seems to be using the ENV var and trying to hit the same API endpoint. This can probably be either refactored or removed entirely - not sure which...

```ruby
  desc "Populate Support::Framework refs"
  task populate_framework_refs: :environment do
    uri = URI.parse(ENV["FAF_FRAMEWORK_ENDPOINT"])
    response =
      Net::HTTP.start(uri.host, uri.port, use_ssl: uri.scheme == "https") do |http|
        request = Net::HTTP::Get.new(uri)
        http.request(request)
      end

    if response.code == "200"
      frameworks = JSON.parse(response.body)
      prepared_frameworks = frameworks.map do |framework|
        {
          name: framework["title"],
          supplier: framework["provider"]["initials"],
          ref: framework["ref"],
        }
      end

      Support::Framework.all.find_each do |framework|
        match = prepared_frameworks.select { |p| p[:name] == framework.name && p[:supplier] == framework.supplier }
        framework.update!(ref: match.first[:ref]) if match.present?
      end
    end
  end
```

retmanski (approved):
Good job, looks much cleaner now 👍 

Note that I haven't tested this yet, would be good to get this onto QA and give the tyres a  kick if possible.

dinesh16:
Tested and everything looks good. The service was run manually and all checks passed. The rake task also ran successfully at midnight.

PR #2475, opened 2026-04-21
Refactor all of the original FABS styles

retmanski:
When FABS code was merged into GHBS we initially copied the entire stylesheet into the codebase. This introduced a fair bit of tech debt with duplicate imports of GDS styles, redundant classes and conflicting approaches to packaging images. That in turn led to image rendering issues (particularly notable in footer styling).

With GHBS, images are rendered directly from the frontend node packages without using filename digests, so they don't need to be present in the `app/assets/images` folder. But with the former FABS app, a custom rake task was used (and patched into `assets:precompile` task) to explicitly copy all images from the frontend node packages to `app/assets/images` folder.

When the FABS code was merged into GHBS we also included these images explicitly in `app/assets/images` rather than using the original rake tasks, which in turn means that when `assets:precompile` runs during deployment all of these images are then only available with digest filenames. All of the stylesheets now use direct URL's without any Rails helpers to include the digests so these then return 500 errors when trying to render images... which is a long explanation for why we see missing images when deployed to Azure.

I've removed all of the duplicated front end images, redundant styles, aligned some styles with correct GDS classes and otherwise refactored FABS styles into appropriate style component groups. This affects a fair few screens but is scoped only to the FABS pages (homepage, categories, solutions etc.). It would be good to audit the original GHBS styles at some point in the future but I didn't want to inflate the scope at this point.

This might be tricky to review as a whole so is perhaps better to go through each commit to see the incremental changes. I've been careful to check before/after styling for each area but might have missed something.


dinesh16 (approved):
LGTM 👍

PR #2476, opened 2026-04-21
Core 62 - Change references from CCS to GCA

serenaabbott11:
## Changes in this PR
Crown Commercial Service (CCS) are rebranding to Government Commercial Agency (GCA). This ticket changes the references where applicable. https://dfedigital.atlassian.net/browse/CORE-62


dinesh16 (approved):
@serenaabbott11 looks good, I think we need to do the same change on the email templates on all the environments as well.

retmanski:
> @serenaabbott11 looks good, I think we need to do the same change on the email templates on all the environments as well.

Yep - two more references as below:

Screenshot 2026-04-21 at 15 20 31

PR #2481, opened 2026-04-28
 CORE-257, CORE-258 | Extend SSL certificate renewal workflow to supp…

dinesh16:
…ort multiple domains

## Changes in this PR

Tickets: 
https://dfedigital.atlassian.net/browse/CORE-257
https://dfedigital.atlassian.net/browse/CORE-258

**Update SSL certificate validation workflow to support multiple domains**

The scheduled GitHub Actions workflow validates SSL certificates before expiry to support automatic renewal. Previously, the workflow only handled a single domain. This change extends the workflow to validate and manage both domains, allowing the existing renewal process to work consistently across multiple domain names.

Test plan:

This workflow cannot be fully validated locally because it depends on Azure-managed certificate validation and GitHub Actions scheduled execution. Changes were reviewed by testing workflow logic locally where possible (including matrix/job flow and script execution).

### DIFF
--- a/.github/workflows/support-update-ssl-cert-validation-implementation.yml
+++ b/.github/workflows/support-update-ssl-cert-validation-implementation.yml
@@ -41,7 +41,7 @@ jobs:
               customDomainName="devghbscustom-domain0"
               zoneName="dev.get-help-buying-for-schools.service.gov.uk"
             else
-              customDomainName="devghbscustom-domain1"
+              customDomainName="dev-get-help-buying-for-schools-education-gov-uk-ab1b"
               zoneName="dev.get-help-buying-for-schools.education.gov.uk"
             fi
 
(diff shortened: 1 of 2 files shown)


.github/workflows/support-update-ssl-cert-validation-implementation.yml
```diff
@@ -26,22 +30,45 @@ jobs:
     steps:
       - name: Setup variables
         id: vars
+        env:
+          DOMAIN_FAMILY: ${{ inputs.domain_family }}
```
retmanski, line 35: Would be nice to add a comment here e.g.

`# Supports "service" and "education" values for *.service.gov.uk and *.education.gov.uk domains`

retmanski (approved):
Nice clean implementation, just one extra comment as a nice to have please :)

PR #2484, opened 2026-05-01
Add feature flag to support deprecation of GHBS Contentful instance

retmanski:
This flag allows individual pages to be retired from GHBS Contentful instance; if the specified slug is present in the `DEPRECATED_GHBS_CONTENTFUL_PAGES` list then it will be served from FABS Contentful instance instead (if it exists there). Once all of the pages are migrated to FABS Contentful this can be removed and the existing Postgres backed `Page` model can be used to add another caching layer for Contentful content.

As this behaviour is controlled by a feature flag we can deprecate pages in dev/staging without affecting production deployments.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -3,6 +3,8 @@ class PagesController < ApplicationController
 
   include Breadcrumbs
 
+  DEPRECATED_GHBS_CONTENTFUL_PAGES = %w[foo].freeze
+
   def show
     if page.present?
       set_breadcrumbs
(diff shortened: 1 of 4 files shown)


app/controllers/pages_controller.rb
```diff
@@ -3,6 +3,8 @@ class PagesController < ApplicationController
 
   include Breadcrumbs
 
+  DEPRECATED_GHBS_CONTENTFUL_PAGES = %w[foo].freeze
```
retmanski, line 6: We'll replace this placeholder value when adding the first page to deprecate and then add more page slugs as needed.

app/controllers/pages_controller.rb
```diff
@@ -30,11 +32,13 @@ def self.bypass_dsi?
...
 
   def page
+    return if deprecated_ghbs_contentful_page?
+
     @page ||= Page.find_by(slug: params[:slug])
   end
 
   def fabs_page
-    @page = FABS::Page.find_by_slug!(params[:slug])
+    @fabs_page ||= FABS::Page.find_by_slug!(params[:slug])
```
retmanski, line 41: Cleaner to memoize this function as it gets called twice, so this cuts out an extra API call to Contentful.

serenaabbott11 (approved):

PR #2485, opened 2026-05-01
CORE-344, CORE-336 | update 404, 500 and 422 error pages

dinesh16:
## Changes in this PR
Tickets: 
https://dfedigital.atlassian.net/browse/CORE-344
https://dfedigital.atlassian.net/browse/CORE-336

Fix 404 and 500 error pages.

### DIFF
--- a/app/views/errors/internal_server_error.html.erb
+++ b/app/views/errors/internal_server_error.html.erb
@@ -1,5 +1,6 @@
-<%= content_for :page_title_prefix, t("errors.internal_server_error.page_title") %>
-<main class="govuk-main-wrapper" id="main-content">
+<%= content_for :title, t("errors.internal_server_error.page_title") %>
+
+<div class="dfe-width-container">
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-full">
       <h1 class="govuk-heading-xl">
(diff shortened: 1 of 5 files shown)


spec/features/specify/errors/custom_errors_spec.rb
```diff
@@ -13,15 +13,17 @@
 
     it "shows the expected error message" do
       visit "/foo"
+      # binding.pry
```
serenaabbott11: can remove this now

serenaabbott11 (approved):
Just left one comment but otherwise good to go!

app/views/errors/internal_server_error.html.erb
```diff
@@ -1,5 +1,6 @@
-<%= content_for :page_title_prefix, t("errors.internal_server_error.page_title") %>
-<main class="govuk-main-wrapper" id="main-content">
+<%= content_for :title, t("errors.internal_server_error.page_title") %>
+
+<div class="dfe-width-container">
```
retmanski: It would be better for accessibility to retain the `main` tag and associated `id` attribute rather than replacing with a `div` but the new class is fine i.e.

```html
<main class="dfe-width-container" id="main-content">
```

The same applies to the other two views. This is needed because there's a "Skip to main content" link in the main application layout.

retmanski: My apologies, @dinesh16 I didn't realise there was already an outer `<main>` tag outside of this template, so your change is good 👍

PR #2487, opened 2026-05-05
Deprecated service privacy policy from GHBS Contentful instance

retmanski:
This is now served from FABS Contentful instead, allowing a common privacy policy across the whole site. Took the opportunity to use the correct govuk helper methods for various links to the privacy policy. Updated test coverage accordingly.

Also moved `shared.dfe_footer.privacy_notice` locale entry and made use of this consistently (there were previously links named "Privacy policy" - this is now "Privacy notice" correctly matching the page title).


app/views/shared/_dfe_footer.html.erb
```diff
@@ -17,7 +17,7 @@
             <a class="govuk-footer__link" href="/accessibility-statement"><%= t(".accessibility_statement") %></a>
           </li>
           <li class="govuk-footer__inline-list-item">
-            <a class="govuk-footer__link" href="https://www.gov.uk/government/organisations/department-for-education/about/personal-information-charter" target="_blank" rel="noopener noreferrer"><%= t(".privacy_notice") %><span class="govuk-visually-hidden"> <%= t("shared.external_link.opens_in_new_tab") %></span></a>
+            <%= govuk_footer_link_to t("shared.privacy_notice"), "/privacy" %>
```
retmanski, line 20: Note this used to open in another tab as it was an external link. It's now an internal link so makes sense to open in the current tab like the other links do.

dinesh16 (approved):
LGTM

PR #2490, opened 2026-05-08
Core 347 remove selenium

retmanski:
This PR fixes (almost) all of the flaky test behaviour, most of which is around feature tests that use full stack browser testing. Historically this has been unusable locally as it results in constant spurious failures, and quite a few failures on CI resulting in a number of tests having to be flagged as `flaky` and thus excluded from the main test run.

It's also not been ideal having the CI and local test environments being quite different - previously CI was using a very old Selenium/Chrome docker image pegged to a browser version that's several years old meaning that tests would pass locally and not remotely or vice-versa, so I've attempted to align those more closely in this PR.

Key change here is to move away from Selenium/Webdriver stack and to use Cuprite. This is an alternate Capybara driver (so retains compatibility with the existing suite) but is far more robust. It should also be faster as it's designed around Chrome Devtools protocol (CDP) for tight integration with Chrome, and removes reliance on WebDriver/ChromeDriver. Given that we only run feature specs against Chrome this feels like a sensible choice.

I've also swapped out the bulky Selenium docker image for a lightweight one that only has Chrome and associated dependencies, but also exposes the necessary CDP port for browser discovery. This is pinned to a recent stable version of Chrome so should closely match our local test environment.

Only two specs still have the `flaky` tag - both are related to file uploads and a race condition on determining upload completion. The cleanest fix for this requires a code change so that will be left for a follow up isolated PR.

Once this is merged we should have a stable test suite that works both locally and on the CI platform - the next steps are then to look at improving the performance.


dinesh16 (approved):
LGTM 👍

serenaabbott11 (approved):

PR #2493, opened 2026-05-15
CORE-21 - Add cookie banner to the fabs pages and bring up to GDS standards

serenaabbott11:
## Changes in this PR
https://dfedigital.atlassian.net/browse/CORE-21

This adds the cookie banner and same google analytics tracking (and app insights) to the incoming FABS pages, and also slightly changes the content of the cookie preferences page to fit with GDS standards. 

Screenshot 2026-05-15 at 10 40 49


app/views/shared/_dfe_base.html.erb
```diff
@@ -1,6 +1,28 @@
 <!DOCTYPE html>
 <html lang="en" class="govuk-template--rebranded">
 <head>
+  <% if record_ga? %>
```
retmanski: As this is a direct copy from the `application.html.erb` layout it would be nice to extract this into a shared partial e.g.`shared/_google_analytics.html.erb` and then include in that layout and this partial too rather than duplicate it.

retmanski (commented):
My preference would be to extract the GA script mess into a dedicated partial rather than having two copies of that. Otherwise looks good 👍

app/views/shared/_dfe_base.html.erb
```diff
@@ -1,6 +1,28 @@
...
+    <!-- Google tag (gtag.js) --> <script async nonce=<%= content_security_policy_nonce%> src="https://www.googletagmanager.com/gtag/js?id=<%= google_analytics_id %>"></script> <script nonce=<%= content_security_policy_nonce%>> window.dataLayer = window.dataLayer || []; function gtag(){dataLayer.push(arguments);} gtag('js', new Date()); gtag('config', "<%= google_analytics_id %>"); </script>
+  <% end %>
+
+  <% if Rails.env.production? && ENV['ApplicationInsights__ConnectionString'].present? %>
```
retmanski: Ah, as we have Azure insights too maybe just change the partial name to something more generic like `shared/_analytics_tags.html.erb`. One partial for both GA and App insights is fine, or separate ones for each if you prefer so long as the markup isn't duplicated.

retmanski (approved):
LGTM - thanks for refactoring into the shared partial 👍

PR #2494, opened 2026-05-15
Updated privacy policy to link to external page

retmanski:
I've added a new helper method for this to avoid repeated markup.

As this won't be hosted in Contentful any more I've removed the `privacy` slug from the deprecated pages list.


dinesh16 (approved):
LGTM 👍

PR #2501, opened 2026-06-01
CORE-420 - Consolidate accessibility statement into one and link from footer

serenaabbott11:
## Changes in this PR
- Changes the accessibility link in the footer to just point at one consolidated external link


retmanski (approved):
Lovely job 👍

PR #2503, opened 2026-06-10
Remove FABS feature flag and associated GHBS_HOMEPAGE_URL

retmanski:
The FABS functionality and domain migration is now fully complete so this feature is enabled globally, meaning the redirect before filter and any test or routing hacks can all be cleaned up. Once this is merged the `GHBS_HOMEPAGE_URL` environment variable can be removed from any Azure or Github secrets too.


dinesh16 (approved):
👍

PR #2505, opened 2026-06-11
CORE-280 | fix incorrect routing on sign-in failure

dinesh16:
## Changes in this PR

Ticket: https://dfedigital.atlassian.net/browse/CORE-280

This is fix for the incorrect routing on sign-in failure, currently if any auth failure it redirects to home page. This fix will redirect to correct page with sign in failure error message.

image


retmanski (approved):
Very elegant 👍

PR #2507, opened 2026-06-16
Core 598 new header and mega menu

retmanski:
**DO NOT MERGE**

DRAFT: This PR introduces a new gem and a couple of GDS components to replace the existing header and search for FABS pages. Whilst this approach did work with a lot of tweaks, it was found that the `govuk_publishing_components` gem has an enormous dependency graph and introduces other gems that conflict with GHBS.

An alternate approach was then taken, see: https://github.com/DFE-Digital/buy-for-your-school/pull/2508


app/views/govuk_publishing_components/components/_search.html.erb
retmanski, line 1: This is copied from the govuk publishing components gem and has a few tweaks to support the GHBS use case - notably to add placeholder support (the inline label behaves weirdly, we used a placeholder before that works better) and to match the endpoint and query params that the search controller expects (this avoids having to rework our downstream analytics reporting).

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
retmanski, line 1: This is copied from the govuk publishing components gem and has a few tweaks to support the GHBS use case - notably to swap out the logo and tweak the search so that it uses a regular search rather than an autocomplete search. Also removes some GA4 tracking stuff that we have handled elsewhere.

app/views/shared/_dfe_homepage_header.html.erb
```diff
@@ -1,8 +1,4 @@
-<header class="dfe-header dfe-header--homepage" data-module="govuk-header">
-  <%= render "shared/dfe_logo_and_menu" %>
-  <%= render "shared/dfe_search_box"%>
-  <%= render "shared/dfe_service_navigation" %>
```
retmanski, line 4: We decided to drop the service navigation links from the homepage for now, but retain those elsewhere.

retmanski:
Closing this PR now - superceded by #2508 without adding the entire gem (that's now merged).

PR #2512, opened 2026-06-26
CORE-608 | category buying options with filters

dinesh16:
## Changes in this PR
Spike: Draft PR for category buying options with Filters


dinesh16:
No longer need this. This is part of spike and actual work is covered in https://github.com/DFE-Digital/buy-for-your-school/pull/2532

PR #2514, opened 2026-07-02
Remove custom id override for "Other" case origin text field

retmanski:
Using the custom override breaks the associated label generated by the form builder (the label still retains "for" attribute for the default input id), which then means the label and input are no longer semantically linked. This looks to be the root cause of the accessibility fails. I can't see any good reason to override the input id (there is no reference to the custom id anywhere in this code base), so the cleanest fix seems to be just reverting to the default id which then matches up with the label and hopefully fixes the accessibility issue raised.

Also refactored feature spec create framework request as a guest. This is notoriously flaky so has been disabled for a while; a key issue with this particular feature spec is that is repeats many previous steps to check the next step, so the steps get increasingly longer and flakiness (seems to be mostly around autocomplete timing) gets worse as the test suite progresses. Changed to use a shared framework request held as test metadata so that each page can work without needing to visit every preceding page. This still tests basically the same things but is more efficient and should be a lot less flaky. More can be done to improve this as things like checking page copy can be simple request specs instead but this already brings the test down from a few minutes to ~30s.


dinesh16 (approved):
LGTM 👍

PR #2521, opened 2026-07-06
CORE-637 | Change the header link to always be the homepage of ghbs

serenaabbott11:
## Changes in this PR
- Fixes the bug where on RFH the header link goes to the RFH page, now anywhere there is a header link it will just go to the home page


dinesh16 (approved):
LGTM 👍

retmanski (approved):
Love it 👍

PR #2525, opened 2026-07-07
CORE-650 | Change robots.txt file to allow everything to improve SEO

serenaabbott11:
## Changes in this PR
- Remove the lines in robots.txt that prevent crawling which will allow for better SEO


retmanski (approved):
LGTM 👍

PR #2526, opened 2026-07-07
CORE-687 | Fix wrong letter case on check your answers invoicing method

serenaabbott11:
## Changes in this PR
- Fix the letter case on a field for the energy check your answers page so it is Paper and Email rather than paper and email


dinesh16 (approved):

retmanski (approved):
Looks fine, but I'd like to clean up the `add_attribute` method at some point.. it's using `send` as a sort of meta programming but allowing that to be overridden anyway so ends up calling methods twice plus feels unnecessarily exposed.

It feels overly complex and all of this code would be cleaner if it just appended things directly to the task list summary. But never mind, that's a problem for another day.

I'll add it to the growing list of things to review :)

PR #2527, opened 2026-07-08
Core 665 azure ai search

retmanski:
Add Azure AI Search as an alternative backend for solution search, with Contentful remaining in use for categories. This PR adds the Azure indexing/search client, webhook support for keeping the Azure index in sync with Contentful changes made throughout the day, plus a nightly reindex job.

Also Introduces a new feature flag to enable Azure AI search for each environment (when enabled this overrides the `USE_OPENSEARCH` environment variable).

Azure AI search supports a number of search approaches; for now this uses the `simple` model which is essentially the same as Elasticsearch (text based, no AI features) and works on the free plan. Rake tasks are added to manage the search index on Azure (very similar to ElasticSearch indexing approach).


doc/search.md
```diff
@@ -1,44 +1,69 @@
...
+If `AZURE_AI_SEARCH_SEMANTIC_CONFIGURATION` is omitted, the application uses the default semantic configuration name defined in code.
 
-- creates the `solution-data` index if it does not already exist
+### Indexing solutions
+
+Azure solution search is populated by:
+
+```sh
+bundle exec rake azure_search:create_index
+bundle exec rake azure_search:index
```
serenaabbott11: should these be azure_ai_search:... rather than azure_search:...? same on line 79 and 85

retmanski: Good spot, thankyou 👍 I renamed the rake task after writing the docs.

serenaabbott11 (approved):

PR #2528, opened 2026-07-10
CORE-647 | Removes the link to create a spec from procurement support page

serenaabbott11:
## Changes in this PR
- There is a link on the procurement support page which says it goes to create a specification - this service no longer exists and instead the link goes to the GHBS homepage, so this entire bullet point is removed.

Screenshot 2026-07-10 at 14 01 55


retmanski (approved):
Good catch 👍

PR #2532, opened 2026-07-15
CORE-606 | Add buying options page with filters 

dinesh16:
## Changes in this PR

Tickets: https://dfedigital.atlassian.net/browse/CORE-606

### Summary

This PR introduces the **Buying Options** page, which users land on after selecting a category from the homepage.

The page displays all available solutions for the selected category and provides filters that allow users to narrow the results by:

* **Subcategories**
* **Ways to buy**

It also includes additional page components such as the promotional banner and the expert help section.

The PR includes new and updated specs to cover the added functionality.

### Review notes

* **Initial review:** Please focus on the implementation and functionality.
* **Follow-up review:** Once the functionality is approved, I'd appreciate a second pass to review the styling and ensure it matches the latest designs.


app/models/FABS/category.rb
```diff
@@ -73,13 +74,27 @@ def self.find_by_slug!(slug)
...
-          subcategory_slugs.include?(subcat.fields[:slug])
-        end
+        subcategory_match =
+          if subcategory_slugs.present?
+            solution.subcategories&.any? do |subcategory|
+              subcategory_slugs.include?(subcategory.fields[:slug])
+            end
+          else
+            true
+          end
```
retmanski: I think would be nice to clean this up a little bit, would be easier to read too e.g.

```ruby
subcategory_match = subcategory_slugs.blank? || solution.subcategories&.any? do
  subcategory_slugs.include? subcategory.fields[:slug]
end

ways_to_buy_match = ways_to_buy_slugs.blank? || ways_to_buy_slugs.include?(solution.ways_to_buy&.fields&.[](:slug))
```

retmanski: Also, under what circumstances would `fields` be nil for ways to buy.. this might be a tad too defensive - would `solution.ways_to_buy&.fields[:slug]` still work?

retmanski (commented):
Very clean, just one comment around some logic that could perhaps be tidied a bit. Nice job 👍

app/models/FABS/category.rb
dinesh16: Hi Andrew, technically it should not be nil, each solution should have  a `ways_to_by` reference unless you add it on. For safe side, as there are 3 environments and if any existing solutions doesn't have/ or forgot to add `way to buy` attributes the filter might break, So added safe navigation.  I think covert this `olution.ways_to_buy&.fields[:slug]` to `solution.ways_to_buy&.fields&.[](:slug)` is rubocop thing.

config/locales/en.yml
```diff
@@ -3077,6 +3083,14 @@ en:
       view_all_buying_options: Browse a list of all DfE-approved buying options
     external_link:
       opens_in_new_tab: opens in new tab
+    right_filters:
+      apply_filters: Apply filers
```
retmanski: Should be "Apply filters"

app/views/categories/show.html.erb
```diff
@@ -1,60 +1,48 @@
...
-      <div class="results-count">
-        <%= t("results.count", count: @solutions.count) %>
-      </div>
-    </div>
-
-    <% unless params.fetch(:subcategory_slugs, []).compact_blank.blank? %>
-      <%= render "categories/filter_labels" %>
-    <% end %>
-  </div>
+<div style="clear:both;"></div>
```
retmanski: This was just intended to be a temporary fix - if clears are needed then would be better to use/define a utility CSS class for this and apply to a more suitable element in the markup.

app/assets/stylesheets/application.sass.scss
```diff
@@ -64,6 +64,14 @@
   white-space: nowrap;
 }
 
+.sidebar {
+  margin-top: 25px;
```
retmanski: This is already defined in `_sidebar.scss` so would prefer to adjust the existing style there.

app/assets/stylesheets/application.sass.scss
```diff
@@ -64,6 +64,14 @@
...
+}
+
+.promo-banner {
```
retmanski: Prefer to move this to `_banner.scss`

app/controllers/categories_controller.rb
```diff
@@ -10,23 +11,52 @@ def index
 
   def show
     add_breadcrumb_on_rails :home_breadcrumb_name, :root_path
+    add_breadcrumb_on_rails :category_breadcrumb_name
```
retmanski: Why is the breadcrumb method being called twice?

app/controllers/categories_controller.rb
```diff
@@ -1,5 +1,6 @@
 class CategoriesController < Fabs::ApplicationController
   before_action :disable_search_in_header, only: :index
+  before_action :ways_to_buy, only: :show
```
retmanski: Given that all the other instance variables in show method are assigned explicitly it would be better to follow that pattern, so remove this before filter.

app/controllers/categories_controller.rb
```diff
@@ -10,23 +11,52 @@ def index
...
-    @subcategories = @category.subcategories
+    @subcategories = category.subcategories
     @selected_subcategories = @subcategories.select { |subcategory| params[:subcategory_slugs]&.include?(subcategory.slug) }
-    @solutions = @category.filtered_solutions(subcategory_slugs: params[:subcategory_slugs]&.compact_blank)
-    @dfe_solutions, @other_solutions = @solutions.partition(&:buying_option_type)
+
+    @solutions = @category.filtered_solutions(subcategory_slugs: params[:subcategory_slugs]&.compact_blank, ways_to_buy_slugs: params[:ways_to_buy_slugs]&.compact_blank)
+
+    @solutions_count_subcategory = solutions_count_by_subcategory(@solutions)
+    @solutions_count_ways_to_buy = solutions_count_by_ways_to_buy(@solutions)
```
retmanski, line 21: Add `@ways_to_buy = ways_to_buy` line here instead of before filter

app/controllers/categories_controller.rb
```diff
@@ -10,23 +11,52 @@ def index
...
+      counts[solution.ways_to_buy&.slug] += 1
+    end
+  end
+
+  def category
+    @category ||= FABS::Category.find_by_slug!(params[:slug])
+  end
+
+  def ways_to_buy
+    @ways_to_buy ||= category.solutions.map(&:ways_to_buy).compact.map { |entry| WaysToBuy.new(entry) }.sort_by(&:title).uniq(&:title)
```
retmanski: Change to

```ruby
def ways_to_buy
  category.solutions.map(&:ways_to_buy).compact.map { |entry| WaysToBuy.new(entry) }.sort_by(&:title).uniq(&:title)
end
```

app/views/categories/show.html.erb
retmanski: See if you can get away with removing it or just using an empty div with no styles maybe?

retmanski (changes requested):
Looking pretty good, left a few comments for stuff to tidy up a bit further please.

app/controllers/categories_controller.rb
dinesh16: This is to support new design, this will add home breadcrumb with a link and  category_breadcrumb without a link. So we need to call it twice in order to achieve this.

image

retmanski: Never mind, I had a look through the gem docs and this is intentional: https://github.com/weppos/breadcrumbs_on_rails

retmanski (approved):
Thanks for revising, looks good now 👍

app/views/categories/show.html.erb
```diff
@@ -6,7 +6,9 @@
   <div class="govuk-grid-column-two-thirds govuk-!-padding-left-0">
     <div class="govuk-grid-row gov-uk-!-padding-left-0">
       <div class="govuk-grid-column-one-third">
-        <%= image_tag @category.banner&.image&.url, alt: @category.banner&.title, width: "100%", height: "auto" %>
+        <% if @category.banner.present? && @category.banner&.image&.url.present? %>
```
retmanski: Could simplify this to

```ruby
<% if @category.banner&.image&.url.present? %>
```

retmanski: or even this if the image always has a url property (which seems reasonable).

```ruby
<% if @category.banner&.image.present? %>
```

app/views/layouts/solution.html.erb
```diff
@@ -0,0 +1,31 @@
...
+      <div class="govuk-grid-column-two-thirds govuk-!-padding-left-0 govuk-!-margin-top-0 govuk-!-margin-bottom-7">
+        <%= yield %>
+      </div>
+      <%= render "shared/dfe_sidebar" %>
+    </div>
+  </div>
+
+  <%= render "shared/dfe_footer" %>
+ <% end %>
+
```
retmanski: This is a new dedicated layout just for showing a single buying solution? Why do we need that and is it in scope for this ticket? We already have separate layouts for homepage, FABS and CMS so I'm reluctant to see this proliferate further (hoping we get some clarity on the homepage soon and that dedicated layout can be removed).

dinesh16: Probably lets discuss this @retmanski

dinesh16: This is removed now Andrew, used yield to support both cases.

app/views/categories/show.html.erb
```diff
@@ -1,46 +1,46 @@
...
       <details class="govuk-details">
         <summary class="govuk-details__summary">
           <span class="govuk-details__summary-text">
-            <b><%= @category&.get_expert_help&.title %></b>
+            <b><%= @get_expert_help&.title %></b>
           </span>
         </summary>
         <div class="govuk-details__text">
-          <%= render_markdown_to_html(@category&.get_expert_help&.description) %>
+          <%= render_markdown_to_html(@get_expert_help&.description) %>
```
retmanski: You don't need the defensive `&` operator here, it's already inside an `if` block that tests it exists.

retmanski: Same above on line 39 for the title.

app/controllers/categories_controller.rb
```diff
@@ -11,22 +11,56 @@ def index
...
+      counts[solution.ways_to_buy&.slug] += 1
+    end
+  end
+
+  def category
+    @category ||= FABS::Category.find_by_slug!(params[:slug])
+  end
+
+  def ways_to_buy
+    category.solutions.map(&:ways_to_buy).compact.map { |entry| WaysToBuy.new(entry) }.sort_by(&:title).uniq(&:title)
```
retmanski, line 59: I think you can remove the `.uniq(&:title)` from this and just change the title attribute to be unique in Contenful instead. Less confusing!

app/controllers/categories_controller.rb
```diff
@@ -11,22 +11,56 @@ def index
...
+  end
+
+  def get_expert_help
+    GetExpertHelp.content
```
retmanski: Not much point having a function for this, it would still look clean inlining `GetExperHelp.content` call on line 21

app/views/categories/show.html.erb
```diff
@@ -1,60 +1,57 @@
...
+  <div class="govuk-!-margin-top-4 govuk-!-margin-bottom-4">
+    <h2 class="govuk-heading-m govuk-!-margin-bottom-0"><%= @category.body_title %></h2>
+    <p class="govuk-body-m"><%= @category.body_description %></p>
+  </div>
+  <% if @category.banner.present? %>
+    <div class="govuk-grid-column-two-thirds govuk-!-padding-left-0 govuk-!-margin-bottom-4">
+      <div class="govuk-grid-row gov-uk-!-padding-left-0">
+        <div class="govuk-grid-column-one-third">
+          <% if @category.banner&.image&.url.present? %>
+            <%= image_tag @category.banner&.image&.url, alt: @category.banner&.title, width: "100%", height: "auto" %>
```
retmanski: Don't need defensive `&` here, it's inside an `if` block that already checks `@category.banner` and `@category.banner.image` is defined.

app/views/shared/_right_filters.html.erb
```diff
@@ -0,0 +1,136 @@
...
+                        <% end %>
+                      </div>
+                    <% end %>
+                  </div>
+                </details>
+              </fieldset>
+            </div>
+          <% end %>
+          <hr />
+          <% if ways_to_buy.present? && ways_to_buy.count > 1 %>
```
retmanski: Not sure if `ways_to_buy` is an `Array` type, but if so then `[].present?` returns false so no need to also check the length.

app/controllers/categories_controller.rb
retmanski: Discussed - the title is unique in Contentful but here comes from many categories so needs this call to prevent duplicates showing.

app/views/shared/_right_filters.html.erb
dinesh16: Its an array, the idea behind this is if we have only one ways_to_buy then we don't show the filter for ways_to_buy. If we have more than one ways_to_buy we show the filter, This make sense as having single value for ways_to_buy doesnt change the result set. this is previously discussed.

app/views/categories/show.html.erb
```diff
@@ -1,60 +1,57 @@
...
+  <div class="govuk-!-margin-top-4 govuk-!-margin-bottom-4">
+    <h2 class="govuk-heading-m govuk-!-margin-bottom-0"><%= @category.body_title %></h2>
+    <p class="govuk-body-m"><%= @category.body_description %></p>
+  </div>
+  <% if @category.banner.present? %>
+    <div class="govuk-grid-column-two-thirds govuk-!-padding-left-0 govuk-!-margin-bottom-4">
+      <div class="govuk-grid-row gov-uk-!-padding-left-0">
+        <div class="govuk-grid-column-one-third">
+          <% if @category.banner.image&.url.present? %>
+            <%= image_tag @category.banner.image&.url, alt: @category.banner.title, width: "100%", height: "auto" %>
```
retmanski: You can remove this `&` guard too - it's tested for presence on the line above

app/views/shared/_right_filters.html.erb
retmanski: Sorry, my mistake. You're correct, its checking for > 0 rather than > 1.

PR #2534, opened 2026-07-16
Core 669 Contentful cleanup

retmanski:
This PR removes legacy code that connects to the GHBS instance of Contentful. Historically this was used by "Create a specification" (CAS) and to display a few static pages. CAS was decommissioned in November last year and the static pages have all been migrated to the FABS instance of Contentful. As such we have no need to connect the GHBS Contentful instance.

## Removed code:
- Contentful client connection to GHBS
- Contentful services related to fetching and persisting content
- Contentful webhook API
- Contentful caching services and associated jobs / scheduling
- Contentful presenters
- Contentful rake tasks
- Contentful specs and associated fixtures / helpers

## Changes:
- Removed old ENV var configuration for GHBS Contentful
- Updated dotenv, example .env files and Github workflows to use FABS Contentful config only
- Updated a few tests to use different test fixtures (e.g. email attachments)

## ENV vars:
- CONTENTFUL_WEBHOOK_SECRET is now **mandatory** (this was missing on all environments and means FABS webhooks are broken)
- Removed CONTENTFUL_SPACE
- Removed CONTENTFUL_DELIVERY_TOKEN
- Removed CONTENTFUL_PREVIEW_TOKEN
- Removed CONTENTFUL_ENTRY_CACHING
- Removed CONTENTFUL_ENTRY_CACHING_TTL
- Removed CONTENTFUL_GHBS_ENVIRONMENT
- Removed CONTENTFUL_WEBHOOK_API_KEY

There is still a lot more code that can be removed but this PR is already large, so have added a Jira ticket to cover removing remaining CAS models/controllers/views/presenters/tables etc. Also once this PR is merged we can cleanup PagesController to avoid needless DB queries.


dinesh16 (approved):
LGTM 👍 , I can see that both code and contentful models are 4/5 years old. Should we also remove GHBS instance from Contentful side ? Hope no one is using it in different projects. It hasn't been updated since 2022 as far as I can see.

PR #2537, opened 2026-07-24
CORE-649 - Rename about this service to about our service

serenaabbott11:
## Changes in this PR
- change the about page from 'this' to 'our' in the menu - change needed in contentful too
- remove unused navigation translations

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -147,8 +147,8 @@ en:
           description: Buy home to school transport and accessories
       - label: Guidance and support
         menu_contents:
-        - label: About this service
-          href: "/about-this-service"
+        - label: About our service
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -147,8 +147,8 @@ en:
           description: Buy home to school transport and accessories
       - label: Guidance and support
         menu_contents:
-        - label: About this service
-          href: "/about-this-service"
+        - label: About our service
+          href: "/about-our-service"
```
retmanski, line 151: Is this a page that's managed in Contentful? So presumably we need to change the slug there also and sync up with the deployment of this?

retmanski (approved):
LGTM 👍

PR #2540, opened 2026-07-28
Add a new rake task to clear down Sidekiq queues in local dev environment

retmanski:
Surprisingly Sidekiq doesn't include this out of the box and my local Sidekiq runner is filled with tonnes of DfE analytics jobs that keep throwing constant console noise (due to a missing `ENABLE_DFE_ANALYTICS=FALSE` ENV var at some point). This just clears everything and is restricted to running in development mode.


dinesh16 (approved):
LGTM 👍

PR #2544, opened 2026-08-04
Add support for redirect rules defined within Contentful CMS

retmanski:
This adds a new model for Redirect - with attributes for title (info only), source_path, destination_path and redirect_type (permanent or temporary). These redirect rules are read on first request and cached for 24 hours (they change rarely and performance is important so a long cache time is appropriate).

Currently the following redirect patterns are supported:

- `/old-page-slug` => `/new-page-slug`
- `/categories/old-name` => `/categories/new-name`
- `/categories/old-name/solution-slug` => `/categories/new-name/solution-slug`

The two category redirects are handled by a single redirect model in Contentful using `source_path` set to `/categories/old-name/*` and `destination_path` set to `/categories/new-name/*`; the `*` acts as an optional placeholder for solution slug. It will be easy to add support for other patterns in future (just needs the concern and before filter adding to relevant controller) but I've kept this scoped to our current use cases.

Also cleaned up pages route/controller slightly; this route no longer accepts anything with a file extension so the custom `favicon.ico` route is no longer required (this will now fall through the pages route and be handled by standard 404 path). The pages controller also redirects directly to the error page rather than via `/404` which causes an unnecessary redirect step.

The existing Contentful webhooks controller has been updated/fixed to also support invalidating the cached Redirect routes when any redirects are added/deleted/updated within Contentful. The cache is then rebuilt on the next request.

### DIFF
--- a/app/controllers/categories_controller.rb
+++ b/app/controllers/categories_controller.rb
@@ -1,5 +1,8 @@
 class CategoriesController < Fabs::ApplicationController
   before_action :disable_search_in_header, only: :index
+  include Redirectable
+
+  before_action :redirect_legacy_slugs, only: :show
 
   DUMMY_SUBCATEGORY_SLUG = "not-applicable".freeze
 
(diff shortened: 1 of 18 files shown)


dinesh16 (approved):
LGTM 👍

PR #2546, opened 2026-08-06
CORE-606 | bug fix, hide ways to buy filter, when there is only single ways to buy option

dinesh16:
## Changes in this PR
Ticket: https://dfedigital.atlassian.net/browse/CORE-606

This bug fix ensures the Ways to buy filter is only displayed when there are multiple options available. If there is only one Ways to buy option, the filter is hidden.


retmanski (approved):
LGTM 👍

PR #2550, opened 2026-08-11
CORE-808 Remove "Buying options for" title prefix

retmanski:
Remove "Buying options for" prefix on Category landing page title.

This now just shows the Category title instead. Removed the now redundant locale entry too.


dinesh16 (approved):
LGTM 👍

